PR #13, opened 2022-11-16
Handle CAN communication for UART and USB

KmakD:


panther_driver/src/panther_can.py
```diff
@@ -172,19 +198,113 @@ def query_runtime_stat_flag(self) -> Generator:
...
-        
\ No newline at end of file
+    def _configure_tpdo(self):
+        with self._lock:
+            for motor_controller in self._motor_controllers:
+
+                try:
+                    motor_controller.can_node.tpdo.read()
+                except canopen.SdoCommunicationError:
+                    rospy.logwarn(f'[{rospy.get_name()}] Failed to read TPDO configuration.')
```
pkowalsk1: This warning pops up every time the driver is started, even though everything is running as planned. Is there anything we can do about it? E.g. to make the first warning pop up after 2-3 read attempts.

panther_driver/src/driver_node.py
```diff
@@ -280,7 +295,7 @@ def _decode_fault_flag(self, flag_val: int) -> FaultFlag:
         ref_flag_val = 0b00000001
 
         msg = FaultFlag()
-        msg.can_net_err = self._panther_can.can_connection_correct()
+        msg.can_net_err = not self._panther_can.can_connection_correct()
```
pkowalsk1: `can_connection_correct()` method is used only for detecting errors and each time is preceded by `not`. Maybe it would be better to rename it `can_connection_error()` and update returned boolean stating?

pkowalsk1 (approved):

pkowalsk1 (changes requested):
Small changes

pkowalsk1 (approved):

PR #22, opened 2022-11-24
add battery counting fnc

pkowalsk1:


panther_battery/src/adc_node.py
```diff
@@ -108,6 +110,16 @@ def _battery_timer_cb(self, *args) -> None:
...
 
+    def _check_battery_count(self) -> int:
+        try:
+            V_temp_bat_2 = self._get_adc_measurement(
+                    path="/sys/bus/iio/devices/iio:device0/in_voltage0_raw", LSB=0.002, offset=0
+                )
+        except:
+            rospy.logerr(f'[{rospy.get_name()}] Battery ADC measurement error excep')
+        
+        return 1 if V_temp_bat_2 > BAT02_DETECT_THRESH else 2
```
KmakD: If reading ADC will fail it will most likely throw error due to absence of `V_temp_bat_2`. If we want node to continue, specify default value, I would also add logwarn saying something like "Failed to evaluate battery count, using default value: 1/2"

KmakD (changes requested):

KmakD (approved):

PR #42, opened 2022-12-28
User animations

KmakD:


panther_lights_controller/README.md
```diff
@@ -22,7 +22,20 @@ Package used to control the Husarion Panther LED panels.
```
Kotochleb, line 23: I can not refer to lines not changed in this PR, but I would modify definition of some parameter.
- `~animations` [*list*, default: **None**]: it is a required ROS parameter containing a list of defined animations.
- `~global_brightness` [*float*, default: **1.0**]: LED global brightness. Range between [0,1]
- `~test` [*bool*, default: **false**]: enables testing mode with some extra functionalities.

Also. Isn't 100 hertz a bit too high frequency for displaying animations?

panther_lights_controller/README.md
```diff
@@ -22,7 +22,20 @@ Package used to control the Husarion Panther LED panels.
 
 ## Animations
 
-Basic animations are parsed as a list using the ROS parameter. They are loaded on node start directly from `config/panther_lights_animations.yaml`. Supported keys are:
+Basic animations are parsed as a list using the ROS parameter. Currently, defined animations are:
```
Kotochleb:
```suggestion
Basic animations are parsed as a list using the ROS parameter. Default animations supplied by Husarion are listed in table below.
```

panther_lights_controller/README.md
```diff
@@ -22,7 +22,20 @@ Package used to control the Husarion Panther LED panels.
...
+| 0   | E_STOP            | red expanding from the center to the edges                  |
+| 1   | READY             | green expanding from center to the edges                    |
+| 2   | ERROR             | red blinking twice                                          |
+| 3   | MANUAL_ACTION     | blue expanding from the center to the edges                 |
+| 4   | AUTONOMOUS_ACTION | orange expanding from center to the edges                   |
+| 5   | GOAL_ACHIEVED     | purple blinking three times                                 |
+| 6   | LOW_BATTERY       | two orange stripes moving towards the center, repeats twice |
+| 7   | CRITICAL_BATTERY  | two red stripes moving towards the center, repeats twice    |
+
+Basic animations are described and loaded on the node start, directly from `config/panther_lights_animations.yaml`. Supported keys are:
```
Kotochleb:
```suggestion
Default animations are described and loaded on the node start, directly from `config/panther_lights_animations.yaml`. Supported keys are:
```

panther_lights_controller/README.md
```diff
@@ -22,7 +22,20 @@ Package used to control the Husarion Panther LED panels.
...
 
 - `id` [*int*]: ID of an animation.
 - `animation` [*dict*]: description of animation.
```
Kotochleb:
```suggestion
- `animation` [*dict*]: definition of animation. See section below for more info.
```

panther_lights_controller/README.md
```diff
@@ -34,17 +47,102 @@ Basic animations are parsed as a list using the ROS parameter. They are loaded o
...
-- `image` [*string*]: path to an image file, if a path is not global will look in the `animations` folder for the given name.
+- `image` [*string*]: path to an image file. If a path is not global will look in the `animations` folder in the ROS package specified in `animations_package` for the given name.
 - `duration` [*float*]: duration of an animation.
-- `repeat` *(optional)* [*int*]: number of times the animtion will be repeated, by default animation will run once.
+- `repeat` *(optional)* [*int*]: number of times the animation will be repeated, by default animation will run once.
 - `brightness` *(optional)* [float]: animation brightness, this will be used instead of controller `global_brightness`.
 - `color` *(optional)* [*int*]: if set the image color will be changed to the specified one, advised way is to use HEX format eg. `0xff0000`.
+- `animations_package` *(optional)* [*string*]: ROS package containing `animations` folder with images. If not specified will default to this package. If using Docker given package has to be inside the same container as this package.
```
Kotochleb: Parameters are not in alphabetical order

Kotochleb: As discussed on meeting. It would be nice if package would be evaluated in path section. So, path starting with **/** would be global path, path starting with **$(find my_awesome_package)** would be path searched within ROS packages and all the other options would be invalid.

panther_lights_controller/README.md
```diff
@@ -34,17 +47,102 @@ Basic animations are parsed as a list using the ROS parameter. They are loaded o
 Animation returning frame to display based on an image. The display duration which is a product of a single image duration and repeat count can't exceed 10 seconds. Supported keys are:
 
 - `type` [*string*]: required field specyfying animation type, for `ImageAnimation` value should be `image_animation`.
-- `image` [*string*]: path to an image file, if a path is not global will look in the `animations` folder for the given name.
+- `image` [*string*]: path to an image file. If a path is not global will look in the `animations` folder in the ROS package specified in `animations_package` for the given name.
 - `duration` [*float*]: duration of an animation.
-- `repeat` *(optional)* [*int*]: number of times the animtion will be repeated, by default animation will run once.
+- `repeat` *(optional)* [*int*]: number of times the animation will be repeated, by default animation will run once.
 - `brightness` *(optional)* [float]: animation brightness, this will be used instead of controller `global_brightness`.
```
Kotochleb:
```suggestion
- `brightness` [float, optional]: animation brightness. This will overwrite `global_brightness` for a given animation.
```

panther_lights_controller/README.md
```diff
@@ -34,17 +47,102 @@ Basic animations are parsed as a list using the ROS parameter. They are loaded o
 Animation returning frame to display based on an image. The display duration which is a product of a single image duration and repeat count can't exceed 10 seconds. Supported keys are:
 
 - `type` [*string*]: required field specyfying animation type, for `ImageAnimation` value should be `image_animation`.
-- `image` [*string*]: path to an image file, if a path is not global will look in the `animations` folder for the given name.
+- `image` [*string*]: path to an image file. If a path is not global will look in the `animations` folder in the ROS package specified in `animations_package` for the given name.
 - `duration` [*float*]: duration of an animation.
-- `repeat` *(optional)* [*int*]: number of times the animtion will be repeated, by default animation will run once.
+- `repeat` *(optional)* [*int*]: number of times the animation will be repeated, by default animation will run once.
```
Kotochleb:
```suggestion
- `repeat` [*int*, optional]: number of times the animation will be repeated, by default animation will run once.
```

panther_lights_controller/README.md
```diff
@@ -34,17 +47,102 @@ Basic animations are parsed as a list using the ROS parameter. They are loaded o
...
 - `brightness` *(optional)* [float]: animation brightness, this will be used instead of controller `global_brightness`.
 - `color` *(optional)* [*int*]: if set the image color will be changed to the specified one, advised way is to use HEX format eg. `0xff0000`.
```
Kotochleb:
```suggestion
- `color` [*int*, optional]: image will be turned into grayscale and then the color will be applied with brightness from grayscale. Values have to be in HEX format.
```

panther_lights_controller/README.md
```diff
@@ -34,17 +47,102 @@ Basic animations are parsed as a list using the ROS parameter. They are loaded o
...
+- `animations_package` *(optional)* [*string*]: ROS package containing `animations` folder with images. If not specified will default to this package. If using Docker given package has to be inside the same container as this package.
+
+### Defining animations
+
+User can define own animations using basic animation types. Similar to basic ones user animations are parsed using a ROS parameter `/panther/lights/lights_controller_node/user_animations`. They can be loaded on node start or updated using the `/panther/lights/controller/update_animations` ROS service. For `ImageAnimation` you can use basic images from the `animations` folder and change their color with the `color` key ([see ImageAnimation](#imageanimation)). Follow the example below to add custom animations. 
+
+1. Create a yaml file with an animation description list. Example file: 
+
+NOTE: ID numbers from 0 to 20 are reserved for system animations.
```
Kotochleb:
```suggestion
NOTE: ID numbers from 0 to 19 are reserved for system animations.
```
Let's keep a total of 20 animations reserved for us

panther_lights_controller/README.md
```diff
@@ -34,17 +47,102 @@ Basic animations are parsed as a list using the ROS parameter. They are loaded o
...
+  # animation with custom image
+  - id: 22
+    name: 'ANIMATION_2'
+    animation:
+      both:
+        type: image_animation
+        image: /animations/custom_image.png
+        duration: 3
+        repeat: 1
+```
```
Kotochleb, line 104:
```suggestion
# animation with custom image from custom ROS package
- id: 23
    name: 'ANIMATION_3'
    animation:
      both:
        type: image_animation
        image: $(find my_custom_animation_package)/animations/custom_image.png
        duration: 3
        repeat: 1
```
```

File: panther_lights_controller/README.md [hunk 11]
```diff
@@ -34,17 +47,102 @@ Basic animations are parsed as a list using the ROS parameter. They are loaded o
...
+
+2. Add docker volume with a previously created animation description list. If using custom images for `ImageAnimation`, add also a docker volume with a folder containing custom images. On Raspberry Pi modify `compose.yaml`:
+
+```yaml
+volumes:
+  - ./user_animations.yaml:/user_animations.yaml
+  - ./animations:/animations
+```
+
```
Kotochleb, line 115:
```suggestion
**Warning**
While using docker you will only be able to find packages that are within that docker. Only images from packages that were built inside that docker imge can be  found using `$(find my_package)` syntax. Global paths work normally, but will refere to paths inside docker container.
```

panther_lights_controller/README.md
```diff
@@ -34,17 +47,102 @@ Basic animations are parsed as a list using the ROS parameter. They are loaded o
...
+```
 
-### Defining new animation
+### Defining new animation type
 
-It is possible to define your own animation with a required behavior. All animation definitions are stored in `/src/animation` and inherit from the basic class `Animation`. Each animation must overwrite the basic method `__call__` which must return an animation frame as a list of integers. An animation should also contain `ANIMATION_NAME` used to identify it. Animation frames are displayed in the controller's main timer. To tweak animation duration time use the `controller_freq` and `_duration` variables. For an example see other animation definitions.
+It is possible to define your own animation type with a required behavior. All animation definitions are stored in `/src/animation` and inherit from the basic class `Animation`. Each animation must overwrite the basic method `__call__` which must return an animation frame as a list of integers. An animation should also contain `ANIMATION_NAME` used to identify it. Animation frames are displayed in the controller's main timer. To tweak animation duration time use the `controller_freq` and `_duration` variables. For an example see other animation definitions.
```
Kotochleb:
```suggestion
It is possible to define your own animation type with expected, new behavior. All animation definitions are stored in `/src/animation` and inherit from the basic class `Animation`. Each animation must overwrite the basic method `__call__` which must return an animation frame as a list of integers. An animation should also contain `ANIMATION_NAME` used to identify it. Animation frames are displayed in the controller's main timer. To tweak animation duration time use the `controller_freq` and `_duration` variables. For an example see other animation definitions.
```

panther_lights_controller/src/animation/image_animation.py
```diff
@@ -20,10 +20,17 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
...
             rospack = rospkg.RosPack()
-            img_path = os.path.join(
-                rospack.get_path('panther_lights_controller') + f'/animations/{img_name}'
-            )
+            try:
+                img_path = os.path.join(
+                    rospack.get_path(animation_package) + f'/animations/{img_name}'
+                )
+            except rospkg.ResourceNotFound:
+                raise KeyError(f'Can\'t find ROS package: {animation_package}')
```
Kotochleb: As discussed, this might be replaced with `$(find my_awesome_package)`

panther_lights_controller/src/controller_node.py
```diff
@@ -178,6 +185,10 @@ def _set_image_animation_cb(
 
         return 'success'
 
+    def _update_animations_cb(self, req: TriggerRequest) -> TriggerResponse:
+        self._update_animations()
+        return TriggerResponse(True, '')
```
Kotochleb:
```suggestion
        return TriggerResponse(True, 'Animations updated successfully')
```

Also, shouldn't be there any try catch statement?

panther_lights_controller/src/controller_node.py
```diff
@@ -251,6 +262,24 @@ def _get_image_animation_description(self, animation: LEDImageAnimation):
 
         return animation_description
 
+    def _update_animations(self):
+
+        user_animations = rospy.get_param('~user_animations', '')
```
Kotochleb, line 267: I am kinda lost right now since this parameter is not documented in README

panther_lights_controller/src/controller_node.py
```diff
@@ -251,6 +262,24 @@ def _get_image_animation_description(self, animation: LEDImageAnimation):
...
+
+        for animation in user_animations:
+            # ID numbers from 0 to 20 are reserved for system animations
+            if animation['id'] > 20:
```
Kotochleb: ID is not checked if it is positive value

Kotochleb (changes requested):
In general, the documentation still need tweaking.
Section about custom parameters does not specify which fields are available for user after inheriting from `Animation` class or that animation is processed in tics.

I will opt for `$(find package_name)` way of specifying image paths.

It is unclear whether duration applies only to an image, or the sequence of repeating images as a whole. For example, if `duration=1` and `repeat=3` will result in 3 times showing same animation that is 1 second long (overall 3 seconds of a whole animation) or in 1 second long sequence or 3 times same animation, where single pass takes 1/3 of a second.

I am also not sure if we should keep the `color` key in image animation. It is already implemented, but I'm uncertain if it is a good idea to give this ability to end user to change animation color.

panther_lights_controller/src/controller_node.py
KmakD: If ID is greater than 20 then it is positive

panther_lights_controller/README.md
KmakD: I used 100Hz to see if everything works in high speed. Definitely too high, will be changed when fixing lights node parameters.

Kotochleb: Ok. Makes sense

panther_lights_controller/README.md
```diff
@@ -14,37 +14,173 @@ Package used to control the Husarion Panther LED panels.
 
 #### Parameters
 
-- `~animations` [*list*]: it is a required ROS parameter containing a list of defined animations.
-- `~global_brightness` [*float*, default: **1.0**]: LED global brightness.
-- `~num_led` [*int*, default: **46**]: number of LEDs in single panel.
+- `~animations` [*list*, default: None]: it is a required ROS parameter containing a list of defined animations.
```
Kotochleb:
```suggestion
- `~animations` [*list*, default: **None**]:  required list of defined animations.
```

panther_lights_controller/README.md
```diff
@@ -14,37 +14,173 @@ Package used to control the Husarion Panther LED panels.
...
 - `~controller_frequency` [*float*, default: **100**]: frequency at which the lights controller node will process animations.
-- `~test` [*bool*, default: **false**]: allows testing mode with some extra functionalities.
+- `~global_brightness` [*float*, default: **1.0**]: LED global brightness. Range between [0,1].
+- `~num_led` [*int*, default: **46**]: number of LEDs in single panel.
+- `~test` [*bool*, default: **false**]: enables testing mode with some extra functionalities.
+- `~user_animations` [*list*, default: None]: it is an optional ROS parameter containing a list of animations defined by the user.
```
Kotochleb:
```suggestion
- `~user_animations` [*list*, default: **None**]: optional list of animations defined by the user.
```
I'm a bit lost right now how are we supposed to define it outside of YAML. But it is ok nevertheless

panther_lights_controller/src/controller_node.py
Kotochleb: IF you put it this way then yes. It is

panther_lights_controller/src/animation/image_animation.py
```diff
@@ -1,6 +1,7 @@
 import imageio
 import numpy as np
 import os
+import re
 from PIL import Image
```
Kotochleb, line 4: Letter "P" is before "r"

panther_lights_controller/README.md
```diff
@@ -14,37 +14,173 @@ Package used to control the Husarion Panther LED panels.
...
+
+- `animation` [*dict*]: definition of animation. See section below for more info.
 - `id` [*int*]: ID of an animation.
-- `animation` [*dict*]: description of animation.
-- `name` *(optional)* [*string*]: name of an animation.
-- `interrupting` *(optional)* [*bool*]: if *true* animation will interrupt currently displayed animation.
+- `interrupting` [*bool*, optional]: if *true* animation will interrupt currently displayed animation.
+- `name` [*string*, optional]: name of an animation.
 
 ### ImageAnimation
```
Kotochleb: There is rough transition between default animations and animation class. An additional subsection saying something like "AnimationTypes" or something like this would give the needed introduction to the topic

panther_lights_controller/README.md
```diff
@@ -14,37 +14,173 @@ Package used to control the Husarion Panther LED panels.
...
+
+1. Add docker volume with a previously created animation description list. If using custom images for `ImageAnimation`, add also a docker volume with a folder containing custom images. On Raspberry Pi modify `compose.yaml`:
+
+```yaml
+volumes:
+  - ./user_animations.yaml:/user_animations.yaml
+  - ./animations:/animations
+```
+
+**Warning**
```
Kotochleb:
```suggestion
:warning: **Warning**
```
This will look even more warningly

panther_lights_controller/README.md
```diff
@@ -14,37 +14,173 @@ Package used to control the Husarion Panther LED panels.
...
+- `repeat` [*int*, optional]: number of times the animation will be repeated, by default animation will run once.
 - `type` [*string*]: required field specyfying animation type, for `ImageAnimation` value should be `image_animation`.
-- `image` [*string*]: path to an image file, if a path is not global will look in the `animations` folder for the given name.
-- `duration` [*float*]: duration of an animation.
-- `repeat` *(optional)* [*int*]: number of times the animtion will be repeated, by default animation will run once.
-- `brightness` *(optional)* [float]: animation brightness, this will be used instead of controller `global_brightness`.
-- `color` *(optional)* [*int*]: if set the image color will be changed to the specified one, advised way is to use HEX format eg. `0xff0000`.
 
-### Defining new animation
+**NOTE:** The overall display duration of an animation is a product of a single image duration and repeat count. It can't exceed 10 seconds.
```
Kotochleb:
```suggestion
:bulb: **NOTE:** The overall display duration of an animation is a product of a single image duration and repeat count. It can't exceed 10 seconds.
```
It might be added, might be omitted

panther_lights_controller/README.md
```diff
@@ -14,37 +14,173 @@ Package used to control the Husarion Panther LED panels.
...
+
+### Defining animations
+
+User can define own animations using basic animation types. Similar to basic ones user animations are parsed using a ROS parameter `/panther/lights/lights_controller_node/user_animations`. They can be loaded on node start or updated using the `/panther/lights/controller/update_animations` ROS service. For `ImageAnimation` you can use basic images from the `animations` folder and change their color with the `color` key ([see ImageAnimation](#imageanimation)). Follow the example below to add custom animations. 
+
+1. Create a yaml file with an animation description list. Example file: 
+
+**NOTE:** ID numbers from 0 to 19 are reserved for system animations.
```
Kotochleb:
```suggestion
:bulb: **NOTE:** ID numbers from 0 to 19 are reserved for system animations.
```
Same as above

panther_lights_controller/README.md
```diff
@@ -14,37 +14,173 @@ Package used to control the Husarion Panther LED panels.
...
+docker compose up -d
+```
+
+5. Display new animations:
+
+```bash
+rosservice call /lights/controller/set/animation "{animation: {id: 21, name: 'ANIMATION_1'}, repeating: false}"
+```
+
+#### Updating animation list at a runtime
```
Kotochleb:
```suggestion
### Updating animation list at a runtime
```
I think this can be a separate step

panther_lights_controller/README.md
```diff
@@ -14,37 +14,173 @@ Package used to control the Husarion Panther LED panels.
...
+- `controller_freq` [*float*]: controller frequency at which animation frames will be processed.
+
+Methods:
+
+- `__call__` - by default it is not implemented and must be overwritten. It returns an animation frame as a list of integers. It should also set the `self._finished` variable `true` when the last animation frame is returned.
+- `reset` - by default it is not implemented. It allows resetting animation.
+
+Properties:
+
+- `brightness` [*int]: returns animation brightness from `self._brightness`.
```
Kotochleb:
```suggestion
- `brightness` [*int*]: returns animation brightness from `self._brightness`.
```

panther_lights_controller/README.md
```diff
@@ -14,37 +14,173 @@ Package used to control the Husarion Panther LED panels.
...
+- `num_led` [*int*]: returns number of LEDs in panel from `self._num_led`.
+- `finished` [*bool*]: returns if animation execution is finished from `self._finished`.
 
-It is possible to define your own animation with a required behavior. All animation definitions are stored in `/src/animation` and inherit from the basic class `Animation`. Each animation must overwrite the basic method `__call__` which must return an animation frame as a list of integers. An animation should also contain `ANIMATION_NAME` used to identify it. Animation frames are displayed in the controller's main timer. To tweak animation duration time use the `controller_freq` and `_duration` variables. For an example see other animation definitions.
+The new animation definition should contain `ANIMATION_NAME` used to identify it. Animation frames are processed in ticks with a frequency of `controller_freq`. To tweak animation duration time use the `controller_freq` and `_duration` variables. For an example see other animation definitions.
```
Kotochleb:
```suggestion
The new animation definition should contain `ANIMATION_NAME` used to identify it. Animation frames are processed in ticks with a frequency of `controller_freq`. To tweak animation duration time use the `controller_freq` and `_duration` variables. As an example see other animation definitions.
```

Kotochleb (changes requested):
I would add headers to the steps with number, so they can stand more from the rest of the text. 

Later when polishing the whole document I would suggest discussing moving the guide on defining own animation to a separate readme file so that the main readme will be an interface to the animations while the guide will have its own file.

Kotochleb (approved):

PR #45, opened 2023-01-04
Priority queue

KmakD:


panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
 
 class PantherAnimation:
+    ANIMATION_DEFAULT_PRIORITY = 3
+    ANIMATION_DEFAULT_TIMEOUT = 120
+
     front: Animation
     rear: Animation
-    interrupting = False
-    repeating = False
+    name: str = 'NAME_NOT_DEFINED'
```
Kotochleb:
```suggestion
    name: str = 'UNDEFINED'
```

Wouldn't it look better like this?

panther_lights/README.md
```diff
@@ -63,8 +63,9 @@ Default animations are described and loaded on the node start, directly from `co
 
 - `animation` [*dict*]: definition of animation. See section below for more info.
 - `id` [*int*]: ID of an animation.
-- `interrupting` [*bool*, optional]: if *true* animation will interrupt currently displayed animation.
-- `name` [*string*, optional]: name of an animation.
+- `name` [*string*, default: **NAME_NOT_DEFINED**]: name of an animation.
+- `priority` [*int*, default: **3**]: priority at which animation will be placed in the queue. The lower the number the higher animation priority. Priority is an integer value in range **<1,3>**. Animations with priority 1 will also remove animations with lower priority from the queue. An animation entering the queue with higher priority than currently didplayed will interrupt it and put back in front of the queue.
+- `timeout` [*float*, default: **120.0**]: time in seconds after which animation will be romoved from the queue.
```
Kotochleb:
```suggestion
- `timeout` [*float*, default: **120.0**]: time in seconds after which animation will be removed from the queue.
```

panther_lights/README.md
```diff
@@ -63,8 +63,9 @@ Default animations are described and loaded on the node start, directly from `co
 
 - `animation` [*dict*]: definition of animation. See section below for more info.
 - `id` [*int*]: ID of an animation.
-- `interrupting` [*bool*, optional]: if *true* animation will interrupt currently displayed animation.
-- `name` [*string*, optional]: name of an animation.
+- `name` [*string*, default: **NAME_NOT_DEFINED**]: name of an animation.
+- `priority` [*int*, default: **3**]: priority at which animation will be placed in the queue. The lower the number the higher animation priority. Priority is an integer value in range **<1,3>**. Animations with priority 1 will also remove animations with lower priority from the queue. An animation entering the queue with higher priority than currently didplayed will interrupt it and put back in front of the queue.
```
Kotochleb:
```suggestion
- `priority` [*int*, default: **3**]: priority at which animation will be placed in the queue. List below shows behaviour when ne animation with given ID arrives:
    - **1** intterupts and removes animation with priorites **2** and **3**.
    - **2** interrupts animations with priority **3**.
    - **3** add adnimation to the end of queue.
```

panther_lights/README.md
```diff
@@ -200,16 +206,18 @@ Arguments:
 
 Methods:
 
-- `__call__` - by default it is not implemented and must be overwritten. It returns an animation frame as a list of integers. It should also set the `self._finished` variable `true` when the last animation frame is returned.
-- `reset` - by default it is not implemented. It allows resetting animation.
+- `__call__` - It returns an animation frame as a list of integers, updates the `progress` variable, and sets the `finished` flag when the animation execution is finished. It uses the `_update_animation` method to get the current animation frame.
+- `reset` - It allows resetting animation to its initial state.
```
Kotochleb:
```suggestion
- `reset` - resets animation to initial state. If overwritten requires calling parent class implementation first.
```

panther_lights/README.md
```diff
@@ -200,16 +206,18 @@ Arguments:
...
+- `_update_animation` - by default it is not implemented. This method is used to update the animation frame.
```
Kotochleb:
```suggestion
- `_update_animation` - returns list of integers of lenght `num_led` with colors for LED panel to be displayed. Colors are stored in **HEX** format as integers. By default not implemented.
```

panther_lights/src/animation/animation.py
```diff
@@ -34,11 +36,35 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
...
+
+    def __call__(self) -> list:
+        if self._current_cycle <= self._loops:
+            frame = self._update_frame()
+            self._anim_iteration += 1
+            self._progress = (
+                self._anim_iteration + self._anim_len * (self._current_cycle - 1)
+            ) / self._full_anim_len
+
+            if self._anim_iteration == self._anim_len:
```
Kotochleb:
```suggestion
            if self._anim_iteration >= self._anim_len:
```
Just for the sake of standardization

panther_lights/src/animation/animation.py
```diff
@@ -34,11 +36,35 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
                 raise KeyError('Brightness has to match boundaries 0 < brightness <= 1')
             self._brightness = int(round(self._brightness * 100))
 
-    def __call__(self) -> None:
-        '''returns new frame'''
-        raise NotImplementedError
+        # evaluate number of animation frames
+        self._anim_len = int(round(self._duration * controller_freq))
```
Kotochleb, line 36: I would add another check. If animation `_anim_len ` smaller than 1 there should be an info that animation is too short to display with given frequency

panther_lights/README.md
```diff
@@ -200,16 +206,18 @@ Arguments:
 
 Methods:
 
-- `__call__` - by default it is not implemented and must be overwritten. It returns an animation frame as a list of integers. It should also set the `self._finished` variable `true` when the last animation frame is returned.
-- `reset` - by default it is not implemented. It allows resetting animation.
+- `__call__` - It returns an animation frame as a list of integers, updates the `progress` variable, and sets the `finished` flag when the animation execution is finished. It uses the `_update_animation` method to get the current animation frame.
```
Kotochleb: Since eventually this method is not a part of an API I think we can not document it here

panther_lights/src/animation/animation.py
```diff
@@ -34,11 +36,35 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
...
+
+            if self._anim_iteration == self._anim_len:
+                self._anim_iteration = 0
+                self._current_cycle += 1
+
+            if self._current_cycle > self._loops:
+                self._finished = True
+
+            return frame
+        raise Animation.AnimationFinished
```
Kotochleb: After all this time, I'm starting to think that this should be generator rather than this implementation with exceptions

panther_lights/src/animation/image_animation.py
```diff
@@ -41,11 +41,8 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
 
         # resize image to match duration
         original_img = imageio.imread(img_path)
```
Kotochleb:
```suggestion
        original_img = Image.open(img_path)
```
This would save dependency on `imageio`

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+
+        if len(self._queue) == self._max_queue_size:
+            rospy.logwarn(f'{rospy.get_name()} Animation queue overloaded')
+            return
+
+        if put_front:
+            self._queue.insert(0, animation)
+        else:
+            self._queue.append(animation)
+        self._queue.sort(key=self._get_priority)
```
Kotochleb:
```suggestion
        self._queue.sort(key=lambda x.priority)
```
I think this would be a better way. Also aren't we planning to implement other option to for example sort by time left for animation as well?

panther_lights/src/controller_node.py
```diff
@@ -112,14 +189,30 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
 
     def _controller_timer_cb(self, *args) -> None:
         with self._lock:
```
Kotochleb: Why there is still this lock since this is only single thread and it is used only once?

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+        return len(self._queue) == 0
+
+    def clear(self) -> None:
+        # clears queue except items with priority 1
+        remove_animation_list = []
+        for animation in self._queue:
+            if animation.priority > 1:
+                remove_animation_list.append(animation)
+        for animation in remove_animation_list:
+            self._queue.remove(animation)
```
Kotochleb: I would either rename the function to `clear_non_crutial` or add priority argument and the function would remove all animations with given and lower priority. I know this is slightly overshot, but this would logically make more sense

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+        return False
+
+    def validate_queue(self) -> None:
+        remove_animation_list = []
+        for animation in self._queue:
+            if rospy.get_time() - animation.init_time > animation.timeout:
+                remove_animation_list.append(animation)
+                rospy.logwarn(
+                    f'{rospy.get_name()} Timeout for animation: {animation.name}. Romoving from the queue'
+                )
```
Kotochleb:
```suggestion
                rospy.loginfo(
                    f'{rospy.get_name()} Timeout for animation: {animation.name}. Romoving from the queue'
                )
```

Personally, I feel more like this should be info rather than warning. Nothing bad can happen if once in a while animation will timeout

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+    def init_time(self) -> float:
+        return self._init_time
+
+
+class AnimationsQueue:
+    def __init__(self, max_queue_size) -> None:
+        self._queue = []
+        self._max_queue_size = max_queue_size
+
+    def put(self, animation: PantherAnimation, put_front: int = False) -> None:
```
Kotochleb:
```suggestion
    def put(self, animation: PantherAnimation, put_front: bool = False) -> None:
```

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+                return True
+        return False
+
+    def validate_queue(self) -> None:
+        remove_animation_list = []
+        for animation in self._queue:
+            if rospy.get_time() - animation.init_time > animation.timeout:
+                remove_animation_list.append(animation)
+                rospy.logwarn(
+                    f'{rospy.get_name()} Timeout for animation: {animation.name}. Romoving from the queue'
```
Kotochleb:
```suggestion
                    f'{rospy.get_name()} Timeout for animation: {animation.name}. Removing from the queue'
```

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+                )
+        for animation in remove_animation_list:
+            self._queue.remove(animation)
```
Kotochleb:
```suggestion
t = rospy.get_time()
self._queue = [anim if t - anim.init_time < anim.timeout
               else f'{rospy.get_name()} Timeout for animation: {anim.name}. Removing from the queue'
               for anim in self._queue]
```
I am quite sure this abomination should be faster

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+        if not self.empty():
+            return self._queue[0].priority
+        else:
+            return PantherAnimation.ANIMATION_DEFAULT_PRIORITY
+
+    def has_animation(self, animation: PantherAnimation) -> bool:
+        for anim in self._queue:
+            if anim == animation:
+                return True
+        return False
```
Kotochleb:
```suggestion
    def has_animation(self, animation: PantherAnimation) -> bool:
        return animation in self._queue
```

Comparing two objects will result in checking if they are indeed the same object. For assumption given animation can exist in a queue only once, this will work. If we can store copies of animations in queue we should compare them by ID as follows:
``` python
    def has_animation(self, animation: PantherAnimation) -> bool:
        return animation.id in [anim.id for anim in self._queue]
```

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+
+    def remove(self, animation) -> None:
+        if self.has_animation(animation):
+            self._queue.remove(animation)
+
+    def get_next_anim_priority(self) -> int:
+        if not self.empty():
+            return self._queue[0].priority
+        else:
+            return PantherAnimation.ANIMATION_DEFAULT_PRIORITY
```
Kotochleb: If the animation is supposed to be next, then why queue element is `0`?

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+class AnimationsQueue:
+    def __init__(self, max_queue_size) -> None:
+        self._queue = []
+        self._max_queue_size = max_queue_size
+
+    def put(self, animation: PantherAnimation, put_front: int = False) -> None:
+        self.validate_queue()
+
+        if animation.priority == 1:
+            self.clear()
```
Kotochleb:
```suggestion
        if animation.priority == 1:
            self.clear()
        self.validate_queue()
```
If most of the animations will be removed, validation will take less time

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,89 @@
...
+
+        if len(self._queue) == self._max_queue_size:
+            rospy.logwarn(f'{rospy.get_name()} Animation queue overloaded')
+            return
+
+        if put_front:
+            self._queue.insert(0, animation)
+        else:
+            self._queue.append(animation)
```
Kotochleb: If there are two critical animations in queue, a new animation at the same priority will be inserted in front of them. This will lead to incorrect time of displaying animations with the same priority. Animation with critical priority inserted even at the very end will possibly work its way to the front with sort function. If sorting will base also on time of lift, all animations should be displayed in correct order. With all that `put_front` parameter is not necessary

panther_lights/src/controller_node.py
```diff
@@ -112,14 +189,30 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
...
-                    self._current_animation = self._default_animation
-                else:
-                    self._current_animation = self._anim_queue.get(block=False)
-                    self._interrupt = False
+            if self._animation_finished:
+                self._anim_queue.validate_queue()
+
+                if self._default_animation and not self._anim_queue.has_animation(
+                    self._default_animation
+                ):
```
Kotochleb:
```suggestion
                if self._default_animation and not
                    self._anim_queue.has_animation(self._default_animation):
```

This is more of a question if this isn't more readable

panther_lights/src/controller_node.py
```diff
@@ -112,14 +189,30 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
...
 
             if self._current_animation:
+                if self._current_animation.priority > self._anim_queue.get_next_anim_priority():
+                    self._current_animation.front.reset()
+                    self._current_animation.rear.reset()
+                    if (
+                        self._current_animation.front.progress < 0.9
+                        and not self._current_animation.repeating
+                    ):
+                        self._anim_queue.put(self._current_animation, put_front=True)
```
Kotochleb: Resting the animation will reset its progress, so the `if` statement will always be true

Kotochleb: If we sorted by lifespan of animation that is left for them, I think this will result with this animation nevertheless ending up in the front of the queue since it would be the oldest animation with this priority in the queue

panther_lights/src/controller_node.py
```diff
@@ -112,14 +189,30 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
...
+                    self._current_animation = self._anim_queue.get()
 
             if self._current_animation:
+                if self._current_animation.priority > self._anim_queue.get_next_anim_priority():
+                    self._current_animation.front.reset()
+                    self._current_animation.rear.reset()
+                    if (
+                        self._current_animation.front.progress < 0.9
+                        and not self._current_animation.repeating
+                    ):
```
Kotochleb: I don't think the brackets are necessary

panther_lights/README.md
```diff
@@ -200,16 +206,18 @@ Arguments:
...
+- `reset` - It allows resetting animation to its initial state.
+- `_update_animation` - by default it is not implemented. This method is used to update the animation frame.
 
 Properties:
 
 - `brightness` [*int*]: returns animation brightness from `self._brightness`.
 - `num_led` [*int*]: returns number of LEDs in panel from `self._num_led`.
 - `finished` [*bool*]: returns if animation execution is finished from `self._finished`.
+- `progress` [*float*]: returns animation execution progress from `self._progress`.
```
Kotochleb: I think only `num_led` would be important for the end user. `brightness`, `finished` and `progress` are computed in parent class and are not left for the user to be part of the interface

Kotochleb (changes requested):
I think implementing animations in a way that they are parsed and created when `_update_animations` is called and then stored inside dict might simplify and speedup the code. This approach will simplify adding animations by ID since the ID can be the key to the dict. Dict can store objects of class `PantherAnimation`. ` _get_animation_by_id` can be later simplified to referencing object from dict and copying it with `deep_copy`.

I am a bit concerned about current implementations problems in case of two same animations waiting in a queue at the same time. This approach I think might remove all the problems with duplicates.

panther_lights/src/controller_node.py
KmakD: Actually, new animation with the same priority will be inserted at the back so order of display is correct, but if we assume sorting by time that does not matter and put_front is not necessary.

panther_lights/src/controller_node.py
KmakD: black auto formatting

panther_lights/src/controller_node.py
KmakD: black auto formatting

panther_lights/src/controller_node.py
KmakD: This method was created to search for a specific animation. Eg. you can have 2 animations with the same ID but one is also a default animation (repeating one)

panther_lights/src/controller_node.py
KmakD: That won't work if we want to keep log info. In this case `else` in this form will put `str` into queue list

panther_lights/src/controller_node.py
Kotochleb: k

panther_lights/src/controller_node.py
Kotochleb: k

panther_lights/src/controller_node.py
```diff
@@ -233,55 +375,36 @@ def _get_animation_by_id(self, animation_id: int) -> PantherAnimation:
...
+                    timeout = anim['timeout']
+                    if timeout <= 0:
+                        rospy.logwarn(
+                            f'{rospy.get_name} Invalid timeout for animation: {animation.name}. Using default'
+                        )
+                        timeout = PantherAnimation.ANIMATION_DEFAULT_TIMEOUT
+                    animation.timeout = timeout
 
-        user_animations = rospy.get_param('~user_animations', '')
+                animations.append(dict(id=anim['id'], animation=animation))
```
Kotochleb:
```suggestion
                animations[anim['id']] = animation
```
I meant something like this

panther_lights/src/controller_node.py
```diff
@@ -189,15 +270,76 @@ def _set_image_animation_cb(
...
+        try:
+            self._update_animations()
+        except (KeyError, FileNotFoundError) as err:
+            return TriggerResponse(False, f'Failed to update animations: {err}')
         return TriggerResponse(True, 'Animations updated successfully')
 
     def _get_animation_by_id(self, animation_id: int) -> PantherAnimation:
+        for animation in self._animations:
+            if animation['id'] == animation_id:
+                return deepcopy(animation['animation'])
```
Kotochleb:
```suggestion
        return deepcopy(self._animations[animation_id])
```
The dict based implementation will simplify this that much

panther_lights/src/controller_node.py
```diff
@@ -111,43 +177,59 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
...
+            if self._current_animation.priority > self._anim_queue.get_first_anim_priority():
+                if (
+                    self._current_animation.front.progress < 0.9
+                    and not self._current_animation.repeating
+                ):
+                    self._anim_queue.put(self._current_animation)
+                self._current_animation.front.reset()
+                self._current_animation.rear.reset()
+                self._animation_finished = True
+                self._current_animation = None
```
Kotochleb:
```suggestion
                else:
                    del self._current_animation
                    self._current_animation = None
```
Since we know we won't need this object in the future, and it is a copy of the original object, we can just explicitly delete it and not care about resetting it

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,78 @@
...
+            self.clear()
+        self.validate_queue()
+
+        if len(self._queue) == self._max_queue_size:
+            rospy.logwarn(f'{rospy.get_name()} Animation queue overloaded')
+            return
+
+        self._queue.append(animation)
+        self._queue.sort(key=lambda x: x.init_time)
+        self._queue.sort(key=lambda x: x.priority)
```
Kotochleb:
```suggestion
        self._queue = sorted(self._queue, key=lambda x: (x.init_time, x.priority))
```
Won't it be more pythonic?

Kotochleb (changes requested):
This is not a full review yet, but those are the most important changes currently

panther_lights/src/controller_node.py
```diff
@@ -111,43 +176,60 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
...
+                self._current_animation = self._anim_queue.get()
+
+        if self._current_animation:
+            if self._current_animation.priority > self._anim_queue.get_first_anim_priority():
+                if (
+                    self._current_animation.front.progress < 0.9
+                    and not self._current_animation.repeating
+                ):
                     self._current_animation.front.reset()
                     self._current_animation.rear.reset()
```
Kotochleb, line 199:
```suggestion
```
Resetting the animation before putting it back to the queue won't cause it to continue being animated, but will reset it even at the end of its life

panther_lights/src/controller_node.py
```diff
@@ -111,43 +176,60 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
...
-                    self._current_animation = None
+                    self._anim_queue.put(self._current_animation)
+                del self._current_animation
```
Kotochleb: Since everything in python is a pointer, won't it cause animation object to be deleted even after putting it back to queue

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,77 @@
...
+    def validate_queue(self) -> None:
+        remove_animation_list = []
+        for animation in self._queue:
+            if rospy.get_time() - animation.init_time > animation.timeout:
+                remove_animation_list.append(animation)
+                rospy.loginfo(
+                    f'{rospy.get_name()} Timeout for animation: {animation.name}. Romoving from the queue'
+                )
+        for animation in remove_animation_list:
+            self._queue.remove(animation)
```
Kotochleb:
```suggestion
        time = rospy.get_time()
        remove_animation_list = [anim if time - anim.init_time > anim.timeout for anim in self._queue]
        for anim in remove_animation_list:
            rospy.loginfo(
                f'{rospy.get_name()} Timeout for animation: {anim.name}. Romoving from the queue'
            )
            self._queue.remove(anim)
```
How bout this one?

panther_lights/src/controller_node.py
```diff
@@ -111,43 +176,60 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
...
+                )
+
+            self._animation_finished = (
+                self._current_animation.front.finished and self._current_animation.rear.finished
+            )
+
+            if self._animation_finished:
+                self._current_animation.front.reset()
+                self._current_animation.rear.reset()
+                self._current_animation = None
```
Kotochleb, line 225: Animation here might also be removed manually

panther_lights/src/controller_node.py
```diff
@@ -189,15 +270,78 @@ def _set_image_animation_cb(
...
+            return TriggerResponse(False, f'Failed to update animations: {err}')
         return TriggerResponse(True, 'Animations updated successfully')
 
     def _get_animation_by_id(self, animation_id: int) -> PantherAnimation:
+        try:
+            animation = deepcopy(self._animations[animation_id])
+            animation.reset_time()
+            return animation
+        except KeyError:
+            raise ValueError(f'No Animation with id: {animation_id}')
```
Kotochleb: Since this function is used only once. Is there a point of having it? IMO it will be equally readable if we will do it inside service callback. But I won't insist on it

panther_lights/src/controller_node.py
```diff
@@ -111,43 +176,60 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
...
                     self._current_animation.rear.reset()
-                    self._current_animation = None
+                    self._anim_queue.put(self._current_animation)
+                del self._current_animation
+                self._animation_finished = True
+                self._current_animation = None
+                return
+
+            if not self._current_animation.front.finished:
+                frame_front = [led for led in self._current_animation.front()]
```
Kotochleb: The conversion can be done in call function by calling `return self._img[self._anim_iteration, :].tolist()`

panther_lights/src/controller_node.py
```diff
@@ -111,43 +176,60 @@ def _percent_to_apa_driver_brightness(self, brightness_percent) -> int:
...
+                brightness_front = self._current_animation.front.brightness
+                self._controller.set_panel_frame(
+                    LightsControllerNode.PANEL_FRONT, frame_front, brightness_front
+                )
+            if not self._current_animation.rear.finished:
+                frame_rear = [led for led in self._current_animation.rear()]
+                brightness_rear = self._current_animation.rear.brightness
+                self._controller.set_panel_frame(
+                    LightsControllerNode.PANEL_REAR, frame_rear, brightness_rear
+                )
```
Kotochleb: Without all the list comprehension and stuff, this could be reduced to evaluation of the variables in a function call

panther_lights/src/controller_node.py
```diff
@@ -189,15 +270,78 @@ def _set_image_animation_cb(
...
+
+        animation_description = {
+            'image': animation.image,
+            'duration': animation.duration,
+            'repeat': animation.repeat,
+            'brightness': animation.brightness,
+        }
+
+        if animation.color:
+            animation_description.update({'color': animation.color})
```
Kotochleb:
```suggestion
            animation_description['color'] = animation.color
```
This is more normal way of specifying this I think

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,77 @@
...
+
+        if len(self._queue) == self._max_queue_size:
+            rospy.logwarn(f'{rospy.get_name()} Animation queue overloaded')
+            return
+
+        self._queue.append(animation)
+        self._queue = sorted(self._queue, key=lambda x: (x.priority, x.init_time))
+
+    def get(self) -> PantherAnimation:
+        return self._queue.pop(0)
```
Kotochleb, line 62:
```suggestion
    def pop(self) -> PantherAnimation:
        return self._queue.pop(0)
```
I think `pop` might more descriptive since it will tell exactly what it does under the hood

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,77 @@
...
+
+    def clear(self, priority: int = 2) -> None:
+        # clears queue except items below specified priority
+        self._queue = [animation for animation in self._queue if animation.priority < priority]
+
+    def remove(self, animation) -> None:
+        if self.has_animation(animation):
+            self._queue.remove(animation)
+
+    def get_first_anim_priority(self) -> int:
```
Kotochleb:
```suggestion
    def first_anim_priority(self) -> int:
```
Since this is only function with `get` prefix I think it can be skipped. Also this could be an property

panther_lights/src/controller_node.py
```diff
@@ -233,55 +377,36 @@ def _get_animation_by_id(self, animation_id: int) -> PantherAnimation:
...
-        return animation_description
+                if 'name' in anim:
+                    animation.name = anim['name']
+
+                if 'priority' in anim:
+                    priority = anim['priority']
+                    if (
+                        not 0 < priority <= PantherAnimation.ANIMATION_DEFAULT_PRIORITY
+                        or not isinstance(priority, int)
+                    ):
```
Kotochleb:
```suggestion
                    if (
                        not 0 < priority <= PantherAnimation.ANIMATION_DEFAULT_PRIORITY
                    ):
```
The rest of the code doesn't care what it gets from YAML so either the rest should check type, or this piece should not

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,77 @@
...
+    def reset_time(self) -> None:
+        self._init_time = rospy.get_time()
+
+    @property
+    def init_time(self) -> float:
+        return self._init_time
+
+
+class AnimationsQueue:
+    def __init__(self, max_queue_size) -> None:
```
Kotochleb:
```suggestion
    def __init__(self, max_queue_size: int = 5) -> None:
```

panther_lights/src/controller_node.py
```diff
@@ -21,10 +20,77 @@
...
+        return self._queue.pop(0)
+
+    def empty(self) -> bool:
+        return len(self._queue) == 0
+
+    def clear(self, priority: int = 2) -> None:
+        # clears queue except items below specified priority
+        self._queue = [animation for animation in self._queue if animation.priority < priority]
+
+    def remove(self, animation) -> None:
```
Kotochleb:
```suggestion
    def remove(self, animation: PantherAnimation) -> None:
```

panther_lights/src/animation/animation.py
```diff
@@ -52,3 +85,7 @@ def num_led(self) -> int:
     @property
     def finished(self) -> bool:
         return self._finished
+
+    @property
+    def progress(self) -> bool:
```
Kotochleb:
```suggestion
    def progress(self) -> flaot:
```

panther_lights/src/animation/animation.py
```diff
@@ -9,6 +11,8 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
         self._num_led = num_led
         self._current_cycle = 1
```
Kotochleb:
```suggestion
        self._current_cycle = 0
```
Maybe iterate cycles from 0?

panther_lights/src/animation/animation.py
```diff
@@ -34,11 +38,40 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
                 raise KeyError('Brightness has to match boundaries 0 < brightness <= 1')
```
Kotochleb:
```suggestion
                raise KeyError('Brightness has to be in range <0,1>)
```

panther_lights/src/animation/animation.py
```diff
@@ -34,11 +38,40 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
...
             self._brightness = int(round(self._brightness * 100))
```
Kotochleb, line 33: I would implement here the same logic as in `_percent_to_apa_driver_brightness`

panther_lights/src/animation/animation.py
```diff
@@ -34,11 +38,40 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
...
-        '''returns new frame'''
-        raise NotImplementedError
+        # evaluate number of animation frames
+        self._anim_len = int(round(self._duration * controller_freq))
+        self._full_anim_len = self._anim_len * self._loops
+
+        if self._anim_len < 1:
+            raise KeyError('Animation duration is too short to display with the current frequency')
+
+    def __call__(self) -> Generator:
```
Kotochleb:
```suggestion
    def __call__(self) -> List:
```

panther_lights/src/animation/animation.py
```diff
@@ -34,11 +38,40 @@ def __init__(self, animation_description: dict, num_led: int, controller_freq: f
...
+
+            if self._anim_iteration >= self._anim_len:
+                self._anim_iteration = 0
+                self._current_cycle += 1
+
+            if self._current_cycle > self._loops:
+                self._finished = True
+
+            for led in frame:
+                yield led
```
Kotochleb:
```suggestion
            frame = self._update_frame()
            self._anim_iteration += 1
            self._progress = (
                self._anim_iteration + self._anim_len * (self._current_cycle - 1)
            ) / self._full_anim_len

            if self._anim_iteration > self._anim_len:
                self._anim_iteration = 0
                self._current_cycle += 1

            if self._current_cycle > self._loops:
                self._finished = True
                
            return frame
        return [0] * self._num_led
```

After conversation, let's give up the idea of generators and return empty frame in case of end of animation.

panther_lights/src/animation/animation.py
```diff
@@ -1,3 +1,5 @@
+from collections.abc import Generator
```
Kotochleb:
```suggestion
```

It can be removed. We are not using generators eventually

Kotochleb (changes requested):
There is partially missing type hinting in `__init__` of LightsControllerNode. SAme goes for  `_percent_to_apa_driver_brightness` and maybe for `_controller_timer_cb`, not sure on this one.

in `image_animation.py` the object parameters could be reduced to:
``` python
def __init__(self, *args, **kwargs) -> None:
        super().__init__(self, *args, **kwargs)
```

panther_lights/src/animation/animation.py
Kotochleb: Anyway that was bad idea

panther_lights/src/controller_node.py
KmakD: I wanted to make custom queue to be similar to standard python `Queue` which has get method. I think it makes more sense and is more intuitive for people who used standard queue before.

panther_lights/src/animation/animation.py
KmakD: According to apa102 driver global_brightnes is an int in range <0, 31> like in `_percent_to_apa_driver_brightness`, but brightness passed to `set_pixel_rgb` is percentage value range 0-100

Kotochleb: k

panther_lights/src/controller_node.py
Kotochleb: Ok. Makes sense

panther_lights/README.md
Kotochleb: Anyway README will be updated later

panther_lights/src/controller_node.py
Kotochleb: Anyway it's fine as it is. I'm resolving this thread

Kotochleb (approved):
Finally, looks good

Kotochleb (approved):

PR #48, opened 2023-01-16
Battery animation

KmakD:


panther_lights/src/animation/battery_animation.py
```diff
@@ -0,0 +1,74 @@
...
+        self._param = value
+        self._create_anim(value)
+
+    def _create_anim(self, battery_percent: float) -> None:
+        resolution = 100
+        # define basic HSV color
+        h = self._h_min
+        s = 1.0
+        v = 1.0
+        rgb = np.array(hsv_to_rgb(h / 360, s, v))
```
Kotochleb: This variable is unused

panther_lights/src/animation/battery_animation.py
```diff
@@ -0,0 +1,74 @@
...
+        display_iterations = battery_percent * resolution * 0.9
+
+        if percent_point < 1:
+            percent_point = 1
+            display_iterations = 0
+
+        # create binary animation
+        for i in range(resolution):
+            if i <= display_iterations:
+                anim_ind = round(i / resolution * self._num_led * 1 / 0.9)
```
Kotochleb:
```suggestion
                anim_ind = round(i / resolution * self._num_led / 0.9)
```

panther_lights/src/animation/battery_animation.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def _update_frame(self) -> list:
+        return self._anim[self._anim_iteration, :].tolist()
+
+    def set_param(self, value: float) -> None:
+        self._param = value
+        self._create_anim(value)
+
+    def _create_anim(self, battery_percent: float) -> None:
+        resolution = 100
```
Kotochleb: I would move this to constructor and keep it as `self` variable

panther_lights/src/animation/battery_animation.py
```diff
@@ -0,0 +1,74 @@
...
+                elif percent_point <= anim_ind < 2 * percent_point:
+                    ind_1 = int(np.ceil(self._num_led / 2 - 2 * percent_point + int(anim_ind)))
+                    ind_2 = int(np.floor(self._num_led / 2 + 2 * percent_point - int(anim_ind) - 1))
+                    frame[ind_1] = np.uint8(rgb * 255)
+                    frame[ind_2] = np.uint8(rgb * 255)
+
+            anim[i] = frame
+
+        img = Image.fromarray(anim)
+        img = img.filter(ImageFilter.GaussianBlur(1.5))
```
Kotochleb: I would also treat it as calss variable

Kotochleb (changes requested):
All divisions resulting in float values are divided by int values. Please update that

panther_lights/src/animation/battery_animation.py
KmakD: `img` variable is only used in this scope and later converted to an array and assigned to the `self._anim`. I don't see a point in making it a class variable

panther_lights/src/animation/battery_animation.py
```diff
@@ -0,0 +1,75 @@
...
+class BatteryAnimation(Animation):
+
+    ANIMATION_NAME = 'battery_animation'
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+        self._anim = np.zeros((self._anim_len, self.num_led))
+        self._h_min = 0
+        self._h_max = 120
```
Kotochleb:
```suggestion
        self._h_min = 0.0
        self._h_max = 120.0
```

panther_lights/src/animation/battery_animation.py
Kotochleb: I mean `1.5` from gaussian filter. Sorry for not being precise

KmakD: Ok, that make more sense

Kotochleb (approved):

PR #62, opened 2023-02-13
Handle battery status

KmakD:


panther_lights/src/scheduler_node.py
```diff
@@ -64,12 +65,15 @@ def __init__(self, name) -> None:
...
+        rospy.loginfo(f'[{rospy.get_name()}] Node started')
 
     def _scheduler_timer_cb(self, *args) -> None:
 
         # call animation service only when e_stop state changes
-        if self._led_e_stop_state != self._e_stop_state and not self._charger_connected:
+        if (
+            self._led_e_stop_state != self._e_stop_state
+            and not self._battery_status in self._charger_connected_states
+        ):
```
Kotochleb, line 76: Since e-stop is latched, message detecting state change I think can be handled easier.

Kotochleb (changes requested):

panther_lights/src/scheduler_node.py
KmakD: When this would be handled in topic callback, If calling service fails, then it won't retry. May cause displaying wrong animation.

Kotochleb (approved):

PR #73, opened 2023-03-07
Ros1 low pass gpio filter

Kotochleb:


panther_power_control/src/power_board_node.py
```diff
@@ -123,16 +127,16 @@ def __init__(self, name: str) -> None:
         #   Timers
         # -------------------------------
 
-        # 2 Hz publish non asynch pin state
-        self._charger_state_timer = rospy.Timer(rospy.Duration(0.5), self._publish_pin_state_cb)
+        # 5 Hz publish non asynch pin state
+        self._charger_state_timer = rospy.Timer(rospy.Duration(0.2), self._publish_pin_state_cb)
```
KmakD:
```suggestion
        self._pub_pin_state_timer = rospy.Timer(rospy.Duration(0.2), self._publish_pin_state_cb)
```
This is a bit misleading as timer does not apply to charger state only

KmakD (changes requested):

KmakD (approved):

PR #77, opened 2023-03-10
Can error check

pkowalsk1:


panther_driver/src/driver_node.py
```diff
@@ -407,15 +409,15 @@ def _publish_odom_cb(self) -> None:
     def _trigger_panther_estop(self) -> bool:
         try:
             response = self._estop_trigger()
-            rospy.logwarn(f'[{rospy.get_name()}] Trying to trigger Panther e-stop... Response: {response.success}')
+            rospy.logwarn_throttle(2.0, f'[{rospy.get_name()}] Trying to trigger Panther e-stop... Response: {response.success}')
 
             if not response.success:
-                return True
-
+                return False
```
Kotochleb: Eeeeeeem...
``` python
return response.success
```
Isn't it the same...?

panther_driver/src/driver_node.py
```diff
@@ -407,15 +409,15 @@ def _publish_odom_cb(self) -> None:
...
+        
         except rospy.ServiceException as e:
             rospy.logerr(f'[{rospy.get_name()}] Can\'t trigger Panther e-stop... \n{e}')
```
Kotochleb, line 416: Also this place does not return anything. If we catch exception being an error we return true

Kotochleb (changes requested):

Kotochleb (approved):

PR #100, opened 2023-03-27
BT manager node draft

KmakD:
README was not updated as it is very likely information written at this point would change before the final merge.


panther_manager_bt/package.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package format="3">
-  <name>panther_manager</name>
-  <version>1.0.0</version>
+  <name>panther_manager_bt</name>
```
Kotochleb: Can't we keep the name `panther_manager` and call the node `panther_manager_bt`?

panther/panther.repos
```diff
@@ -6,4 +6,8 @@ repositories:
   ros_components_description:
     type: git
     url: https://github.com/husarion/ros_components_description.git
-    version: ros1
\ No newline at end of file
+    version: ros1
+  BehaviorTree.CPP:
+    type: git
+    url: https://github.com/BehaviorTree/BehaviorTree.CPP.git
+    version: d7073b8
```
Kotochleb:
```suggestion
    version:  4.1.0 
```
Fixing to tag would be better I think

panther_manager_bt/src/manager_node.cpp
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  factory_.registerBehaviorTreeFromFile(xml_filename_);
+
+  setup_behavior_tree(nh, lights_tree_, lights_config_, "Lights");
+  setup_behavior_tree(nh, safety_tree_, safety_config_, "Safety");
+  setup_behavior_tree(nh, shutdown_tree_, shutdown_config_, "Shutdown");
+
+  e_stop_sub_ = nh->subscribe("hardware/e_stop", 10, &ManagerNode::e_stop_cb, this);
+  io_state_sub_ = nh->subscribe("hardware/io_state", 10, &ManagerNode::io_state_cb, this);
```
Kotochleb:
```suggestion
  e_stop_sub_ = nh->subscribe("hardware/e_stop", 2, &ManagerNode::e_stop_cb, this);
  io_state_sub_ = nh->subscribe("hardware/io_state", 2, &ManagerNode::io_state_cb, this);
```
Queues can be much smaller due to the nature of those topics

Kotochleb (changes requested):
Use shared pointer for the node handle

Kotochleb (approved):

PR #101, opened 2023-03-31
ROS 1 manager bt add basic plugins

KmakD:


panther_manager/config/manager_bt_config.yaml
```diff
@@ -0,0 +1,4 @@
+plugin_libs:
```
Kotochleb, line 1:
```suggestion
bt_plugin_libs:
```
Wouldn't it look better?

Kotochleb (approved):

PR #103, opened 2023-04-06
Bat-current-check

pkowalsk1:


Kotochleb (changes requested):
This way it is absolutely  wrong. This logic should be encapsulated in the battery node, not in the power control

PR #104, opened 2023-04-12
Ros1 manager bt create basic trees

KmakD:


panther_manager/README.md
```diff
@@ -4,9 +4,9 @@ A package containing nodes responsible for high-level control of Husarion Panthe
 
 ## ROS Nodes
 
-### manager_node.py
+### manager_bt_node.cpp
```
Kotochleb: I would name this after executable, not file. `.py` is an executable, but in this case I would leave it without cpp

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,128 @@
...
-        <CallTriggerService service_name="hardware/e_stop_trigger"
-                            timeout="100"
-                            _skipIf="e_stop_state"/>
-        <CallSetLedAnimationService id="0"
-                                    param=""
-                                    repeating="false"
-                                    service_name="lights/controller/set/animation"
-                                    timeout="100"
-                                    _skipIf="!e_stop_state"/>
+        <Sequence _skipIf="battery_status != 1 &amp;&amp; battery_status != 4">
```
Kotochleb: Battery statuses can be stored as constants in blackboard as "BATTERY_CHARGING", etc, with code similar to:
``` c++
std::map<std::string, std::any> cosntsnts_bb = {
  {"BATTERY_STATE_ANIM_ID", panther_msgs::LEDAnimation::BATTERY_STATE},
};
```

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,128 @@
...
+          <Script name="Charging"
+                  code="anim_id := 9; param := battery_percent_round"/>
```
Kotochleb: I would also create symbolic constants in blackboard for animations

panther_manager/include/panther_manager/manager_bt_node.hpp
```diff
@@ -9,30 +10,41 @@
...
+  ManagerNode(std::shared_ptr<ros::NodeHandle> nh, std::shared_ptr<ros::NodeHandle> pnh);
   ~ManagerNode() {}
 
 private:
+  float battery_percent_;
+  float update_charging_anim_step_;
+  float shutdown_timeout_;
   std::string node_name_;
-  std::string xml_filename_;
+  std::optional<unsigned> battery_status_;
```
Kotochleb, line 36:
```suggestion
  std::optional<std::uint8_t> battery_status_;
```
The message field is not purely unsigned

panther_manager/include/panther_manager/manager_bt_node.hpp
```diff
@@ -9,30 +10,41 @@
...
 
+  ros::Subscriber battery_sub_;
+  ros::Subscriber driver_state_sub_;
   ros::Subscriber e_stop_sub_;
   ros::Subscriber io_state_sub_;
+  ros::Subscriber system_status_sub_;
   ros::Timer lights_tree_timer_;
   ros::Timer safety_tree_timer_;
   std::shared_ptr<ros::NodeHandle> nh_;
+  std::shared_ptr<ros::NodeHandle> pnh_;
```
Kotochleb:
```suggestion
  std::shared_ptr<ros::NodeHandle> ph_;
```
We discussed this already

panther_manager/include/panther_manager/manager_bt_node.hpp
```diff
@@ -1,6 +1,7 @@
 #ifndef PANTHER_MANAGER_MANAGER_NODE_HPP_
 #define PANTHER_MANAGER_MANAGER_NODE_HPP_
 
+#include <any>
 #include <behaviortree_cpp/bt_factory.h>
```
Kotochleb, line 10: Those are ROS packages, not standard libraries

panther_manager/include/panther_manager/moving_average.hpp
```diff
@@ -0,0 +1,39 @@
...
+#define PANTHER_MANAGER_MOVING_AVERAGE_HPP_
+
+namespace panther_manager
+{
+
+template <typename T>
+class MovingAverage
+{
+public:
+  MovingAverage(int window_size = 5) : window_size_(window_size), sum_(T(0)) {}
```
Kotochleb:
```suggestion
  MovingAverage(const int window_size = 5) : window_size_(window_size), sum_(T(0)) {}
```

panther_manager/include/panther_manager/moving_average.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+  void add_value(T value)
```
Kotochleb:
```suggestion
  void add_value(const T value)
```

panther_manager/include/panther_manager/plugins/action/shutdown_hosts_from_file_node.hpp
```diff
@@ -0,0 +1,40 @@
+#ifndef PANTHER_MANAGER_SHUTDOWN_HOST_FROM_FILE_NODE_HPP_
+#define PANTHER_MANAGER_SHUTDOWN_HOST_FROM_FILE_NODE_HPP_
+
+#include <panther_manager/shutdown_host_node.hpp>
+#include <yaml-cpp/yaml.h>
```
Kotochleb:
```suggestion
#include <panther_manager/shutdown_host_node.hpp>

#include <yaml-cpp/yaml.h>
```

panther_manager/include/panther_manager/manager_bt_node.hpp
```diff
@@ -9,30 +10,41 @@
...
+#include <panther_manager/moving_average.hpp>
 
 namespace panther_manager
 {
 
 class ManagerNode
 {
 public:
-  ManagerNode(std::shared_ptr<ros::NodeHandle> nh);
+  ManagerNode(std::shared_ptr<ros::NodeHandle> nh, std::shared_ptr<ros::NodeHandle> pnh);
```
Kotochleb:
```suggestion
  ManagerNode(const std::shared_ptr<ros::NodeHandle> nh, const std::shared_ptr<ros::NodeHandle> ph);
```

panther_manager/include/panther_manager/plugins/action/shutdown_hosts_from_file_node.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+class ShutdownHostsFromFile : public ShutdownHost
+{
+public:
+  ShutdownHostsFromFile(const std::string & name, const BT::NodeConfig & conf)
+  : ShutdownHost(name, conf)
+  {
+  }
+
+  static BT::PortsList providedPorts()
```
Kotochleb, line 19:
```suggestion
  static BT::PortsList providedPorts() const
```

panther_manager/include/panther_manager/plugins/action/shutdown_single_host_node.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+class ShutdownSingleHost : public ShutdownHost
+{
+public:
+  ShutdownSingleHost(const std::string & name, const BT::NodeConfig & conf)
+  : ShutdownHost(name, conf)
+  {
+  }
+
+  static BT::PortsList providedPorts()
```
Kotochleb, line 17:
```suggestion
  static BT::PortsList providedPorts() const
```

panther_manager/include/panther_manager/plugins/decorator/tick_after_timeout_node.hpp
```diff
@@ -0,0 +1,29 @@
+#ifndef PANTHER_MANAGER_TICK_AFTER_TIMEOUT_NODE_HPP_
+#define PANTHER_MANAGER_TICK_AFTER_TIMEOUT_NODE_HPP_
+
+#include <behaviortree_cpp/bt_factory.h>
+#include <behaviortree_cpp/decorator_node.h>
```
Kotochleb: This should be after `<ros/ros.h>`

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
+#ifndef PANTHER_MANAGER_SHUTDOWN_HOST_NODE_HPP_
+#define PANTHER_MANAGER_SHUTDOWN_HOST_NODE_HPP_
+
+#include <behaviortree_cpp/action_node.h>
+#include <behaviortree_cpp/bt_factory.h>
```
Kotochleb: Should be included after <ros/ros.h>

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
...
+    session_ = ssh_new();
+    if (session_ == NULL) return -1;
+
+    ssh_options_set(session_, SSH_OPTIONS_HOST, host);
+    ssh_options_set(session_, SSH_OPTIONS_USER, user);
+    ssh_options_set(session_, SSH_OPTIONS_PORT, &port_);
+    ssh_options_set(session_, SSH_OPTIONS_LOG_VERBOSITY, &verbosity_);
+
+    int rc = ssh_connect(session_);
+    if (rc != SSH_OK) {
```
Kotochleb:
```suggestion
    if (ssh_connect(session_) != SSH_OK) {
```
Since `rc` is not used later this could be reduced. The same could be applied later on

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
...
+      ROS_ERROR("[%s] Error connecting to host: %s", node_name_.c_str(), ssh_get_error(session_));
+      ssh_free(session_);
+      return -1;
```
Kotochleb: In case of error, an exception should be raised instead of `return -1`. The same should be applied later on.

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
...
+    ssh_channel_close(channel_);
+    ssh_channel_free(channel_);
+    ssh_disconnect(session_);
+    ssh_free(session_);
+  }
+
+  std::string get_node_name() { return node_name_; }
+
+private:
+  int verbosity_ = SSH_LOG_NOLOG;
```
Kotochleb:
```suggestion
  const int verbosity_ = SSH_LOG_NOLOG;
```

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
...
+  int port_ = 22;
```
Kotochleb:
```suggestion
  int port_ = 22;
```
This could be a bt port with default set to 22

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
...
+  std::string node_name_;
```
Kotochleb: This could be const set in initialization list

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
+#ifndef PANTHER_MANAGER_SHUTDOWN_HOST_NODE_HPP_
+#define PANTHER_MANAGER_SHUTDOWN_HOST_NODE_HPP_
+
+#include <behaviortree_cpp/action_node.h>
```
Kotochleb: This whole file should be in `plugins` folder

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
...
+#include <behaviortree_cpp/bt_factory.h>
+#include <libssh/libssh.h>
+
+#include <ros/ros.h>
+
+namespace panther_manager
+{
```
Kotochleb: IMO there might be an additional namespace like `panther_ssh` or something. This can be discussed later on

panther_manager/package.xml
```diff
@@ -17,13 +17,17 @@
...
   <depend>panther_msgs</depend>
   <depend>roscpp</depend>
   <depend>roslib</depend>
+  <depend>sensor_msgs</depend>
   <depend>std_msgs</depend>
   <depend>std_srvs</depend>
+  <depend>yaml-cpp</depend>
 
   <!-- Python dependencies -->
+  <depend>iputils-ping</depend>
```
Kotochleb: This is not python dependency

panther_manager/include/panther_manager/ros_service_node.hpp
```diff
@@ -60,6 +50,16 @@ class RosServiceNode : public BT::SyncActionNode
...
   {
+    if (!getInput<std::string>("service_name", srv_name_) || srv_name_ == "") {
+      throw BT::RuntimeError("[", name(), "] Failed to get input [service_name]");
+    }
+
+    unsigned srv_timeout_ms;
+    if (!getInput<unsigned>("timeout", srv_timeout_ms)) {
+      throw BT::RuntimeError("[", name(), "] Failed to get input [timeout]");
+    }
+    srv_timeout_ = ros::Duration(static_cast<double>(srv_timeout_ms) * 1e-3);
```
Kotochleb, line 70: I am struggling with this one, because cast from ms to s should be done using chrono, but it will be absolutely pointless in this case...

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,31 +3,87 @@
...
-
-  for (const auto & p : plugin_libs_) {
+  auto xml_filename = pnh_->param<std::string>("xml_filename", default_xml);
+  auto plugin_libs = pnh_->param<std::vector<std::string>>("plugin_libs", default_plugin_libs);
+  auto battery_temp_window_len = pnh_->param<int>("battery_temp_window_len", 6);
+  auto cpu_temp_window_len = pnh_->param<int>("cpu_temp_window_len", 6);
+  auto driver_temp_window_len = pnh_->param<int>("driver_temp_window_len", 6);
+  auto shutdown_hosts_file = pnh_->param<std::string>("shutdown_hosts_file", "");
+  shutdown_timeout_ = pnh_->param<float>("shutdown_timeout", 15.0);
+  // lights tree params
```
Kotochleb, line 52:
```suggestion

  // lights tree params
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,31 +3,87 @@
...
+  auto critical_battery_anim_period =
+    pnh_->param<float>("lights/critical_battery_anim_period", 15.0);
+  auto critical_battery_threshold_percent =
+    pnh_->param<float>("lights/critical_battery_threshold_percent", 0.1);
+  auto battery_state_anim_period = pnh_->param<float>("lights/battery_state_anim_period", 120.0);
+  auto low_battery_anim_period = pnh_->param<float>("lights/low_battery_anim_period", 30.0);
+  auto low_battery_threshold_percent =
+    pnh_->param<float>("lights/low_battery_threshold_percent", 0.4);
+  update_charging_anim_step_ = pnh_->param<float>("lights/update_charging_anim_step", 0.1);
+  // safety tree params
```
Kotochleb, line 64:
```suggestion

  // safety tree params
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,31 +3,87 @@
...
+    {"driver_fan_on_temp", driver_fan_on_temp},
+    {"driver_fan_off_temp", driver_fan_off_temp},
+  };
+  std::map<std::string, std::any> shutdown_initial_bb = {
+    {"shutdown_hosts_file", shutdown_hosts_file.c_str()},
+  };
+
+  setup_behaviortree(lights_tree_, lights_config_, "Lights", lights_initial_bb);
+  setup_behaviortree(safety_tree_, safety_config_, "Safety", safety_initial_bb);
+  setup_behaviortree(shutdown_tree_, shutdown_config_, "Shutdown", shutdown_initial_bb);
```
Kotochleb: This should be separated into config setup and later invoking bt factory separately

panther_manager/src/manager_bt_node.cpp
```diff
@@ -41,6 +97,12 @@ ManagerNode::ManagerNode(std::shared_ptr<ros::NodeHandle> nh) : nh_(std::move(nh
     ros::spinOnce();
   }
 
+  while (ros::ok() && !battery_status_.has_value()) {
+    ROS_INFO("[%s] Waiting for battery message to arrive", node_name_.c_str());
+    ros::Duration(1.0).sleep();
+    ros::spinOnce();
+  }
```
Kotochleb: Spinning could be at higher frequency with throttled logs

panther_manager/src/manager_bt_node.cpp
```diff
@@ -86,19 +191,42 @@ void ManagerNode::lights_tree_timer_cb()
...
+
+  // tick shutdown tree
+  BT::StdCoutLogger logger_cout(shutdown_tree_);  // debuging
+  shutdown_tree_status_ = BT::NodeStatus::RUNNING;
+  auto start_time = ros::Time::now();
+  while (ros::ok() && shutdown_tree_status_ == BT::NodeStatus::RUNNING) {
+    auto shutdown_timeout = (ros::Time::now() - start_time) > ros::Duration(shutdown_timeout_);
+    shutdown_config_.blackboard->set<bool>("shutdown_timeout", shutdown_timeout);
+    shutdown_tree_status_ = shutdown_tree_.tickOnce();
+  }
```
Kotochleb, line 295: At the end there could be `ros::shutdown() invoked

panther_manager/src/manager_bt_node.cpp
```diff
@@ -49,6 +111,50 @@ ManagerNode::ManagerNode(std::shared_ptr<ros::NodeHandle> nh) : nh_(std::move(nh
...
+      config.blackboard->set<double>(item.first, std::any_cast<double>(item.second));
+    } else if (type == typeid(long)) {
+      config.blackboard->set<long>(item.first, std::any_cast<long>(item.second));
+    } else if (type == typeid(const char *)) {
+      config.blackboard->set<std::string>(item.first, std::any_cast<const char *>(item.second));
+    } else {
+      throw std::invalid_argument(
+        "Invalid type for blackboard entry. Valid types are: bool, int, float, double, long,"
+        " const char*");
+    }
```
Kotochleb: As you mentioned in conversation, this could be replaced with switch statement, or with templates...

panther_manager/src/manager_bt_node.cpp
```diff
@@ -49,6 +111,50 @@ ManagerNode::ManagerNode(std::shared_ptr<ros::NodeHandle> nh) : nh_(std::move(nh
...
+  }
+
+  tree = factory_.createTree(tree_name, config.blackboard);
+}
```
Kotochleb: As mentioned before, the whole thing should be separated into two functions

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,31 +3,87 @@
 namespace panther_manager
 {
 
-ManagerNode::ManagerNode(std::shared_ptr<ros::NodeHandle> nh) : nh_(std::move(nh))
+ManagerNode::ManagerNode(
+  std::shared_ptr<ros::NodeHandle> nh, std::shared_ptr<ros::NodeHandle> private_nh)
```
Kotochleb:
```suggestion
  const std::shared_ptr<ros::NodeHandle> nh, const std::shared_ptr<ros::NodeHandle> private_nh)
```

panther_manager/src/main.cpp
```diff
@@ -5,8 +5,9 @@ int main(int argc, char ** argv)
   ros::init(argc, argv, "manager_bt_node");
   ros::AsyncSpinner spinner(0);
   auto nh = std::make_shared<ros::NodeHandle>();
+  auto pnh = std::make_shared<ros::NodeHandle>("~");
```
Kotochleb:
```suggestion
  auto ph = std::make_shared<ros::NodeHandle>("~");
```

panther_manager/src/main.cpp
```diff
@@ -5,8 +5,9 @@ int main(int argc, char ** argv)
...
 
-  panther_manager::ManagerNode manager_bt_node(nh);
+  panther_manager::ManagerNode manager_bt_node(nh, pnh);
```
Kotochleb:
```suggestion
  panther_manager::ManagerNode manager_bt_node(nh, ph);
```

panther_manager/README.md
```diff
@@ -27,45 +27,63 @@ echo $USERNAME 'ALL=(ALL) NOPASSWD: /sbin/poweroff, /sbin/reboot, /sbin/shutdown
...
+
+### Nodes
+
+#### Actions
+
+- `CallSetBoolService` - allows calling standard **std_srvs/SetBool** ROS service.
+- `CallSetLedAnimationService` - allows calling custom type **panther_msgs/SetLEDAnimation** ROS service.
+- `CallTriggerService` - allows calling standard **std_srvs/Trigger** ROS service.
+- `ShutdownHostsFromFile` - allows to shutdown devices based on YAML file. Returns FAILURE only when YAML file is incorrect or a path to the file does not exists. If it fails to execute command in a remote host node will proceed tothe next device from the list.
+- `ShutdownSingleHost` - allows to shutdown single device. Will return SUCCESS only when device can be reached and the command was executed.
```
Kotochleb:
```suggestion
- `CallSetBoolService` - allows calling standard **std_srvs/SetBool** ROS service.
- `CallSetLedAnimationService` - allows calling custom type **panther_msgs/SetLEDAnimation** ROS service.
- `CallTriggerService` - allows calling standard **std_srvs/Trigger** ROS service.
- `ShutdownHostsFromFile` - allows to shutdown devices based on YAML file. Returns `FAILURE` only when YAML file is incorrect or a path to the file does not exists. If it fails to execute command in a remote host node will proceed tothe next device from the list.
- `ShutdownSingleHost` - allows to shutdown single device. Will return `SUCCESS` only when device can be reached and the command was executed.
```
Nodes also descriptions miss available ports with their types

panther_manager/README.md
```diff
@@ -27,45 +27,63 @@ echo $USERNAME 'ALL=(ALL) NOPASSWD: /sbin/poweroff, /sbin/reboot, /sbin/shutdown
...
+
+#### Decorators
+
+- `TickAfterTimeout` - will skip child before specified time has passed. Can be used to control frequency at witch given behavior in the tree is triggered.
```
Kotochleb:
```suggestion
- `TickAfterTimeout` - will skip child before specified time has passed. Can be used to specify frequency at witch node or subrtee is triggered.
```

panther_manager/README.md
```diff
@@ -27,45 +27,63 @@ echo $USERNAME 'ALL=(ALL) NOPASSWD: /sbin/poweroff, /sbin/reboot, /sbin/shutdown
...
+
+### Trees
+
+- `Lights` - this tree is responsible for scheduling animations displayed on LED panels based on the Husarion Panther robot's system state.
+- `Safety` - this tree is responsible for monitoring the Panther robot's state and handling safety measures like cooling the robot in case of high CPU or battery temperature.
+- `Shutdown` - this tree is responsible for gracefull shutdown of robot components and the built-in computer.
```
Kotochleb:
```suggestion
- `Lights` - tree responsible for scheduling animations displayed on LED panels, based on the Husarion Panther robot's system state.
- `Safety` - tree responsible for monitoring the Panther robot's state and handling safety measures, such as  cooling the robot in case of high CPU or battery temperature.
- `Shutdown` - tree responsible for gracefull shutdown of robot components and the built-in computers.
```

panther_manager/config/PantherBT.btproj
```diff
@@ -19,5 +19,16 @@
             <input_port name="service_name">ROS service name</input_port>
             <input_port name="timeout" default="100">timeout in ms to wait for service to be active</input_port>
         </Action>
+        <Action ID="ShutdownHostsFromFile" editable="true">
+            <input_port name="shutdown_hosts_file"/>
```
Kotochleb:
```suggestion
            <input_port name="shutdown_hosts_file">global path to YAML file with hosts to hsutdown</input_port>
```

panther_manager/config/PantherBT.btproj
```diff
@@ -19,5 +19,16 @@
...
+        </Action>
+        <Action ID="ShutdownSingleHost" editable="true">
+            <input_port name="command">(optional)</input_port>
+            <input_port name="ip"/>
+            <input_port name="user"/>
```
Kotochleb:
```suggestion
            <input_port name="command">command to execute on shutdown (optional)</input_port>
            <input_port name="ip">ip of the host to shudwon</input_port>
            <input_port name="user">user to log into while executing shutdown command</input_port>
```

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,128 @@
...
+                            _skipIf="!e_stop_state">
+            <CallSetLedAnimationService name="SetEStopAnimation"
+                                        id="0"
+                                        param=""
+                                        repeating="false"
+                                        service_name="lights/controller/set/animation"
+                                        timeout="100"/>
+          </TickAfterTimeout>
+        </Sequence>
+        <Sequence name="Discharging"
```
Kotochleb:
```suggestion
        <Sequence name="DischargingSequence"
```

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,128 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <root BTCPP_format="4">
   <BehaviorTree ID="Lights">
-    <Delay delay_msec="5000">
+    <Sequence>
```
Kotochleb, line 4:
```suggestion
    <Sequence name="ChargingSequence"
```

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,128 @@
...
+                              timeout="100"
+                              _skipIf="!aux_state || bat_temp &lt;= critical_bat_temp"/>
+          <CallSetBoolService name="EnableFan"
+                              data="true"
+                              service_name="hardware/fan_enable"
+                              timeout="100"
+                              _skipIf="fan_state"/>
+        </Sequence>
+      </Sequence>
+      <Sequence _skipIf="bat_temp &gt; high_bat_temp">
```
Kotochleb:
```suggestion
      <Sequence name="ManagePowerSequence" _skipIf="bat_temp &gt; high_bat_temp">
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,31 +3,87 @@
...
+  auto driver_fan_off_temp = pnh_->param<float>("safety/driver_fan_off_temp", 35.0);
+
+  battery_temp_ma_ = MovingAverage<double>(battery_temp_window_len);
+  cpu_temp_ma_ = MovingAverage<double>(cpu_temp_window_len);
+  front_driver_temp_ma_ = MovingAverage<double>(driver_temp_window_len);
+  rear_driver_temp_ma_ = MovingAverage<double>(driver_temp_window_len);
+
+  ROS_INFO("[%s] Register BehaviorTree from: %s", node_name_.c_str(), xml_filename.c_str());
+
+  for (const auto & p : plugin_libs) {
```
Kotochleb, line 82:
```suggestion
  // export plugins for a behaviour tree
  for (const auto & p : plugin_libs) {
```

panther_manager/include/panther_manager/moving_average.hpp
Kotochleb:
```suggestion
  void roll(T value)
```
add value implies extending the underlying array. Roll is similar to how they named it in [pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rolling.html)

panther_manager/include/panther_manager/ros_service_node.hpp
```diff
@@ -18,16 +18,6 @@ class RosServiceNode : public BT::SyncActionNode
     std::shared_ptr<ros::NodeHandle> & nh, const std::string & name, const BT::NodeConfig & conf)
   : BT::SyncActionNode(name, conf), nh_(nh)
```
Kotochleb: The Constructor should be public,
there should be protected or public `providedBasicPorts` doing the same thing as `providedPorts` right now.

``` c++
nh_ = config().blackboard->get<std::shared_ptr<ros::NodeHandle>>("nh");
getInput("service_name", srv_name_);
```
This should be moved to the constructor of the `RosServiceNode` class

panther_manager/include/panther_manager/plugins/action/call_set_bool_service_node.hpp
```diff
@@ -8,7 +8,7 @@
 namespace panther_manager
 {
 
-class CallSetBoolService : public panther_manager::RosServiceNode<std_srvs::SetBool>
+class CallSetBoolService : public RosServiceNode<std_srvs::SetBool>
 {
 public:
   CallSetBoolService(const std::string & name, const BT::NodeConfig & conf);
```
Kotochleb: Everything is public. As well as, `nh_` is being redefined. How the hell this compiled without warning?

panther_manager/include/panther_manager/ros_service_node.hpp
```diff
@@ -18,16 +18,6 @@ class RosServiceNode : public BT::SyncActionNode
...
   {
-    if (!getInput<std::string>("service_name", srv_name_) || srv_name_ == "") {
-      throw BT::RuntimeError("[", name, "] Failed to get input [service_name]");
-    }
```
Kotochleb, line 23: This actually should be reverted

panther_manager/include/panther_manager/plugins/action/call_set_bool_service_node.hpp
```diff
@@ -8,7 +8,7 @@
 namespace panther_manager
 {
 
-class CallSetBoolService : public panther_manager::RosServiceNode<std_srvs::SetBool>
+class CallSetBoolService : public RosServiceNode<std_srvs::SetBool>
 {
 public:
```
Kotochleb, line 18: If the parent class will be updated with the ports this node can be reduced to header only

panther_manager/include/panther_manager/plugins/action/call_set_led_animation_service_node.hpp
```diff
@@ -8,7 +8,7 @@
 namespace panther_manager
 {
 
-class CallSetLedAnimationService : public panther_manager::RosServiceNode<panther_msgs::SetLEDAnimation>
+class CallSetLedAnimationService : public RosServiceNode<panther_msgs::SetLEDAnimation>
 {
 public:
   CallSetLedAnimationService(const std::string & name, const BT::NodeConfig & conf);
```
Kotochleb: As in `set_bool` everything is public, `nh_` is redefined, `srv_name_` as well. This time, provided ports is overwriting one from `RosServiceNode`, so `timeout` cannot be set. We will have to see if this class can be header only

panther_manager/include/panther_manager/plugins/action/call_trigger_service_node.hpp
```diff
@@ -8,7 +8,7 @@
 namespace panther_manager
 {
 
-class CallTriggerService : public panther_manager::RosServiceNode<std_srvs::Trigger>
+class CallTriggerService : public RosServiceNode<std_srvs::Trigger>
 {
 public:
   CallTriggerService(const std::string & name, const BT::NodeConfig & conf);
```
Kotochleb: If applied same changes as in `set_bol_service` this will be fine. Header only class applies here

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
...
+public:
+  ShutdownHost(const std::string & name, const BT::NodeConfig & conf)
+  : BT::StatefulActionNode(name, conf)
+  {
+    node_name_ = ros::this_node::getName();
+  }
+
+  ssh_session session_;
+  ssh_channel channel_;
```
Kotochleb: This should be private with `channel_` in getter or wrappers to `ssh_channel_is_open` and `ssh_channel_read_nonblocking` within `ShutdownHost` class

panther_manager/plugins/action/shutdown_single_host_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+  if ((nbytes_ = ssh_channel_read_nonblocking(channel_, buffer_, sizeof(buffer_), 0)) >= 0) {
+    output_.append(buffer_, nbytes_);
+    return BT::NodeStatus::RUNNING;
+  }
+
+  ROS_INFO("[%s] Device response:\n%s", get_node_name().c_str(), output_.c_str());
+  close_connection();
+  return BT::NodeStatus::SUCCESS;
+}
```
Kotochleb, line 43: There could be a port `wait_for_shutdown`, if set to true, the node will continue returning `BT::NodeStatus::RUNNING` as long as the device can be pinged

panther_manager/include/panther_manager/shutdown_host_node.hpp
```diff
@@ -0,0 +1,109 @@
+#ifndef PANTHER_MANAGER_SHUTDOWN_HOST_NODE_HPP_
+#define PANTHER_MANAGER_SHUTDOWN_HOST_NODE_HPP_
```
Kotochleb: Logic of pinging the device, logging into ssh and shutting it down could be implemented in this class. This could be implemented as map with IP string keys. In case of single host this would only have single entry in the map

Kotochleb (changes requested):

panther_manager/include/panther_manager/shutdown_host_node.hpp
KmakD: As discussed, shutdown node will change so for now this changes won't be applied

panther_manager/src/manager_bt_node.cpp
KmakD: Unfortunately it is not possible to replace it with switch, because `typeid()` function is not a constant expression. Also, I played with templates for a bit and haven't found any workaround. As bad as this "ifish" code looks a can't see a better solution.

panther_manager/include/panther_manager/ros_service_node.hpp
KmakD: this was moved to `tick()` method

KmakD:
As discussed, changes and detailed review of `shutdown_host_node` will be a part of another task and PR

panther_manager/README.md
Kotochleb: There is still description of ports of nodes missing

panther_manager/README.md
```diff
@@ -27,45 +27,63 @@ echo $USERNAME 'ALL=(ALL) NOPASSWD: /sbin/poweroff, /sbin/reboot, /sbin/shutdown
...
+
+#### Decorators
+
+- `TickAfterTimeout` - will skip child until the specified time has passed. It can be used to specify the frequency at which a node or subtree is triggered.
+
+### Trees
+
+- `Lights` - tree responsible for scheduling animations displayed on LED panels, based on the Husarion Panther robot's system state.
+- `Safety` - tree responsible for monitoring the Panther robot's state and handling safety measures, such as  cooling the robot in case of high CPU or battery temperature.
+- `Shutdown` - tree responsible for gracefull shutdown of robot components and the built-in computer.
```
Kotochleb:
```suggestion
- `Shutdown` - tree responsible for gracefull shutdown of robot components, user computers and the and the built-in computer.
```

panther_manager/include/panther_manager/manager_bt_node.hpp
Kotochleb: After discussion I am dropping this idea

panther_manager/include/panther_manager/manager_bt_node.hpp
Kotochleb: Changed everywhere. Resolving this

panther_manager/include/panther_manager/manager_bt_node.hpp
Kotochleb: After discussion I am accepting it

panther_manager/include/panther_manager/plugins/decorator/tick_after_timeout_node.hpp
Kotochleb: Since we decided to treat `BehaviourTree.CPP` as an external library, this is ok

panther_manager/include/panther_manager/shutdown_host_node.hpp
Kotochleb: Since we decided to treat `BehaviourTree.CPP` as an external library, this is ok

panther_manager/include/panther_manager/shutdown_host_node.hpp
Kotochleb: Shutdowon hosts will be separate PR. Can be skipped by now

panther_manager/include/panther_manager/shutdown_host_node.hpp
Kotochleb: SSH will be moved to the different PR. I will accept leaving it as it is

panther_manager/include/panther_manager/ros_service_node.hpp
Kotochleb: There is no entirely clean solution, so I will leve it as it is

panther_manager/src/manager_bt_node.cpp
Kotochleb: Kinda big blob of code, but is acceptable. Can stay as it is

panther_manager/include/panther_manager/shutdown_host_node.hpp
Kotochleb: Since shutdown hosts will be separate PR I will leave it this way

panther_manager/plugins/action/shutdown_single_host_node.cpp
Kotochleb: This feature will be moved to separate PR

panther_manager/include/panther_manager/shutdown_host_node.hpp
Kotochleb: This will be moved to separate PR

panther_manager/src/manager_bt_node.cpp
Kotochleb: I just realized this will be never called since we firs call `sudo shutdown` on host device... Using it as a separate tree leaf doesn't make sense either...

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,143 @@
...
   <BehaviorTree ID="Lights">
-    <Delay delay_msec="5000">
-      <Sequence>
-        <CallTriggerService service_name="hardware/e_stop_trigger"
-                            timeout="100"
-                            _skipIf="e_stop_state"/>
-        <CallSetLedAnimationService id="0"
+    <Sequence>
+      <Sequence _skipIf="battery_status != POWER_SUPPLY_STATUS_CHARGING \
+&amp;&amp; battery_status != POWER_SUPPLY_STATUS_FULL">
```
Kotochleb:
```suggestion
      <Sequence name="ChargingSequence"
                _skipIf="battery_status != POWER_SUPPLY_STATUS_CHARGING \
```

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,143 @@
...
+                          _skipIf="!e_stop_state">
+          <CallSetLedAnimationService name="SetEStopAnimation"
+                                      id="{E_STOP_ANIM_ID}"
+                                      param=""
+                                      repeating="false"
+                                      service_name="lights/controller/set/animation"
+                                      timeout="100"/>
+        </TickAfterTimeout>
+      </Sequence>
+      <Sequence name="Discharging"
```
Kotochleb:
```suggestion
      <Sequence name="DischargingSequence"
```

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,143 @@
...
+                                    id="{E_STOP_ANIM_ID}"
                                     param=""
-                                    repeating="false"
+                                    repeating="true"
                                     service_name="lights/controller/set/animation"
                                     timeout="100"
-                                    _skipIf="!e_stop_state"/>
+                                    _skipIf="!e_stop_state || current_anim_id == E_STOP_ANIM_ID"
+                                    _onSuccess="current_anim_id = E_STOP_ANIM_ID"/>
+        <Sequence>
```
Kotochleb:
```suggestion
        <Sequence name="DisplayAnimationStatusSequence">
```

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,143 @@
...
-      <Sequence>
-        <CallSetBoolService data="true"
-                            service_name="hardware/enable_fan"
+    <Sequence>
+      <Sequence name="ManagePowerSequence"
+                _skipIf="bat_temp &lt;= high_bat_temp">
+        <Script name="SignalShutdown"
+                code="signal_shutdown := &quot;Battery reached fatal temperature&quot;"
+                _skipIf="bat_temp &lt;= fatal_bat_temp"/>
+        <Sequence _skipIf="bat_temp &gt; fatal_bat_temp">
```
Kotochleb:
```suggestion
        <Sequence name="FatalBatterySequence"
                  _skipIf="bat_temp &gt; fatal_bat_temp">
```

panther_manager/config/PantherManagerBT.xml
```diff
@@ -1,34 +1,143 @@
...
+                              timeout="100"
+                              _skipIf="!aux_state || bat_temp &lt;= critical_bat_temp"/>
+          <CallSetBoolService name="EnableFan"
+                              data="true"
+                              service_name="hardware/fan_enable"
+                              timeout="100"
+                              _skipIf="fan_state"/>
+        </Sequence>
+      </Sequence>
+      <Sequence _skipIf="bat_temp &gt; high_bat_temp">
```
Kotochleb:
```suggestion
      <Sequence name="TemperatureControlSequence"
                _skipIf="bat_temp &gt; high_bat_temp">
```

panther_manager/include/panther_manager/manager_bt_node.hpp
```diff
@@ -45,11 +60,20 @@ class ManagerNode
...
+
+  void battery_cb(const sensor_msgs::BatteryState::ConstPtr & battery);
+  void driver_state_cb(const panther_msgs::DriverState::ConstPtr & driver_state);
   void e_stop_cb(const std_msgs::Bool::ConstPtr & e_stop);
   void io_state_cb(const panther_msgs::IOState::ConstPtr & io_state);
+  void system_status_cb(const panther_msgs::SystemStatus::ConstPtr & system_status);
   void safety_tree_timer_cb();
   void lights_tree_timer_cb();
+  void shutdown_robot(const std::string & message);
+  BT::NodeConfig create_bt_config(const std::map<std::string, std::any> bb_values = {});
```
Kotochleb:
```suggestion
  BT::NodeConfig create_bt_config(const std::map<std::string, std::any> bb_values = {}) const;
```

This function is not modifying anything within class, hence can be `const`

panther_manager/src/manager_bt_node.cpp
```diff
@@ -49,6 +140,53 @@ ManagerNode::ManagerNode(std::shared_ptr<ros::NodeHandle> nh) : nh_(std::move(nh
   ROS_INFO("[%s] Node started", node_name_.c_str());
 }
 
+BT::NodeConfig ManagerNode::create_bt_config(const std::map<std::string, std::any> bb_values)
```
Kotochleb:
```suggestion
BT::NodeConfig ManagerNode::create_bt_config(const std::map<std::string, std::any> bb_values) const
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,41 +3,132 @@
...
+
+  // export plugins for a behaviour tree
+  for (const auto & p : plugin_libs) {
     factory_.registerFromPlugin(BT::SharedLibrary::getOSName(p));
   }
 
-  factory_.registerBehaviorTreeFromFile(xml_filename_);
+  factory_.registerBehaviorTreeFromFile(xml_filename);
+
+  std::map<std::string, std::any> lights_initial_bb = {
```
Kotochleb:
```suggestion
  const std::map<std::string, std::any> lights_initial_bb = {
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,41 +3,132 @@
...
+     unsigned(sensor_msgs::BatteryState::POWER_SUPPLY_STATUS_UNKNOWN)},
+    {"POWER_SUPPLY_STATUS_CHARGING",
+     unsigned(sensor_msgs::BatteryState::POWER_SUPPLY_STATUS_CHARGING)},
+    {"POWER_SUPPLY_STATUS_DISCHARGING",
+     unsigned(sensor_msgs::BatteryState::POWER_SUPPLY_STATUS_DISCHARGING)},
+    {"POWER_SUPPLY_STATUS_NOT_CHARGING",
+     unsigned(sensor_msgs::BatteryState::POWER_SUPPLY_STATUS_NOT_CHARGING)},
+    {"POWER_SUPPLY_STATUS_FULL", unsigned(sensor_msgs::BatteryState::POWER_SUPPLY_STATUS_FULL)},
+  };
+  std::map<std::string, std::any> safety_initial_bb = {
```
Kotochleb:
```suggestion
  const std::map<std::string, std::any> safety_initial_bb = {
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,41 +3,132 @@
...
+    {"high_bat_temp", high_bat_temp},
+    {"critical_bat_temp", critical_bat_temp},
+    {"fatal_bat_temp", fatal_bat_temp},
+    {"cpu_fan_on_temp", cpu_fan_on_temp},
+    {"cpu_fan_off_temp", cpu_fan_off_temp},
+    {"driver_fan_on_temp", driver_fan_on_temp},
+    {"driver_fan_off_temp", driver_fan_off_temp},
+  };
+  std::map<std::string, std::any> shutdown_initial_bb = {
```
Kotochleb:
```suggestion
  const std::map<std::string, std::any> shutdown_initial_bb = {
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,41 +3,132 @@
...
 
-  factory_.registerBehaviorTreeFromFile(xml_filename_);
+  factory_.registerBehaviorTreeFromFile(xml_filename);
+
+  std::map<std::string, std::any> lights_initial_bb = {
+    {"critical_battery_anim_period", critical_battery_anim_period},
+    {"critical_battery_threshold_percent", critical_battery_threshold_percent},
+    {"battery_state_anim_period", battery_state_anim_period},
+    {"low_battery_anim_period", low_battery_anim_period},
+    {"low_battery_threshold_percent", low_battery_threshold_percent},
```
Kotochleb: Those could be written in capital letters to keep them as blackboard constants, similarly to animations and battery constants. This way reading behavior tree it will be more visible what was meant to be constant. @KmakD thoughts?

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,41 +3,132 @@
...
+    {"POWER_SUPPLY_STATUS_FULL", unsigned(sensor_msgs::BatteryState::POWER_SUPPLY_STATUS_FULL)},
+  };
+  std::map<std::string, std::any> safety_initial_bb = {
+    {"high_bat_temp", high_bat_temp},
+    {"critical_bat_temp", critical_bat_temp},
+    {"fatal_bat_temp", fatal_bat_temp},
+    {"cpu_fan_on_temp", cpu_fan_on_temp},
+    {"cpu_fan_off_temp", cpu_fan_off_temp},
+    {"driver_fan_on_temp", driver_fan_on_temp},
+    {"driver_fan_off_temp", driver_fan_off_temp},
```
Kotochleb: Those could be constants as in `lights_initial_bb`

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,41 +3,132 @@
...
+  };
+  std::map<std::string, std::any> shutdown_initial_bb = {
+    {"shutdown_hosts_file", shutdown_hosts_file.c_str()},
```
Kotochleb: This is also kinda constant

panther_manager/src/manager_bt_node.cpp
```diff
@@ -57,48 +195,71 @@ void ManagerNode::e_stop_cb(const std_msgs::Bool::ConstPtr & e_stop)
...
+
+  // tick shutdown tree
+  // BT::StdCoutLogger logger_cout(shutdown_tree_);  // debugging
+  shutdown_tree_status_ = BT::NodeStatus::RUNNING;
+  auto start_time = ros::Time::now();
+  while (ros::ok() && shutdown_tree_status_ == BT::NodeStatus::RUNNING) {
+    auto shutdown_timeout = (ros::Time::now() - start_time) > ros::Duration(shutdown_timeout_);
+    shutdown_config_.blackboard->set<bool>("shutdown_timeout", shutdown_timeout);
+    shutdown_tree_status_ = shutdown_tree_.tickOnce();
+  }
```
Kotochleb: There should be some sleep in the loop. Otherwise, we will end up with 100% CPU usage and a few hundred hertz of tree ticking frequency

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,41 +3,132 @@
...
   const std::vector<std::string> default_plugin_libs = {};
 
-  xml_filename_ = nh_->param<std::string>("xml_filename", default_xml);
-  plugin_libs_ = nh_->param<std::vector<std::string>>("plugin_libs", default_plugin_libs);
+  auto xml_filename = ph_->param<std::string>("xml_filename", default_xml);
+  auto plugin_libs = ph_->param<std::vector<std::string>>("plugin_libs", default_plugin_libs);
+  auto battery_temp_window_len = ph_->param<int>("battery_temp_window_len", 6);
+  auto cpu_temp_window_len = ph_->param<int>("cpu_temp_window_len", 6);
+  auto driver_temp_window_len = ph_->param<int>("driver_temp_window_len", 6);
+  auto shutdown_hosts_file = ph_->param<std::string>("shutdown_hosts_file", "");
```
Kotochleb:
```suggestion
  const auto xml_filename = ph_->param<std::string>("xml_filename", default_xml);
  const auto plugin_libs = ph_->param<std::vector<std::string>>("plugin_libs", default_plugin_libs);
  const auto battery_temp_window_len = ph_->param<int>("battery_temp_window_len", 6);
  const auto cpu_temp_window_len = ph_->param<int>("cpu_temp_window_len", 6);
  const auto driver_temp_window_len = ph_->param<int>("driver_temp_window_len", 6);
  const auto shutdown_hosts_file = ph_->param<std::string>("shutdown_hosts_file", "");
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,41 +3,132 @@
...
 
-  ROS_INFO("[%s] Register BehaviorTree from: %s", node_name_.c_str(), xml_filename_.c_str());
+  // lights tree params
+  auto critical_battery_anim_period =
+    ph_->param<float>("lights/critical_battery_anim_period", 15.0);
+  auto critical_battery_threshold_percent =
+    ph_->param<float>("lights/critical_battery_threshold_percent", 0.1);
+  auto battery_state_anim_period = ph_->param<float>("lights/battery_state_anim_period", 120.0);
+  auto low_battery_anim_period = ph_->param<float>("lights/low_battery_anim_period", 30.0);
+  auto low_battery_threshold_percent =
```
Kotochleb:
```suggestion
  const auto critical_battery_anim_period =
    ph_->param<float>("lights/critical_battery_anim_period", 15.0);
  const auto critical_battery_threshold_percent =
    ph_->param<float>("lights/critical_battery_threshold_percent", 0.1);
  const auto battery_state_anim_period = ph_->param<float>("lights/battery_state_anim_period", 120.0);
  const auto low_battery_anim_period = ph_->param<float>("lights/low_battery_anim_period", 30.0);
  const auto low_battery_threshold_percent =
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -3,41 +3,132 @@
...
 
-  for (const auto & p : plugin_libs_) {
+  // safety tree params
+  auto high_bat_temp = ph_->param<float>("safety/high_bat_temp", 55.0);
+  auto critical_bat_temp = ph_->param<float>("safety/critical_bat_temp", 59.0);
+  auto fatal_bat_temp = ph_->param<float>("safety/fatal_bat_temp", 62.0);
+  auto cpu_fan_on_temp = ph_->param<float>("safety/cpu_fan_on_temp", 70.0);
+  auto cpu_fan_off_temp = ph_->param<float>("safety/cpu_fan_off_temp", 60.0);
+  auto driver_fan_on_temp = ph_->param<float>("safety/driver_fan_on_temp", 45.0);
+  auto driver_fan_off_temp = ph_->param<float>("safety/driver_fan_off_temp", 35.0);
```
Kotochleb:
```suggestion
  const auto high_bat_temp = ph_->param<float>("safety/high_bat_temp", 55.0);
  const auto critical_bat_temp = ph_->param<float>("safety/critical_bat_temp", 59.0);
  const auto fatal_bat_temp = ph_->param<float>("safety/fatal_bat_temp", 62.0);
  const auto cpu_fan_on_temp = ph_->param<float>("safety/cpu_fan_on_temp", 70.0);
  const auto cpu_fan_off_temp = ph_->param<float>("safety/cpu_fan_off_temp", 60.0);
  const auto driver_fan_on_temp = ph_->param<float>("safety/driver_fan_on_temp", 45.0);
  const auto driver_fan_off_temp = ph_->param<float>("safety/driver_fan_off_temp", 35.0);
```

panther_manager/include/panther_manager/moving_average.hpp
```diff
@@ -0,0 +1,39 @@
...
+  T get_average() const
+  {
+    if (values.size() == 0) {
+      return T(0);
+    }
+    return sum_ / static_cast<T>(values.size());
+  }
+
+private:
+  int window_size_;
```
Kotochleb:
```suggestion
  const unsigned window_size_;
```

panther_manager/include/panther_manager/moving_average.hpp
```diff
@@ -0,0 +1,39 @@
...
+#define PANTHER_MANAGER_MOVING_AVERAGE_HPP_
+
+namespace panther_manager
+{
+
+template <typename T>
+class MovingAverage
+{
+public:
+  MovingAverage(const int window_size = 5) : window_size_(window_size), sum_(T(0)) {}
```
Kotochleb:
```suggestion
  MovingAverage(const unsigned window_size = 5) : window_size_(window_size) {}
```

panther_manager/include/panther_manager/moving_average.hpp
```diff
@@ -0,0 +1,39 @@
...
+    if (values.size() == 0) {
+      return T(0);
+    }
+    return sum_ / static_cast<T>(values.size());
+  }
+
+private:
+  int window_size_;
+  std::deque<T> values;
+  T sum_;
```
Kotochleb, line 33:
```suggestion
  T sum_ = 0;
```

panther_manager/include/panther_manager/manager_bt_node.hpp
```diff
@@ -1,38 +1,53 @@
 #ifndef PANTHER_MANAGER_MANAGER_NODE_HPP_
 #define PANTHER_MANAGER_MANAGER_NODE_HPP_
```
Kotochleb:
```suggestion
#ifndef PANTHER_MANAGER_MANAGER_BT_NODE_HPP_
#define PANTHER_MANAGER_MANAGER_BT_NODE_HPP_
```

Kotochleb: The file should be either `manager_node.hpp`, or class should be `ManagerBTNode`.

panther_manager/include/panther_manager/plugins/ros_service_node.hpp
```diff
@@ -13,53 +13,57 @@ namespace panther_manager
...
 
-public:
+  virtual ~RosServiceNode() = default;
+
   using ServiceType = ServiceT;
   using RequestType = typename ServiceT::Request;
   using ResponseType = typename ServiceT::Response;
 
-  static BT::PortsList providedPorts()
+  static BT::PortsList providedBasicPorts(BT::PortsList addition)
```
Kotochleb:
```suggestion
  static BT::PortsList providedBasicPorts(const BT::PortsList addition)
```

panther_manager/include/panther_manager/plugins/ros_service_node.hpp
```diff
@@ -13,53 +13,57 @@ namespace panther_manager
...
+    ports.insert(addition.begin(), addition.end());
+    return ports;
   }
 
   // methods to be implemented by user
   virtual void update_request(RequestType & request) = 0;
   virtual BT::NodeStatus on_response(const ResponseType & response) = 0;
 
   std::string get_node_name() { return node_name_; }
+  std::string get_srv_name() { return srv_name_; }
```
Kotochleb:
```suggestion
  std::string get_node_name() const { return node_name_; }
  std::string get_srv_name() const { return srv_name_; }
```

panther_manager/include/panther_manager/plugins/action/call_set_led_animation_service_node.hpp
```diff
@@ -3,28 +3,29 @@
...
-    return {
+    return providedBasicPorts({
       BT::InputPort<unsigned>("id", "animation ID"),
       BT::InputPort<std::string>("param", "optional parameter"),
       BT::InputPort<bool>("repeating", "indicates if animation should repeat"),
-    };
+    });
   }
 
   unsigned aniamtion_id_;
```
Kotochleb:
```suggestion
```
Unused variable

panther_manager/plugins/action/call_set_led_animation_service_node.cpp
```diff
@@ -3,14 +3,6 @@
 namespace panther_manager
```
Kotochleb, line 10: I can not refer to this line, but can't we assume that by default animation is not repeating?

panther_manager/include/panther_manager/plugins/action/call_trigger_service_node.hpp
```diff
@@ -3,18 +3,20 @@
...
+  CallTriggerService(const std::string & name, const BT::NodeConfig & conf)
+  : RosServiceNode(name, conf)
+  {
+  }
 
-  std::string srv_name_;
-  std::shared_ptr<ros::NodeHandle> nh_;
+  static BT::PortsList providedPorts() { return providedBasicPorts({}); }
 
   void update_request(RequestType & request) override;
```
Kotochleb:
```suggestion
  void update_request(RequestType & request) override {};
```

panther_manager/plugins/action/call_trigger_service_node.cpp
```diff
@@ -3,26 +3,19 @@
...
 {
 
-CallTriggerService::CallTriggerService(const std::string & name, const BT::NodeConfig & conf)
-: RosServiceNode(nh_, name, conf)
-{
-  nh_ = config().blackboard->get<std::shared_ptr<ros::NodeHandle>>("nh");
-  getInput("service_name", srv_name_);
-}
-
 void CallTriggerService::update_request(RequestType & request) {}
```
Kotochleb:
```suggestion
```
I don't think there is a point of defining it in the cpp file, since this function does exactly nothing

panther_manager/include/panther_manager/plugins/action/shutdown_hosts_from_file_node.hpp
Kotochleb: Came to realization static function can not be const

Kotochleb (changes requested):
Most of the changes are just a matter of const correctness so we are on a way to close this one soon

Kotochleb (changes requested):
Most of the changes are just a matter of const correctness so we are on a way to close this one soon

panther_manager/src/manager_bt_node.cpp
KmakD: Good idea

panther_manager/include/panther_manager/moving_average.hpp
KmakD: const won't work in here. If class instance is defined using default constructor (eg. in header file) it will be impossible to overwrite window_size later in cpp file.

panther_manager/include/panther_manager/shutdown_host_node.hpp
Kotochleb: We will work on details later in SSH closing task

panther_manager/src/manager_bt_node.cpp
Kotochleb: You used in this case is `ros::Rate rate(10.0)`, while in constructor you use `ros::Duration(0.1).sleep();` I think it would be better if in both cases the same approach was used

panther_manager/include/panther_manager/moving_average.hpp
Kotochleb: Using `std::unique_ptr` and dynamically creating the MovingAverage object solves the problem. IMO this is a more acceptable way than using `window_size_` which should be constant as mutable

panther_manager/README.md
```diff
@@ -27,45 +27,63 @@ echo $USERNAME 'ALL=(ALL) NOPASSWD: /sbin/poweroff, /sbin/reboot, /sbin/shutdown
...
+
+#### Decorators
+
+- `TickAfterTimeout` - will skip child until the specified time has passed. It can be used to specify the frequency at which a node or subtree is triggered.
+
+### Trees
+
+- `Lights` - tree responsible for scheduling animations displayed on LED panels, based on the Husarion Panther robot's system state.
+- `Safety` - tree responsible for monitoring the Panther robot's state and handling safety measures, such as  cooling the robot in case of high CPU or battery temperature.
+- `Shutdown` - tree responsible for gracefull shutdown of robot components, user computers and the built-in computer.
```
Kotochleb: There is still no information about black board input and output ports and no information about constants like animation IDs and so on. Can you at least make draft to be finished in a separate task?

Kotochleb (changes requested):
Includes are still incomplete

panther_manager/include/panther_manager/manager_bt_node.hpp
```diff
@@ -1,38 +1,51 @@
...
 namespace panther_manager
 {
 
-class ManagerNode
+class ManagerBTNode
 {
 public:
-  ManagerNode(std::shared_ptr<ros::NodeHandle> nh);
-  ~ManagerNode() {}
+  ManagerBTNode(const std::shared_ptr<ros::NodeHandle> nh, const std::shared_ptr<ros::NodeHandle> ph);
```
Kotochleb:
```suggestion
  ManagerBTNode(const std::shared_ptr<ros::NodeHandle> & nh, const std::shared_ptr<ros::NodeHandle> & ph);
```

panther_manager/include/panther_manager/manager_bt_node.hpp
```diff
@@ -45,13 +58,23 @@ class ManagerNode
...
+
+  void battery_cb(const sensor_msgs::BatteryState::ConstPtr & battery);
+  void driver_state_cb(const panther_msgs::DriverState::ConstPtr & driver_state);
   void e_stop_cb(const std_msgs::Bool::ConstPtr & e_stop);
   void io_state_cb(const panther_msgs::IOState::ConstPtr & io_state);
+  void system_status_cb(const panther_msgs::SystemStatus::ConstPtr & system_status);
   void safety_tree_timer_cb();
   void lights_tree_timer_cb();
+  void shutdown_robot(const std::string & message);
+  BT::NodeConfig create_bt_config(const std::map<std::string, std::any> bb_values = {}) const;
```
Kotochleb:
```suggestion
  BT::NodeConfig create_bt_config(const std::map<std::string, std::any> & bb_values = {}) const;
```
That was actually copying the whole map before delegating this function

panther_manager/include/panther_manager/plugins/action/call_set_bool_service_node.hpp
```diff
@@ -1,26 +1,31 @@
 #ifndef PANTHER_MANAGER_CALL_SET_BOOL_SERVICE_NODE_HPP_
 #define PANTHER_MANAGER_CALL_SET_BOOL_SERVICE_NODE_HPP_
 
+#include <string>
+
+#include <behaviortree_cpp/basic_types.h>
+#include <behaviortree_cpp/tree_node.h>
+
```
Kotochleb, line 8:
```suggestion

#include <ros/console.h>

```

panther_manager/plugins/action/call_set_led_animation_service_node.cpp
```diff
@@ -1,48 +1,40 @@
 #include <panther_manager/plugins/action/call_set_led_animation_service_node.hpp>
 
+#include <string>
+
+#include <behaviortree_cpp/basic_types.h>
+#include <behaviortree_cpp/exceptions.h>
```
Kotochleb, line 6:
```suggestion
#include <behaviortree_cpp/exceptions.h>

#include <ros/console.h>

```

panther_manager/plugins/action/call_trigger_service_node.cpp
```diff
@@ -1,28 +1,21 @@
 #include <panther_manager/plugins/action/call_trigger_service_node.hpp>
 
-namespace panther_manager
-{
+#include <behaviortree_cpp/basic_types.h>
```
Kotochleb, line 3:
```suggestion
#include <behaviortree_cpp/basic_types.h>

#include <ros/console.h>

```

panther_manager/include/panther_manager/plugins/decorator/tick_after_timeout_node.hpp
```diff
@@ -0,0 +1,32 @@
+#ifndef PANTHER_MANAGER_TICK_AFTER_TIMEOUT_NODE_HPP_
+#define PANTHER_MANAGER_TICK_AFTER_TIMEOUT_NODE_HPP_
+
+#include <string>
+
+#include <behaviortree_cpp/basic_types.h>
+#include <behaviortree_cpp/decorator_node.h>
+#include <behaviortree_cpp/tree_node.h>
+
+#include <ros/ros.h>
```
Kotochleb:
```suggestion
#include <ros/duration.h>
#include <ros/time.h>
```

panther_manager/plugins/decorator/tick_after_timeout_node.cpp
```diff
@@ -0,0 +1,49 @@
+#include <panther_manager/plugins/decorator/tick_after_timeout_node.hpp>
+
+#include <string>
+
+#include <behaviortree_cpp/basic_types.h>
+#include <behaviortree_cpp/tree_node.h>
```
Kotochleb, line 6:
```suggestion
#include <behaviortree_cpp/tree_node.h>

#include <ros/console.h>
#include <ros/duration.h>
#include <ros/time.h>

```

panther_manager/include/panther_manager/plugins/ros_service_node.hpp
```diff
@@ -13,53 +18,57 @@ namespace panther_manager
...
-      throw BT::RuntimeError("[", name, "] Failed to get input [service_name]");
-    }
-
-    unsigned srv_timeout_ms;
-    if (!getInput<unsigned>("timeout", srv_timeout_ms)) {
-      throw BT::RuntimeError("[", name, "] Failed to get input [timeout]");
-    }
-    srv_timeout_ = ros::Duration(static_cast<double>(srv_timeout_ms) * 1e-3);
-
+    nh_ = std::move(config().blackboard->get<std::shared_ptr<ros::NodeHandle>>("nh"));
```
Kotochleb:
```suggestion
    nh_ = std::make_shared(config().blackboard->get<std::shared_ptr<ros::NodeHandle>>("nh"));
```
This should be copying constructor. Not moving the pointer

panther_manager/include/panther_manager/plugins/ros_service_node.hpp
```diff
@@ -13,53 +18,57 @@ namespace panther_manager
...
 
-public:
+  virtual ~RosServiceNode() = default;
+
   using ServiceType = ServiceT;
   using RequestType = typename ServiceT::Request;
   using ResponseType = typename ServiceT::Response;
 
-  static BT::PortsList providedPorts()
+  static BT::PortsList providedBasicPorts(const BT::PortsList addition)
```
Kotochleb:
```suggestion
  static BT::PortsList providedBasicPorts(const BT::PortsList & addition)
```

panther_manager/src/manager_bt_node.cpp
```diff
@@ -1,104 +1,299 @@
...
+#include <panther_msgs/SystemStatus.h>
+
+#include <panther_manager/moving_average.hpp>
+
 namespace panther_manager
 {
 
-ManagerNode::ManagerNode(std::shared_ptr<ros::NodeHandle> nh) : nh_(std::move(nh))
+ManagerBTNode::ManagerBTNode(
+  const std::shared_ptr<ros::NodeHandle> nh, const std::shared_ptr<ros::NodeHandle> ph)
```
Kotochleb:
```suggestion
  const std::shared_ptr<ros::NodeHandle> & nh, const std::shared_ptr<ros::NodeHandle> & ph)
```

panther_manager/include/panther_manager/moving_average.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace panther_manager
+{
+
+template <typename T>
+class MovingAverage
+{
+public:
+  MovingAverage(const unsigned window_size = 5, const T initial_value = T(0))
+  : window_size_(window_size), sum_(initial_value)
```
Kotochleb:
```suggestion
  : window_size_(window_size), sum_(initial_value * static_cast<T>(window_size_))
```
Otherwise, in the beginning `get_average()` will return `1 / window_size_` of the initial_value

Kotochleb (changes requested):
Request and Response types should be explicit and not inherited from parent class typedef

panther_manager/include/panther_manager/plugins/action/call_set_bool_service_node.hpp
KmakD: I believe you meant to add this in `call_set_bool_service_node.cpp` file

Kotochleb: Yes. I have no idea why it ended up here

panther_manager/plugins/decorator/tick_after_timeout_node.cpp
Kotochleb: `#include <ros/console.h>` turned out to be a bit too far-fetched comment. This is not required here

panther_manager/include/panther_manager/plugins/ros_service_node.hpp
Kotochleb: After discussion let's leave it as it is

panther_manager/plugins/action/call_set_led_animation_service_node.cpp
```diff
@@ -1,48 +1,43 @@
...
   if (!getInput<bool>("repeating", repeating)) {
     throw(BT::RuntimeError("[", name(), "] Failed to get input [repeating]"));
   }
-  if (!getInput<std::string>("param", param)) {
-    param = "";
-  }
 
   request.animation.id = animation_id;
-  request.animation.param = param;
+  request.animation.param = getInput<std::string>("param").value();
```
Kotochleb: There should be condition checking if blackboard values is available

Kotochleb (changes requested):

Kotochleb (approved):

PR #105, opened 2023-04-12
Light dimmer

rafal-gorecki:
bump::minor

Reverts https://github.com/husarion/panther_ros/pull/105

### DIFF
--- a/panther_lights/src/animation/battery_animation.py
+++ b/panther_lights/src/animation/battery_animation.py
@@ -3,16 +3,17 @@ import numpy as np
 from PIL import Image, ImageFilter
 
 from animation import Animation
-class BatteryAnimation(Animation):
 
+
+class BatteryAnimation(Animation):
     ANIMATION_NAME = 'battery_animation'
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
 
         self._anim = np.zeros((self._anim_len, self.num_led))
-        self._h_min = 0.0       # red color
-        self._h_max = 0.4       # green color
+        self._h_min = 0.0  # red color
+        self._h_max = 0.4  # green color
         self._resolution = 100
         self._start_fade = 0.80 * self._resolution
         self._end_anim = 0.90 * self._resolution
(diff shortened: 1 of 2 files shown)


panther_lights/src/animation/battery_animation.py
```diff
@@ -6,16 +6,17 @@
...
         super().__init__(*args, **kwargs)
 
         self._anim = np.zeros((self._anim_len, self.num_led))
-        self._h_min = 0.0
-        self._h_max = 120.0
+        self._h_min = 0.0  # red color
+        self._h_max = 0.4  # green color
         self._resolution = 100
+        self._start_fade = 0.80 * self._resolution
+        self._end_anim = 0.90 * self._resolution
```
KmakD: Starting fade before `_end_anim` won't work for `battery_percent = 1.0`. Animation won't have enough time to finish. I think starting fade at `_end_anim` point should be enough to provide smooth fade.
Also, round to one decimal place should be enough.

panther_lights/src/animation/battery_animation.py
```diff
@@ -6,16 +6,17 @@
...
     ANIMATION_NAME = 'battery_animation'
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
 
         self._anim = np.zeros((self._anim_len, self.num_led))
-        self._h_min = 0.0
-        self._h_max = 120.0
+        self._h_min = 0.0  # red color
+        self._h_max = 0.4  # green color
```
KmakD:
```suggestion
        self._h_max = 120.0 / 360.0  # green color
```
For someone familiar with HSV this value is a bit abstract and also actual green would be 0.(3). If we want to keep `hsv_to_rgb` method format I suggest simple division to make it more descriptive.

panther_lights/src/animation/battery_animation.py
```diff
@@ -6,16 +6,17 @@
 
 
 class BatteryAnimation(Animation):
-
     ANIMATION_NAME = 'battery_animation'
```
KmakD: I would keep that line.  It's just a matter of preference, but other animations have spacing  between class and ANIMAITON_NAME

KmakD (changes requested):

panther_lights/src/animation/battery_animation.py
rafal-gorecki: I leave dark panel for last 5% of time, prevents sudden jump

Aayli (approved):

panther_lights/src/animation/battery_animation.py
```diff
@@ -6,16 +6,19 @@
 
 
 class BatteryAnimation(Animation):
-
+    
     ANIMATION_NAME = 'battery_animation'
```
KmakD:
```suggestion

    ANIMATION_NAME = 'battery_animation'
```

KmakD (changes requested):
One formatting issue, the rest LGTM

KmakD (approved):

rafal-gorecki:
Reopen

PR #110, opened 2023-04-19
Thread lock

rafal-gorecki:
bump::patch


panther_battery/src/adc_node.py
```diff
@@ -59,72 +63,75 @@ def __init__(self, name: str) -> None:
...
-            I_bat_1 = self._get_adc_measurement(
-                path='/sys/bus/iio/devices/iio:device1/in_voltage2_raw', LSB=0.04, offset=625.0
-            )
-            I_bat_2 = self._get_adc_measurement(
-                path='/sys/bus/iio/devices/iio:device1/in_voltage1_raw', LSB=0.04, offset=625.0
-            )
-        except:
-            rospy.logerr(f'[{rospy.get_name()}] Battery ADC measurement error')
-            return
+        with self._lock:
```
pkowalsk1: In my opinion, using _lock in this scenario is unnecessary since the local variables are assigned and read within the same thread, much like a local variable, there is no possibility of asynchronous reading and writing. Using lock in this method may increase the time needed for the timer callback to execute.

panther_driver/src/driver_node.py
```diff
@@ -258,129 +261,135 @@ def __init__(self, name: str) -> None:
...
 
     def _reset_roboteq_script_cb(self, req: TriggerRequest) -> TriggerResponse:
-        try:
-            if self._panther_can.restart_roboteq_script():
-                return TriggerResponse(True, 'Roboteq script reset successful.')
-        except Exception as err:
-            return TriggerResponse(False, f'Roboteq script reset failed: \n{err}')
-
-        return TriggerResponse(False, f'Roboteq script reset failed')
+        with self._lock:
```
pkowalsk1: In my opinion, the use of `_lock` in this case is unnecessary since there are no cross-thread operations performed on global variables. Although the use of `Lock` is necessary for the `restart_roboteq_script()` method, it is already implemented within the method itself.

panther_driver/src/driver_node.py
```diff
@@ -258,129 +261,135 @@ def __init__(self, name: str) -> None:
...
+            if any(self._panther_can.can_connection_error()):
+                if not self._e_stop_cliented:
+                    self._trigger_panther_e_stop()
+                    self._stop_cmd_vel_cb = True
+
+                rospy.logerr_throttle(
+                    10.0,
+                    f'[{rospy.get_name()}] Unable to communicate with motor controllers (CAN interface connection failure). '
+                    f'Please ensure that controllers are powered on.',
+                )
```
pkowalsk1:
```suggestion
        with self._lock:
            e_stop_cliented = self._e_stop_cliented
                
        if any(self._panther_can.can_connection_error()):
            if not e_stop_cliented:
                self._trigger_panther_e_stop()
                self._stop_cmd_vel_cb = True

            rospy.logerr_throttle(
                10.0,
                 f'[{rospy.get_name()}] Unable to communicate with motor controllers (CAN interface connection failure). '
                 f'Please ensure that controllers are powered on.',
            )
```
I may be already clinging, but there is no need to use `_lock` in the entire callback method. We can block the program only when we need to read `self._e_stop_clicked` and execute the rest of the code in a multithreaded way. One advantage of this solution may be that we only need to read `self._e_stop_clicked` once which should be faster. However, since it's only Python, it may not make a significant difference :D

Since the methods from the PantherCAN class are already locked, there is no need to worry about using it without `Lock`.

PS. However, there may be a significant issue. As the `self._lock` object is used throughout the entire node, there is a risk that the callback may be blocked and not executed in critical situations due to insufficient time before the next timer callback starts. Is there a possibility that `self._lock` may get blocked and the callback won't be executed?

E.g The `RemoteNode` objects in `PantherCan` (`sdo`) have a timeout when reading data from the CAN interface, which is set to approximately 0.3 seconds. However, due to a communication problem, `_main_timer` and `_driver_state_timer` are unable to execute properly as the time required exceeds the timer period. In this situation, it is necessary to trigger the estop immediately, but can we be certain that the `_safety_timer` will respond promptly?

panther_driver/src/driver_node.py
```diff
@@ -258,129 +261,135 @@ def __init__(self, name: str) -> None:
...
-            self._driver_state_msg.front.left_motor.runtime_error,
-            self._driver_state_msg.rear.right_motor.runtime_error,
-            self._driver_state_msg.rear.left_motor.runtime_error,
-        ] = [
-            self._driver_flag_loggers['runtime_errors'](flag_val)
-            for flag_val in self._panther_can.query_runtime_stat_flag()
-        ]
-
-        self._driver_state_pub.publish(self._driver_state_msg)
+        with self._lock:
```
pkowalsk1: In this scenario, `_driver_state_msg` is declared global solely for the purpose of storing unchanging fields, like `motor_joint_name`. Since there are no multithreaded operations performed on it, using `_lock` would be unnecessary, in my view. Additionally, given the size of this code, it appears to be more efficient to execute it in a multithreaded manner.

The use of separate `Lock`s is appropriate here.

panther_driver/src/driver_node.py
```diff
@@ -258,129 +261,135 @@ def __init__(self, name: str) -> None:
...
-
-        if self._publish_joints:
-            self._publish_joint_state_cb()
-        if self._publish_pose:
-            self._publish_pose_cb()
-        if self._publish_odom:
-            self._publish_odom_cb()
-        if self._publish_tf:
-            self._publish_tf_cb()
+        with self._lock:
```
pkowalsk1: Since there are no multithreaded operations performed on it, using `_lock` would be unnecessary, in my view. It may be reasonable to use `Lock` in `PantherKinematics` i.e `forward_kinematics()` and `inverse_kinematics()`.

If we want to ensure maximum thread safety, we can declare 3 `Lock` objects, one for each timer, and use them only when reading/writing variables used in specific timers:
```python
self._main_timer_msg_lock = Lock()
self._driver_state_msg_lock = Lock()
self._safety_timer_msg_lock = Lock()
```
However, in my opinion, it's not necessary at the moment. It could be considered for future use as our code grows. The decision is up to you.

panther_driver/src/driver_node.py
```diff
@@ -258,129 +261,135 @@ def __init__(self, name: str) -> None:
...
-            self._panther_kinematics.inverse_kinematics(Twist())
+        with self._lock:
+            # Block all motors if any Roboteq controller returns a fault flag or runtime error flag
+            if not self._stop_cmd_vel_cb:
+                self._panther_kinematics.inverse_kinematics(data)
+            else:
+                self._panther_kinematics.inverse_kinematics(Twist())
 
-        self._cmd_vel_command_last_time = rospy.Time.now()
+            self._cmd_vel_command_last_time = rospy.Time.now()
```
pkowalsk1: I wrote about `Lock`s in PantherKinematics. These changes may affect the form of this callback.

panther_lights/src/controller_node.py
```diff
@@ -174,113 +176,116 @@ def __init__(self, name: str) -> None:
...
-            animation.reset_time()
-            animation.repeating = req.repeating
-            self._add_animation_to_queue(animation)
-        except ValueError as err:
-            return SetLEDAnimationResponse(False, f'Failed to add animation to queue: {err}')
-
-        return SetLEDAnimationResponse(
-            True, f'Successfully set an animation with id {req.animation.id}'
-        )
+        with self._lock:
```
pkowalsk1: Is it necessary to use it here, even with `deepcopy()` and `_add_animation_to_queue()`?

panther_lights/src/controller_node.py
```diff
@@ -174,113 +176,116 @@ def __init__(self, name: str) -> None:
...
+                True, f'Successfully set an animation with id {req.animation.id}'
+            )
 
-    def _set_image_animation_cb(
-        self, req: SetLEDImageAnimationRequest
-    ) -> SetLEDImageAnimationResponse:
-        animation = PantherAnimation()
-        animation.repeating = req.repeating
+    def _set_image_animation_cb(self, req: SetLEDImageAnimationRequest) -> SetLEDImageAnimationResponse:
+        with self._lock:
```
pkowalsk1: Same here

panther_lights/src/controller_node.py
```diff
@@ -174,113 +176,116 @@ def __init__(self, name: str) -> None:
...
-        return SetLEDImageAnimationResponse(True, f'Successfully set custom animation')
+            return SetLEDImageAnimationResponse(True, f'Successfully set custom animation')
 
     def _update_animations_cb(self, req: TriggerRequest) -> TriggerResponse:
-        try:
-            self._update_user_animations()
-        except Exception as err:
-            return TriggerResponse(False, f'Failed to update animations: {err}')
-        return TriggerResponse(True, 'Animations updated successfully')
+        with self._lock:
```
pkowalsk1: In my opinion, we should only use `_lock` when reading/writing the `BASIC_ANIMATIONS` dictionary. If we use `_lock` only when there is a risk of updating and reading a global variable at the same time, it will be safer and faster. Personally, I don't think it's necessary to protect the entire callback with `_lock`, but ultimately, it's up to you.

panther_lights/src/scheduler_node.py
```diff
@@ -26,21 +29,17 @@ def __init__(self, name: str) -> None:
         ]
 
         self._critical_battery_anim_period = rospy.get_param('~critical_battery_anim_period', 15.0)
-        self._critical_battery_threshold_percent = rospy.get_param(
-            '~critical_battery_threshold_percent', 0.1
-        )
+        self._critical_battery_threshold_percent = rospy.get_param('~critical_battery_threshold_percent', 0.1)
```
pkowalsk1: Formatting ;)

panther_manager/src/manager_node.py
```diff
@@ -105,22 +108,16 @@ def __init__(self, name: str) -> None:
         # -------------------------------
 
         self._battery_sub = rospy.Subscriber('battery', BatteryState, self._battery_cb)
-        self._driver_state_sub = rospy.Subscriber(
-            'driver/motor_controllers_state', DriverState, self._driver_state_cb
-        )
+        self._driver_state_sub = rospy.Subscriber('driver/motor_controllers_state', DriverState, self._driver_state_cb)
```
pkowalsk1: According to the accepted formatting style, the maximum line length is 100 characters.

panther_manager/src/manager_node.py
```diff
@@ -105,22 +108,16 @@ def __init__(self, name: str) -> None:
...
         self._e_stop_sub = rospy.Subscriber('hardware/e_stop', Bool, self._e_stop_cb)
         self._io_state_sub = rospy.Subscriber('hardware/io_state', IOState, self._io_state_cb)
-        self._system_status_sub = rospy.Subscriber(
-            'system_status', SystemStatus, self._system_status_cb
-        )
+        self._system_status_sub = rospy.Subscriber('system_status', SystemStatus, self._system_status_cb)
```
pkowalsk1: Same as above.

panther_manager/src/manager_node.py
```diff
@@ -105,22 +108,16 @@ def __init__(self, name: str) -> None:
...
 
         # -------------------------------
         #   Service servers
         # -------------------------------
 
-        self._overwrite_fan_control_server = rospy.Service(
-            'manager/overwrite_fan_control', SetBool, self._overwrite_fan_control_cb
-        )
+        self._overwrite_fan_control_server = rospy.Service('manager/overwrite_fan_control', SetBool, self._overwrite_fan_control_cb)
```
pkowalsk1: Same as above.

panther_manager/src/manager_node.py
```diff
@@ -142,85 +139,92 @@ def __init__(self, name: str) -> None:
...
-            )
-        else:
-            self._front_driver_temp_window = [
-                driver_state.front.temperature
-            ] * self._driver_window_len
-            self._rear_driver_temp_window = [
-                driver_state.rear.temperature
-            ] * self._driver_window_len
+        with self._lock:
+            if self._front_driver_temp_window is not None and self._rear_driver_temp_window is not None:
```
pkowalsk1: Line length.

panther_power_control/src/power_board_node.py
```diff
@@ -146,35 +148,40 @@ def __init__(self, name: str) -> None:
...
 
     def _cmd_vel_cb(self, *args) -> None:
-        self._cmd_vel_msg_time = rospy.get_time()
+        with self._lock:
+            self._cmd_vel_msg_time = rospy.get_time()
 
     def _motor_controllers_state_cb(self, msg: DriverState) -> None:
-        self._can_net_err = any({msg.rear.fault_flag.can_net_err, msg.front.fault_flag.can_net_err})
+        with self._lock:
+            self._can_net_err = any({msg.rear.fault_flag.can_net_err, msg.front.fault_flag.can_net_err})
```
pkowalsk1: This variable is used in `_e_stop_reset_cb`, so we should include `_lock` there. 

Additionally, we could consider using different `Lock` objects in various situations, which may help reduce reaction time in critical scenarios and enhance the _safety factor_. However, the decision is ultimately yours.

panther_power_control/src/power_board_node.py
```diff
@@ -146,35 +148,40 @@ def __init__(self, name: str) -> None:
...
-        if rospy.get_time() - self._e_stop_interrupt_time > self._gpio_wait:
-            self._e_stop_event()
-            self._e_stop_interrupt_time = float('inf')
+            if rospy.get_time() - self._e_stop_interrupt_time > self._gpio_wait:
+                self._e_stop_event()
+                self._e_stop_interrupt_time = float('inf')
 
     def _watchdog_cb(self, *args) -> None:
-        self._watchdog()
+        with self._lock:
```
pkowalsk1: Is it necessary to have it here?

panther_power_control/src/power_board_node.py
```diff
@@ -2,6 +2,7 @@
```
pkowalsk1, line 2: I am unable to add a comment in line [267](https://github.com/husarion/panther_ros/blob/37319304dbb76bac243e2d541b75d5f5e6522c0b/panther_power_control/src/power_board_node.py#L267), but I believe that we can use '_lock' there. It is currently functioning, so the decision is up to you. The same situation applies to `_clearing_e_stop`.

panther_power_control/src/relays_node.py
```diff
@@ -73,37 +76,42 @@ def __init__(self, name: str) -> None:
...
+            self._e_stop_state = True
+            self._e_stop_state_pub.publish(self._e_stop_state)
+            return TriggerResponse(True, 'E-SROP triggered successful')
 
     def _set_motor_state_timer_cb(self, *args) -> None:
-        motor_state = GPIO.input(self._pins.STAGE2_INPUT)
-        GPIO.output(self._pins.MOTOR_ON, motor_state)
-        if motor_state != self._motor_on_pub.impl.latch.data:
-            self._motor_on_pub.publish(motor_state)
+        with self._lock:
```
pkowalsk1: You need to consider what should be included in methods like this, as the implementation of `_e_stop_event` method in `power_board_node.py` differs.

panther_battery/src/adc_node.py
```diff
@@ -59,72 +63,75 @@ def __init__(self, name: str) -> None:
...
+            self._V_driv_front = driver_state.front.voltage
+            self._V_driv_rear = driver_state.front.current
+            self._I_driv_front = driver_state.rear.voltage
+            self._I_driv_rear = driver_state.rear.current
 
     def _io_state_cb(self, io_state: IOState) -> None:
-        self._charger_connected = io_state.charger_connected
+        with self._lock:
+            self._charger_connected = io_state.charger_connected
```
pkowalsk1, line 88: The variable _charger_connected is read within the _publish_battery_msg method, so we can utilize a `_lock` there.

pkowalsk1 (changes requested):

panther_battery/src/adc_node.py
rafal-gorecki: Added

pkowalsk1 (approved):

PR #115, opened 2023-05-10
Power Control rewritten to Libgpiod

Kotochleb:
bump::patch

Remove RPi.GPIO and use Libgpiod for GPIO control.
~Currently, the implementation assumes hard-coded pin numbers for older robots.~


panther_power_control/src/power_board_node.py
```diff
@@ -52,22 +32,50 @@ def turn_on(self) -> None:
...
 
         self._clearing_e_stop = False
-        self._pins = PatherGPIO()
 
-        self._setup_gpio()
+        out_line_names = [
+            'AUX_PW_EN',  # Enable auxiliary power, eg. supply to robotic arms etc.
+            'CHRG_DISABLE',  # Disable charger
+            'DRIVER_EN',  # Enable motor drivers (1 - on)
+            # OUT - send 1 to reset estop
```
KmakD: i guess this refers to `E_STOP_RESET` pin

panther_power_control/src/power_board_node.py
```diff
@@ -52,22 +32,50 @@ def turn_on(self) -> None:
...
+            'FAN_SW',  # Turn on the fan (1 - on)
+            'VDIG_OFF',  # Turn the digital power off eg. NUC, Router etc. (1 - off)
+            'VMOT_ON',  # Enable mamin power supply to motors (1 - on)
+            'WATCHDOG',  # Watchdog pin, if PWM is on this pin Panther will work
+        ]
+        in_line_names = [
+            'CHRG_SENSE',  # Charger sensor (1 - charger plugged in)
+            'E_STOP_RESET',  # Works as IN/OUT, IN - gives info if E-stop in on (1 - off),
```
KmakD:
```suggestion
            'E_STOP_RESET',  # Works as IN/OUT, IN - gives info if E-stop is on (1 - off),
```

panther_power_control/src/power_board_node.py
```diff
@@ -52,22 +32,50 @@ def turn_on(self) -> None:
...
+        self._lines = {name: self._chip.find_line(name) for name in out_line_names + in_line_names}
+        for name, line in self._lines.items():
+            req_type = gpiod.LINE_REQ_DIR_OUT if name in out_line_names else gpiod.LINE_REQ_DIR_IN
+            req_flag = gpiod.LINE_REQ_FLAG_ACTIVE_LOW if name in inv_lines else 0
+            line.request(self._node_name, type=req_type, flags=req_flag)
+
+        self._lines['CHRG_DISABLE'].set_value(True)
+        self._lines['AUX_PW_EN'].set_value(False)
+        self._lines['FAN_SW'].set_value(False)
+        self._lines['VDIG_OFF'].set_value(True)
```
KmakD: That should be false by default. Shouldn't we also defined initial states for `DRIVER_EN`, `VMOT_ON` as it was done previously?

panther_power_control/src/power_board_node.py
```diff
@@ -171,54 +171,52 @@ def _motor_controllers_state_cb(self, msg: DriverState) -> None:
...
-                    rospy.loginfo(f'[{rospy.get_name()}] Shutdown button pressed.')
-                    self._publish_io_state('power_button', True)
-                self._chrg_sense_interrupt_time = float('inf')
-
-            if rospy.get_time() - self._e_stop_interrupt_time > self._gpio_wait:
-                self._e_stop_event()
-                self._e_stop_interrupt_time = float('inf')
+            if self._lines['SHDN_INIT'].get_value():
+                if rospy.get_time() - self._chrg_sense_interrupt_time > self._gpio_wait:
+                    if self._lines['SHDN_INIT'].get_value():
```
KmakD: same conditions repeated

panther_power_control/src/power_board_node.py
```diff
@@ -171,54 +171,52 @@ def _motor_controllers_state_cb(self, msg: DriverState) -> None:
...
+                    self._chrg_sense_interrupt_time = float('inf')
+                else:
+                    self._chrg_sense_interrupt_time = rospy.get_time()
+
+            if self._lines['E_STOP_RESET'].get_value():
+                if rospy.get_time() - self._e_stop_interrupt_time > self._gpio_wait:
+                    self._e_stop_event()
+                    self._e_stop_interrupt_time = float('inf')
+                else:
+                    self._e_stop_interrupt_time = rospy.get_time()
```
KmakD: looking at condition this will only trigger when resetting e-stop

panther_power_control/src/power_board_node.py
```diff
@@ -171,54 +171,52 @@ def _motor_controllers_state_cb(self, msg: DriverState) -> None:
...
-        res = self._set_bool_srv_handle(req.data, self._pins.VDIG_OFF, 'Digital power enable')
+        res = self._set_bool_srv_handle(req.data, 'VDIG_OFF', 'Digital power enable')
         if res.success:
             self._publish_io_state('digital_power', req.data)
         return res
 
     def _e_stop_reset_cb(self, req: TriggerRequest) -> TriggerResponse:
         with self._e_stop_lock:
-            if self._validate_gpio_pin(self._pins.E_STOP_RESET, False):
+            if self._lines['E_STOP_RESET'].get_value() == False:
```
KmakD:
```suggestion
            if not self._lines['E_STOP_RESET'].get_value():
```

panther_power_control/src/power_board_node.py
```diff
@@ -233,7 +231,7 @@ def _e_stop_reset_cb(self, req: TriggerRequest) -> TriggerResponse:
 
         self._reset_e_stop()
 
-        if self._validate_gpio_pin(self._pins.E_STOP_RESET, True):
+        if self._lines['E_STOP_RESET'].get_value() == True:
```
KmakD:
```suggestion
        if self._lines['E_STOP_RESET'].get_value():
```

panther_power_control/src/power_board_node.py
```diff
@@ -292,19 +290,26 @@ def _set_bool_srv_handle(self, value: bool, pin: int, name: str) -> SetBoolRespo
 
     def _reset_e_stop(self) -> None:
         self._clearing_e_stop = True
-        GPIO.setup(self._pins.E_STOP_RESET, GPIO.OUT)
+
+        req_type = gpiod.LINE_REQ_DIR_OUT
+        req_flag = gpiod.LINE_REQ_FLAG_ACTIVE_LOW
```
KmakD: Is there a way to assign this using already defined pin? This way we don't have to worry if pin is in inverse logic. Something like:
```
req_flag = self._lines['E_STOP_RESET'].type()
```

panther_power_control/src/relays_node.py
```diff
@@ -13,25 +13,21 @@
...
 
         self._lock = Lock()
 
-        self._pins = PatherGPIO()
-        self._setup_gpio()
+        chip = gpiod.Chip('gpiochip0', gpiod.Chip.OPEN_BY_NAME)
 
-        self._e_stop_state = not GPIO.input(self._pins.STAGE2_INPUT)
+        line_names = {'MOTOR_ON': 6, 'STAGE2_INPUT': 22}
+        self._lines = {name: chip.get_line(line_names[name]) for name in line_names}
```
KmakD: In power_board_node.py you use list and `chip.find_line()`. Is there a reason for this to be different between files?

panther_power_control/src/power_board_node.py
```diff
@@ -171,54 +173,52 @@ def _motor_controllers_state_cb(self, msg: DriverState) -> None:
...
+                    if self._lines['SHDN_INIT'].get_value():
+                        rospy.loginfo(f'[{rospy.get_name()}] Shutdown button pressed.')
+                        self._publish_io_state('power_button', True)
+                    self._chrg_sense_interrupt_time = float('inf')
+                elif math.isinf(self._chrg_sense_interrupt_time):
+                    self._chrg_sense_interrupt_time = rospy.get_time()
+
+            if self._lines['E_STOP_RESET'].get_value() != self._last_e_stop_state:
+                if rospy.get_time() - self._e_stop_interrupt_time > self._gpio_wait:
+                    self._e_stop_event()
```
KmakD: The first condition is not necessary as it is checked in `_e_stop_event()` method

KmakD (changes requested):

panther_power_control/src/power_board_node.py
Kotochleb: I checked it once again and could not find anything similar in current API

panther_power_control/src/relays_node.py
Kotochleb: "Currently, the implementation assumes hard-coded pin numbers for older robots." ~ PR description

panther_power_control/src/power_board_node.py
Kotochleb: Fixed

panther_power_control/src/power_board_node.py
```diff
@@ -52,26 +33,56 @@ def turn_on(self) -> None:
...
+
+        self._watchdog = Watchdog(self._lines['WATCHDOG'])
         self._motor_start_sequence()
-        self._watchdog = Watchdog(self._pins.WATCHDOG)
 
         self._gpio_wait = 0.05  # seconds
+        self._last_e_stop_state = not self._lines['E_STOP_RESET'].get_value()
         self._e_stop_interrupt_time = float('inf')
-        self._chrg_sense_interrupt_time = float('inf')
+        self._chrg_sense_pressed_time = float('inf')
```
KmakD: This relates to shutdown button, shouldn't it be sth like `shdn_init_pressed_time`

panther_power_control/src/power_board_node.py
```diff
@@ -50,38 +50,39 @@ def __init__(self, name: str) -> None:
...
+                self._node_name,
+                gpiod.LINE_REQ_DIR_OUT if name in out_line_names else gpiod.LINE_REQ_DIR_IN,
+            )
 
         self._lines['CHRG_DISABLE'].set_value(True)
         self._lines['AUX_PW_EN'].set_value(False)
         self._lines['FAN_SW'].set_value(False)
-        self._lines['VDIG_OFF'].set_value(True)
+        # This pin is inverted so default has to be high
+        self._lines['VDIG_OFF'].set_value(False)
```
KmakD: this comment does not make sense

panther_power_control/src/power_board_node.py
```diff
@@ -52,26 +33,56 @@ def turn_on(self) -> None:
...
+            'AUX_PW_EN',  # Enable auxiliary power, eg. supply to robotic arms etc.
+            'CHRG_DISABLE',  # Disable charger
+            'DRIVER_EN',  # Enable motor drivers (1 - on)
+            'FAN_SW',  # Turn on the fan (1 - on)
+            'VDIG_OFF',  # Turn the digital power off eg. NUC, Router etc. (1 - off)
+            'VMOT_ON',  # Enable mamin power supply to motors (1 - on)
+            'WATCHDOG',  # Watchdog pin, if PWM is on this pin Panther will work
+        ]
+        in_line_names = [
+            'CHRG_SENSE',  # Charger sensor (1 - charger plugged in)
```
KmakD: This is active low pin, which means 1 - charger not connected
Set this using LINE_REQ_FLAG_ACTIVE_LOW or make clear comment that this pin is in inverted logic

KmakD (changes requested):

panther_power_control/src/relays_node.py
```diff
@@ -13,25 +12,25 @@
...
+        self._node_name = name
+        rospy.init_node(self._node_name, anonymous=False)
 
         self._lock = Lock()
 
-        self._pins = PatherGPIO()
-        self._setup_gpio()
+        self._chip = gpiod.Chip('gpiochip0', gpiod.Chip.OPEN_BY_NAME)
+
+        line_names = {'MOTOR_ON': 6, 'STAGE2_INPUT': 22}
```
KmakD:
```suggestion
        line_names = {
            'MOTOR_ON': 6,
            'STAGE2_INPUT': 22,
        }
```
Personally I like this formatting for dicts - with `,` after last element

KmakD (changes requested):

KmakD (approved):

panther_power_control/src/power_board_node.py
```diff
@@ -52,26 +32,59 @@ def turn_on(self) -> None:
...
-        self._pins = PatherGPIO()
 
-        self._setup_gpio()
+        out_line_names = {
+            'AUX_PW_EN': False,  # Enable auxiliary power, eg. supply to robotic arms etc.
+            'CHRG_DISABLE': True,  # Disable charger
+            'DRIVER_EN': False,  # Enable motor drivers (1 - on)
+            'FAN_SW': False,  # Turn on the fan (1 - on)
+            'VDIG_OFF': False,  # Turn the digital power off eg. NUC, Router etc. (1 - off)
+            'VMOT_ON': False,  # Enable mamin power supply to motors (1 - on)
```
macstepien:
```suggestion
            'VMOT_ON': False,  # Enable main power supply to motors (1 - on)
```

panther_power_control/src/power_board_node.py
```diff
@@ -171,54 +179,62 @@ def _motor_controllers_state_cb(self, msg: DriverState) -> None:
...
     def _aux_power_enable_cb(self, req: SetBoolRequest) -> SetBoolResponse:
-        res = self._set_bool_srv_handle(req.data, self._pins.AUX_PW_EN, 'Aux power enable')
+        res = self._set_bool_srv_handle(req.data, 'AUX_PW_EN', 'Aux power enable')
         if res.success:
             self._publish_io_state('aux_power', req.data)
         return res
 
     def _charger_enable_cb(self, req: SetBoolRequest) -> SetBoolResponse:
-        res = self._set_bool_srv_handle(not req.data, self._pins.CHRG_DISABLE, 'Charger disable')
+        res = self._set_bool_srv_handle(not req.data, 'CHRG_DISABLE', 'Charger disable')
```
macstepien, line 230: this is up for discussion, but I would prefer having type instead of string (`PatherGPIO` but with strings instead of numbers), because it's harder to make a typo. If we have a type, where values `pins.CHRG_DISABLE` are mapped to string `CHRG_DISABLE` and use `pins.CHRG_DISABLE` it will be harder to make a mistake.

panther_power_control/src/relays_node.py
```diff
@@ -13,25 +12,28 @@
...
 
         self._lock = Lock()
 
-        self._pins = PatherGPIO()
-        self._setup_gpio()
+        self._chip = gpiod.Chip('gpiochip0', gpiod.Chip.OPEN_BY_NAME)
+
+        line_names = {
+            'MOTOR_ON': 6,
+            'STAGE2_INPUT': 22,
```
macstepien: comments are missing, and again I think that `PatherGPIO` class was better

macstepien (changes requested):
I think that python code wasn't formatted with `black`

panther_power_control/src/power_board_node.py
macstepien: ok to leave with strings, using wrong name will cause error

panther_power_control/src/relays_node.py
```diff
@@ -13,25 +12,35 @@
...
-        rospy.init_node(name, anonymous=False)
+        self._node_name = name
+        rospy.init_node(self._node_name, anonymous=False)
 
         self._lock = Lock()
 
-        self._pins = PatherGPIO()
-        self._setup_gpio()
+        line_names = {
+            'MOTOR_ON': False, # Used to enable SSR2
```
macstepien: what's the `SSR2`? :D

macstepien:
> I think that python code wasn't formatted with `black`

bump

panther_power_control/src/relays_node.py
Kotochleb: It was copied from documentation, anyway I changed it

macstepien (approved):

macstepien (approved):

PR #117, opened 2023-05-11
Ros1 manager bt signal shutdown

KmakD:


panther_manager/include/panther_manager/plugins/action/signal_shutdown_node.hpp
```diff
@@ -0,0 +1,32 @@
...
+{
+public:
+  explicit SignalShutdown(const std::string & name, const BT::NodeConfig & conf)
+  : BT::SyncActionNode(name, conf)
+  {
+  }
+
+  static BT::PortsList providedPorts()
+  {
+    return {BT::InputPort<std::string>("message", "", "reason to shutdown robot")};
```
Kotochleb: Add `OutputPort<std::pair<bool, std::string>>("signal_shutdown")` to make to code more self exlpanatory

panther_manager/src/manager_bt_node.cpp
```diff
@@ -268,15 +268,16 @@ void ManagerBTNode::safety_tree_timer_cb()
...
+  if (safety_config_.blackboard->get<std::pair<bool, std::string>>(
+        "signal_shutdown", signal_shutdown)) {
+    if (signal_shutdown.first) shutdown_robot(signal_shutdown.second);
   }
 }
 
 void ManagerBTNode::shutdown_robot(const std::string & message)
 {
-  ROS_WARN("[%s] %s. Soft shutdown initialized.", node_name_.c_str(), message.c_str());
+  ROS_WARN("[%s] Soft shutdown initialized. %s", node_name_.c_str(), message.c_str());
```
Kotochleb: Instead of message, can we use word "reason"?

panther_manager/config/PantherManagerBT.xml
```diff
@@ -178,6 +177,10 @@ current_anim_id = CHARGING_BATTERY_ANIM_ID"/>
       <input_port name="ip">ip of the host to shutdown</input_port>
       <input_port name="user">user to log into while executing shutdown command</input_port>
     </Action>
+    <Action ID="SignalShutdown"
+            editable="true">
+      <input_port name="message">reason to shutdown robot</input_port>
```
Kotochleb:
```suggestion
      <input_port name="reason">reason to shutdown robot</input_port>
```
How about naming it reason?

Kotochleb (changes requested):

Kotochleb (changes requested):
I was wrong. Using ports was a bad idea. Can you bring back the blackboard approach?

Kotochleb (approved):

PR #118, opened 2023-05-19
Ros1 check bat current

pkowalsk1:
bump::patch


panther_battery/src/roboteq_republisher_node.py
```diff
@@ -79,7 +81,9 @@ def _battery_pub_timer_cb(self, *args) -> None:
                 battery_msg.voltage = self._battery_voltage
                 battery_msg.temperature = float('nan')
                 battery_msg.current = self._battery_current
-                battery_msg.percentage = (battery_msg.voltage - 32.0) / 10.0
+                battery_msg.percentage = (battery_msg.voltage - self.V_BAT_MIN) / (
+                    self.V_BAT_FULL - self.V_BAT_MIN
+                )
```
Kotochleb: Since we say that the fatal voltage is 27 V we will end up with -0.53 % of charge in case of failure. I think this should be saturated in case of exceeding 1 and going below 0. Also If we reach 27 V hardware will turn of the robot

panther_battery/src/adc_node.py
```diff
@@ -15,9 +13,12 @@
 
 
 class ADCNode:
+    BAT01_CHARGING_CURR_THRESH = 0.1
```
Kotochleb:
```suggestion
    BAT_CHARGING_CURR_THRESH = 0.1
```
Since this parameter is used for bot batteries, I don't think there is the point specifying that `01` indicating its assignment to the battery 1

panther_battery/src/adc_node.py
```diff
@@ -181,36 +192,38 @@ def _voltage_to_deg(self, V_temp: float) -> float:
...
     ) -> None:
         battery_msg = BatteryState()
         battery_msg.header.stamp = rospy.Time.now()
         battery_msg.voltage = V_bat
         battery_msg.temperature = temp_bat
         battery_msg.current = I_bat
-        battery_msg.percentage = (battery_msg.voltage - 32.0) / 10.0
+        battery_msg.percentage = (battery_msg.voltage - self.V_BAT_MIN) / (
+            self.V_BAT_FULL - self.V_BAT_MIN
+        )
```
Kotochleb: Same as in republisher node

Kotochleb (changes requested):

panther_battery/src/adc_node.py
```diff
@@ -240,19 +250,35 @@ def _count_volt_mean(self, label: Union[rospy.Publisher, str], new_val: float) -
...
+        # in order to avoid recalculating the entire sum every time.
+        self._I_bat_charge_mean[label] += (
+            new_val - self._I_bat_charge_hist[label][0]
+        ) / self._curr_mean_length
+
+        self._I_bat_charge_hist[label].pop(0)
+        self._I_bat_charge_hist[label].append(new_val)
+
+        return self._I_bat_charge_mean[label]
+
```
KmakD: This method looks exactly as `_count_volt_mean`. Maybe this should be one method with additional arguments?

KmakD (changes requested):

Kotochleb (approved):

pkowalsk1:
@KmakD I merged branch 'ros1' into `ros1-check-bat-current`

KmakD (approved):

PR #121, opened 2023-05-29
Ros1 manager bt update

KmakD:
This PR looks big but most of changes was just moving code around and also it was necessary to merge branch ros1. Main changes were:
- changed BT project organization
- added params for launching each tree
- added BT project for panther v1.06 and below
- removed scheduler node from panther_lights
- fixed animation queuing for anim with priority 2 and below

Files to check:
- `panther_manager` package
- `panther_bringup/bringup.launch`
- `panther_lights/lights.launch`
- `panther_lights/controller_node.py` - line 307 (this commit: https://github.com/husarion/panther_ros/pull/121/commits/b4149e19875e7e09e1ff8abf4268672c0950c2be)


panther_manager/launch/manager_bt.launch
```diff
@@ -1,14 +1,18 @@
 <launch>
   <arg name="panther_version" />
+  <arg name="bt_project_file" />
```
rafal-gorecki, line 3: Why are you does not use default? This would allow for a quick launch of a separate launch and testing some of the functionality.

panther_manager/src/manager_bt_node.cpp
```diff
@@ -90,58 +92,73 @@ ManagerBTNode::ManagerBTNode(
...
-  lights_tree_ = factory_.createTree("Lights", lights_config_.blackboard);
-  safety_tree_ = factory_.createTree("Safety", safety_config_.blackboard);
-  shutdown_tree_ = factory_.createTree("Shutdown", shutdown_config_.blackboard);
+  factory_.registerBehaviorTreeFromFile(bt_project_file);
+
+  if (launch_safety_tree && !launch_shutdown_tree_) {
+    ROS_ERROR(
+      "[%s] Can't launch safety tree without shutdown tree. Killing node.", node_name_.c_str());
+    ros::requestShutdown();
+  }
```
rafal-gorecki, line 101: I don't know how BT looks like, so I don't understand why this should be met. I see there is empty space in README.md for [BT}(https://github.com/husarion/panther_ros/tree/ros1-manager-bt-update/panther_manager#behaviortree)

rafal-gorecki (commented):
Code look fine for me, but I don't know how tree looks like. Could you add image in readme?
I will test it on Panther in Monday.

panther_manager/launch/manager_bt.launch
KmakD: All `panther_ros` packages use this approach, default values are provided only in `panther_bringup` package.

panther_manager/src/manager_bt_node.cpp
KmakD: One of Safety tree tasks is to signal robot shutdown in case of reaching fatal battery temperature.  This requires the Shutdown tree to be present.

rafal-gorecki:
Zmiany przetestowane, nie zauważono błędów.

PR #125, opened 2023-06-02
ROS 2 panther utils package

KmakD:


panther_utils/include/panther_utils/moving_average.hpp
```diff
@@ -0,0 +1,46 @@
+#ifndef PANTHER_UTILS_MOVING_AVERAGE_HPP_
```
Kotochleb, line 1: There could be a reset function to make it more general. The reset might set the whole internal queue to some value rather than zero. That will also require additional test case

panther_utils/test/test_moving_average.cpp
```diff
@@ -0,0 +1,48 @@
...
+  ma->Roll(5.0);
+  ma->Roll(5.0);
+  EXPECT_EQ(5.0, ma->GetAverage());
+}
+
+TEST(TestMovingAverage, TestIntialValue)
+{
+  std::unique_ptr<panther_utils::MovingAverage<float>> ma;
+  ma = std::make_unique<panther_utils::MovingAverage<float>>(4, 1.0);
+  EXPECT_EQ(1.0, ma->GetAverage());
```
Kotochleb, line 32: Missing additional `Roll` execution to check internal queue

panther_utils/test/test_moving_average.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+TEST(TestMovingAverage, TestIntType)
+{
+  std::unique_ptr<panther_utils::MovingAverage<int>> ma;
+  ma = std::make_unique<panther_utils::MovingAverage<int>>();
+  ma->Roll(1);
+  ma->Roll(2);
+  EXPECT_EQ(1, ma->GetAverage());
```
Kotochleb: Comments before tests could be nice. In this case, I am struggling to understand what exactly is tested here

Kotochleb (changes requested):

panther_utils/test/test_moving_average.cpp
KmakD: It is stated in Test name: TEST(TestMovingAverage, TestIntType)

panther_utils/include/panther_utils/moving_average.hpp
```diff
@@ -0,0 +1,51 @@
...
+    if (values_.size() > window_size_) {
+      sum_ -= values_.front();
+      values_.pop_front();
+    }
+  }
+
+  void Reset() {
+    values_.erase(values_.begin(), values_.end());
+    sum_ = T(0);
+  }
```
Kotochleb, line 32: In comment, I mentioned it not to reset to zero, but set value

panther_utils/test/test_moving_average.cpp
Kotochleb: Fair enough

Kotochleb (changes requested):
Tests do not have to assume the class is inside smart pointer. It can be just a calss itself

panther_utils/test/test_moving_average.cpp
```diff
@@ -0,0 +1,91 @@
...
+  ma.Reset();
+  EXPECT_EQ(0.0, ma.GetAverage());
+
+  ma.Roll(2.0);
+  ma.Roll(4.0);
+  EXPECT_EQ(3.0, ma.GetAverage());
+
+  panther_utils::MovingAverage<double> ma_1(4, 7.0);
+  ma_1.Roll(1.0);
+  ma_1.Roll(2.0);
```
Kotochleb:
```suggestion
  ma_1.Roll(2.0);
  EXPECT_EQ(4.25, ma_1.GetAverage());
```

panther_utils/test/test_moving_average.cpp
```diff
@@ -0,0 +1,91 @@
...
+  ma.Roll(5.0);
+  EXPECT_EQ(5.0, ma.GetAverage());
+}
+
+TEST(TestMovingAverage, TestHighOverload)
+{
+  panther_utils::MovingAverage<double> ma(1000);
+
+  double sum;
+  for (int i = 1; i <= 10000; i++) {
```
Kotochleb:
```suggestion
  for (std::size_t i = 1; i <= 10000; i++) {
```

panther_utils/test/test_moving_average.cpp
```diff
@@ -0,0 +1,91 @@
...
+    sum += double(i);
+    ma.Roll(double(i));
+
+    // test every 1000 rolls expected average
+    if (i % 1000 == 0) {
+      EXPECT_EQ(sum / 1000, ma.GetAverage());
+      sum = 0.0;
+    }
+  }
```
Kotochleb:
```suggestion
  const std::size_t window_len = 1000;
  double sum;
  for (int i = 1; i <= window_len * 10; i++) {
    sum += double(i);
    ma.Roll(double(i));

    // test every 1000 rolls expected average
    if (i % window_len == 0) {
      EXPECT_EQ(sum / double(window_len), ma.GetAverage());
      sum = 0.0;
    }
  }
```

panther_utils/include/panther_utils/moving_average.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include <deque>
+
+namespace panther_utils
+{
+
+template <typename T>
+class MovingAverage
+{
+public:
+  MovingAverage(const unsigned window_size = 5, const T initial_value = T(0))
```
Kotochleb:
```suggestion
  MovingAverage(const std::size_t window_size = 5, const T initial_value = T(0))
```
Maybe `std::size_t` will suit better in this case. In general this is `unsigned`, but in more suitable way

Kotochleb (changes requested):

panther_utils/test/test_moving_average.cpp
KmakD: This part  of the task is supposed to check if average is reset to initial value. I don't see a point in checking returned values as they were tested before in this task. I added a comment to make it a bit more clear. Plus, initial value is only returned when no roll commands were called (it is not added to sum) so expected value in this case won't be 4.25.

Kotochleb: This is to make sure the difference occurred after reset. If the roll for some reason might stop working, the test will pass

Kotochleb (changes requested):

panther_utils/test/test_moving_average.cpp
```diff
@@ -0,0 +1,93 @@
+#include <gtest/gtest.h>
+
+#include <panther_utils/moving_average.hpp>
+
+TEST(TestMovingAverage, TestOutputValues)
+{
+  panther_utils::MovingAverage<double> ma(4);
+  EXPECT_EQ(0.0, ma.GetAverage());
```
pawelirh, line 8: The first check looks like a separate case `TestDefaultInitialValue`

panther_utils/test/test_moving_average.cpp
```diff
@@ -0,0 +1,93 @@
...
+  EXPECT_EQ(1.0, ma.GetAverage());
+
+  panther_utils::MovingAverage<double> ma_1(10, 3.7);
+  EXPECT_EQ(3.7, ma_1.GetAverage());
+
+  panther_utils::MovingAverage<int> ma_2(3, 4);
+  EXPECT_EQ(4, ma_2.GetAverage());
+}
+
+TEST(TestMovingAverage, TestIntType)
```
pawelirh: Shouldn't it be `TestIntFloorRound` or another more self-explanatory name?

panther_utils/test/test_moving_average.cpp
```diff
@@ -0,0 +1,93 @@
...
+  ma.Roll(2.0);
+  ma.Roll(4.0);
+  EXPECT_EQ(3.0, ma.GetAverage());
+
+  // check if average resets to initial value
+  panther_utils::MovingAverage<double> ma_1(4, 7.0);
+  ma_1.Roll(1.0);
+  ma_1.Roll(2.0);
+  ma_1.Reset();
+  EXPECT_EQ(7.0, ma_1.GetAverage());
```
pawelirh: Reset initial value should be separated

Kotochleb: Ok

Kotochleb (approved):
LGTM leaving it to @pawelirh

PR #127, opened 2023-06-15
Ekf localization

rafal-gorecki:
bump::patch


panther_bringup/config/ekf_config.yaml
```diff
@@ -0,0 +1,75 @@
+# Ref: http://docs.ros.org/en/noetic/api/robot_localization/html/state_estimation_nodes.html
+
+frequency: 20
```
pkowalsk1, line 3: Given that the lowest input frequency is 15Hz, is there a need to update the EKF at 20Hz? Do we want to rely on the IMU, despite its higher frequency, considering the higher interference it experiences?

rafal-gorecki: We KF always relay only current state and sensor, can be used to slightly increase the frequency of publication. I found 20 Hz to be a nice frequency.

panther_driver/src/driver_node.py
```diff
@@ -170,7 +171,7 @@ def __init__(self, name: str) -> None:
         if self._publish_odom:
             self._odom_msg = Odometry()
             self._odom_msg.pose.covariance = [0.1 if (i % 7) == 0 else 0.0 for i in range(36)]
-            self._odom_msg.twist.covariance = [0.1 if (i % 7) == 0 else 0.0 for i in range(36)]
+            self._odom_msg.twist.covariance = np.diag([1e-5, 1e-5, 1e6, 1e6, 1e6, 7.2e-5]).flatten().tolist()
```
macstepien: It would be better to move it to a parameter

panther_driver/src/driver_node.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/python3
 
+import numpy as np
```
macstepien: I think that we wanted to keep dependencies minimal, so including numpy just to create a matrix isn't really optimal

panther_bringup/config/imu_config.yaml
```diff
@@ -1,18 +1,25 @@
...
 PhidgetsSpatialNodelet:
   data_interval_ms: 8 # supported data rates: 4 8 16 24 32 40 ... 1000 (in ms)
   publish_rate: 100.0 # optional param publish_rate, defaults to 0
   # serial: 123456  # optional param serial, default is -1
   # hub_port: 0 # optional param hub_port, used if connected to a VINT hub
   # frame_id: imu_link  # optional param frame_id, default is "imu_link"
-  # linear_acceleration_stdev: 0.002745862  # optional param linear_acceleration_stdev, default is 280ug
-  # angular_velocity_stdev: 0.000349  # optional param angular_velocity_stdev, default is 0.095 deg/s
+  linear_acceleration_stdev: 2.4e-1  # optional param linear_acceleration_stdev, default is 280ug
+  angular_velocity_stdev: 3.0e-2  # optional param angular_velocity_stdev, default is 0.095 deg/s
```
macstepien: These values seems to be quite large, where do they come from? I think that it would be best to use ones from datasheet. Especially `angular_velocity_stdev` seems to be large when compared to the value from wheels, from my experience usually IMU data is more accurate than wheels one (when orientation is considered). Did you try making a loop (marking a starting position and returning to it) with different EKF configurations (only wheel odom orientation, only IMU orientation, fused orientation ...) ?

panther_driver/src/driver_node.py
macstepien: Also I think that according to your calculations y value should be 2e-6 and probably these parameters for diff drive and mecanum should be separated.

panther_bringup/config/imu_config.yaml
```diff
@@ -1,18 +1,25 @@
...
+  gain: 0.1
+  zeta: 0.0
+  mag_bias_x: 0.0
+  mag_bias_y: 0.0
+  mag_bias_z: 0.0
+  orientation_stddev: 1e-2
   publish_tf: false
   use_mag: false
-  use_magnetic_field_msg: false
+  use_magnetic_field_msg: true
```
macstepien: Will magnetometer data be useful in our case? IMU is quite close to motors and additionally it closed in the robot body, so I think that it won't be really accurate.

panther_bringup/config/ekf_config.yaml
```diff
@@ -0,0 +1,75 @@
...
+                           0,    0,    0,    0,    0,    5e-3, 0,    0,    0,    0,    0,    0,    0,    0,    0,
+                           0,    0,    0,    0,    0,    0,    2e-6, 0,    0,    0,    0,    0,    0,    0,    0,
+                           0,    0,    0,    0,    0,    0,    0,    2e-6, 0,    0,    0,    0,    0,    0,    0,
+                           0,    0,    0,    0,    0,    0,    0,    0,    2e-6, 0,    0,    0,    0,    0,    0,
+                           0,    0,    0,    0,    0,    0,    0,    0,    0,    5e-4, 0,    0,    0,    0,    0,
+                           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    5e-4, 0,    0,    0,    0,
+                           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    5e-4, 0,    0,    0,
+                           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    5e-3, 0,    0,
+                           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    5e-3, 0,
+                           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    5e-3]
```
macstepien: would be nice to add some explanation for these values

macstepien (changes requested):
Generally it would be nice to add more comments on why setting was chosen, where values come from

panther_bringup/config/ekf_config.yaml
rafal-gorecki: komentarz sposób obliczania

panther_bringup/config/imu_config.yaml
rafal-gorecki: data corrected to data from the catalog note, taking into account the drift https://www.phidgets.com/?tier=3&catid=10&pcid=8&prodid=1205

panther_bringup/config/imu_config.yaml
rafal-gorecki: this change only message type to correct one, dont publish mag data (`use_mag: false` is still on false)

panther_driver/src/driver_node.py
```diff
@@ -170,7 +173,13 @@ def __init__(self, name: str) -> None:
         if self._publish_odom:
             self._odom_msg = Odometry()
             self._odom_msg.pose.covariance = [0.1 if (i % 7) == 0 else 0.0 for i in range(36)]
-            self._odom_msg.twist.covariance = [0.1 if (i % 7) == 0 else 0.0 for i in range(36)]
+
+            twist_cov = [0.0] * 36
+            cov_value = [self._v_x_var, self._v_y_var, 1e6, 1e6, 1e6, self._v_yaw_var]
+            for i in range (len(cov_value)):
+                twist_cov[7*i] = cov_value[i]
```
macstepien: I think that it would be simpler to just assign values directly with index, instead of for loop

panther_bringup/config/ekf_config.yaml
macstepien: It could be also set to 20, but during our test we used 15 to tune other values, so I think that we should use 15 then.

panther_bringup/config/ekf_config.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+imu0_remove_gravitational_acceleration: true
+
+reset_on_time_jump: false
+predict_to_current_time: false
+print_diagnostics: false
+debug: false
+debug_out_file: /path/to/debug/file.txt
+
+# Selected values ​​experimentally so as to ensure relatively fast convergence (values ​​should be about 10x higher than the sensor variacne values)
```
macstepien:
```suggestion
# Selected values ​​experimentally so as to ensure relatively fast convergence (values ​​should be about 10x higher than the sensor variance values)
```

panther_bringup/config/imu_config.yaml
```diff
@@ -1,19 +1,26 @@
...
+  gain: 0.1
+  zeta: 0.0
+  mag_bias_x: 0.0
+  mag_bias_y: 0.0
+  mag_bias_z: 0.0
+  orientation_stddev: 1e-2
   publish_tf: false
   use_mag: false
-  use_magnetic_field_msg: false
+  use_magnetic_field_msg: true # If set to true, subscribe to the /imu/mag topic as a sensor_msgs/MagneticField; if set to false (deprecated), use geometry_msgs/Vector3Stamped.
```
macstepien:
```suggestion
  use_magnetic_field_msg: true # If set to true, subscribe to the /imu/mag topic as a sensor_msgs/MagneticField; if set to false (deprecated), use geometry_msgs/Vector3Stamped. Only changes type of the message, whether mag data is used is controlled by the use_mag parameter
```

panther_bringup/config/imu_config.yaml
```diff
@@ -1,19 +1,26 @@
...
   publish_rate: 100.0 # optional param publish_rate, defaults to 0
   # serial: 123456  # optional param serial, default is -1
   # hub_port: 0 # optional param hub_port, used if connected to a VINT hub
   # frame_id: imu_link  # optional param frame_id, default is "imu_link"
-  # linear_acceleration_stdev: 0.002745862  # optional param linear_acceleration_stdev, default is 280ug
-  # angular_velocity_stdev: 0.000349  # optional param angular_velocity_stdev, default is 0.095 deg/s
-  # magnetic_field_stdev: 1.1e-7  # optional param magnetic_field_stdev, default is 1.1 milligauss
+  linear_acceleration_stdev: 3.5e-3   # Currently used IMU specification 10 * (2mg ± 1.5mg) (Accelerometer Drift Max ± Accelerometer Noise for 100 Hz) (https://www.phidgets.com/?tier=3&catid=10&pcid=8&prodid=1205)
+  angular_velocity_stdev: 1.5e-2      # Currently used IMU specification 0.1°/s ± 0.05°/s (Gyroscope Drift Max ± Gyroscope Noise for 100 Hz) (https://www.phidgets.com/?tier=3&catid=10&pcid=8&prodid=1205)
+  # magnetic_field_stdev: 10.0          # Currently used IMU specification ± 10mG (± Magnetometer Noise) (https://www.phidgets.com/?tier=3&catid=10&pcid=8&prodid=1205)
```
macstepien:
```suggestion
  # stdev values from imu specification (https://www.phidgets.com/?tier=3&catid=10&pcid=8&prodid=1205)
  linear_acceleration_stdev: 3.5e-3   # Currently used IMU specification 10 * (2mg ± 1.5mg) (Accelerometer Drift Max ± Accelerometer Noise for 100 Hz)
  angular_velocity_stdev: 1.5e-2      # Currently used IMU specification 0.1°/s ± 0.05°/s (Gyroscope Drift Max ± Gyroscope Noise for 100 Hz)
  # magnetic_field_stdev: 10.0          # Currently used IMU specification ± 10mG (± Magnetometer Noise)
```

panther_description/config/WH01.yaml
```diff
@@ -6,3 +6,7 @@ inertia_y_offset: 0.0
 mesh_package: panther_description
 folder_path: meshes/WH01
 kinematics: differential
+
+velocity_x_stderr: 3.2e-3
```
macstepien, line 11: Add comment, that this values were measured.

panther_description/config/WH02.yaml
```diff
@@ -6,3 +6,7 @@ inertia_y_offset: 0.0
 mesh_package: panther_description
 folder_path: meshes/WH02
 kinematics: mecanum
+
+velocity_x_stderr: 3.2e-3     # Selected intuitively, the need for thorough testing
+velocity_y_stderr: 1.4e-3     # Selected intuitively, the need for thorough testing
+velocity_yaw_stderr: 8.5e-3   # Selected intuitively, the need for thorough testing
```
macstepien: These are mecanum wheels, they shouldn't have the same values. It should be probably copied to WH04 instead. Also change to comment to something like `selected based on WH01 results`

macstepien (changes requested):

panther_driver/src/driver_node.py
rafal-gorecki: Looking on pose covariance I decide to live it in that way. Now is more similar. It could be also like this:
```
cov_value = [self._v_x_var, self._v_y_var, 1e6, 1e6, 1e6, self._v_yaw_var]
twist_cov = [cov_value[i//7] if (i % 7) == 0 else 0.0 for i in range(36)]
```
bur this is not intuitive.

panther_bringup/config/ekf_config.yaml
rafal-gorecki: I test it also on 20 and looks fine

panther_driver/src/driver_node.py
rafal-gorecki: I decide to assign to full 6x6 matrix

panther_description/config/WH01.yaml
```diff
@@ -6,3 +6,7 @@ inertia_y_offset: 0.0
 mesh_package: panther_description
 folder_path: meshes/WH01
 kinematics: differential
+
+velocity_x_stderr: 3.2e-3    # Value measured (using Husarion research toolkit)
+velocity_y_stderr: 1.4e-3    # Value measured (using Husarion research toolkit)
+velocity_yaw_stderr: 8.5e-3  # Value measured (using Husarion research toolkit)
```
macstepien:
```suggestion

# Value measured (using Husarion research toolkit)
velocity_x_stderr: 3.2e-3
velocity_y_stderr: 1.4e-3
velocity_yaw_stderr: 8.5e-3
```

panther_description/config/WH02.yaml
```diff
@@ -6,3 +6,7 @@ inertia_y_offset: 0.0
 mesh_package: panther_description
 folder_path: meshes/WH02
 kinematics: mecanum
+
+velocity_x_stderr: 3.2e-3     # Selected intuitively based on WH01 results
+velocity_y_stderr: 3.2e-3     # Selected intuitively based on WH01 results
+velocity_yaw_stderr: 8.5e-3   # Selected intuitively based on WH01 results
```
macstepien:
```suggestion

# Selected intuitively based on WH01 results
velocity_x_stderr: 3.2e-3
velocity_y_stderr: 3.2e-3
velocity_yaw_stderr: 8.5e-3
```

panther_description/config/WH04.yaml
```diff
@@ -6,3 +6,7 @@ inertia_y_offset: 0.0
 mesh_package: panther_description
 folder_path: meshes/WH04
 kinematics: differential
+
+velocity_x_stderr: 1.6e-3     # Selected intuitively based on WH01 results
+velocity_y_stderr: 7.3e-4     # Selected intuitively based on WH01 results
+velocity_yaw_stderr: 4.3e-3   # Selected intuitively based on WH01 results
```
macstepien:
```suggestion

# Selected intuitively based on WH01 results
velocity_x_stderr: 1.6e-3
velocity_y_stderr: 7.3e-4
velocity_yaw_stderr: 4.3e-3
```

macstepien (changes requested):

macstepien (approved):

PR #128, opened 2023-06-20
ROS 2 panther battery package

KmakD:


panther_battery/include/panther_battery/roboteq_republisher_node.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <sensor_msgs/msg/battery_state.hpp>
+
+#include <panther_msgs/msg/driver_state.hpp>
+
+#include <panther_utils/moving_average.hpp>
+
+#define V_BAT_FATAL_MIN 27.0
+#define V_BAT_FATAL_MAX 43.0
+#define V_BAT_FULL 41.4
+#define V_BAT_MIN 32.0
```
Kotochleb: Those could be private const variables of the class. Also, preprocessor macros should be replaced with const and constexpressions in C++

panther_battery/src/roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+  battery_voltage_ma_->Roll((msg.front.voltage + msg.rear.voltage) / 2.0);
+  battery_current_ma_->Roll(msg.front.current + msg.rear.current);
+}
+
+void RoboteqRepublisherNode::BatteryPubTimerCB()
+{
+  auto battery_msg = BatteryStateMsg();
+  battery_msg.header.stamp = this->get_clock()->now();
+  battery_msg.capacity = 20.0;
+  battery_msg.design_capacity = 20.0;
```
Kotochleb: Those could be constants in the class

panther_battery/src/roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+  battery_pub_timer_ = this->create_wall_timer(
+    std::chrono::milliseconds(100), std::bind(&RoboteqRepublisherNode::BatteryPubTimerCB, this));
+
+  RCLCPP_INFO(this->get_logger(), "Node started");
+}
+
+void RoboteqRepublisherNode::MotorControllersStateSubCB(const DriverStateMsg & msg)
+{
+  if (msg.front.fault_flag.can_net_err || msg.rear.fault_flag.can_net_err) {
+    return;
```
Kotochleb, line 51: Can't the moving average be reset also in here? This will reduce the time vale of voltage goes to naan in case of error on CAN bus

panther_battery/src/roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  auto battery_voltage = battery_voltage_ma_->GetAverage();
+  auto battery_current = battery_current_ma_->GetAverage();
+
+  if (
+    std::isnan(battery_voltage) || std::isnan(battery_current) ||
+    (this->get_clock()->now() - last_battery_info_time_) >
+      rclcpp::Duration::from_seconds(battery_timeout_)) {
+    battery_voltage_ma_->Reset();
+    battery_current_ma_->Reset();
```
Kotochleb, line 82: This will reset the moving average every time even though it was already reset. Another `if` checking nan value would prevent that

panther_battery/src/roboteq_republisher_node.cpp
Kotochleb: According to [documentation](http://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/BatteryState.html) `cell_voltage` could be added as an array of 10 nans. Just for the sake of absolute correctness

panther_battery/src/main.cpp
Kotochleb, line 1: Since we already know this will not be the only node in this package maybe this should be renamed to version matching two main files

panther_utils/include/panther_utils/test/test_utils.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <rclcpp/rclcpp.hpp>
+
+using namespace std::chrono_literals;
+
+namespace panther_utils::test_utils
+{
+
+template <typename NodeT, typename MsgT>
+bool WaitForMsg(
+  const std::shared_ptr<NodeT> node, std::shared_ptr<MsgT> & msg,
```
Kotochleb:
```suggestion
  const std::shared_ptr<NodeT> & node, std::shared_ptr<MsgT> & msg,
```
If it is not, reference copy of the node is created on every invocation of this function

panther_utils/include/panther_utils/test/test_utils.hpp
```diff
@@ -0,0 +1,34 @@
...
+{
+  msg = nullptr;
+  rclcpp::Time start_time = node->now();
+
+  while (rclcpp::ok() && node->now() - start_time <= rclcpp::Duration(timeout)) {
+    if (msg) {
+      return true;
+    }
+    rclcpp::spin_some(node->get_node_base_interface());
+    std::this_thread::sleep_for(10ms);
```
Kotochleb:
```suggestion

namespace panther_utils::test_utils
{

template <typename NodeT, typename MsgT>
bool WaitForMsg(
  const std::shared_ptr<NodeT> node, std::shared_ptr<MsgT> & msg,
  const std::chrono::nanoseconds & timeout)
{
  msg = nullptr;
  rclcpp::Time start_time = node->now();

  while (rclcpp::ok() && node->now() - start_time <= rclcpp::Duration(timeout)) {
    if (msg) {
      return true;
    }
    rclcpp::spin_some(node->get_node_base_interface());
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
```

panther_battery/src/main.cpp
```diff
@@ -0,0 +1,14 @@
+#include <panther_battery/roboteq_republisher_node.hpp>
+
+using namespace panther_battery;
```
Kotochleb: I would skip this using namespace and either put everything in the namespace or just used `panther_battery::RoboteqRepublisherNode`

panther_battery/src/main.cpp
```diff
@@ -0,0 +1,14 @@
...
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+
+  auto roboteq_republisher_node = std::make_shared<RoboteqRepublisherNode>();
+
+  rclcpp::spin(roboteq_republisher_node);
+  rclcpp::shutdown();
+  return 0;
+}
```
Kotochleb: There could be try catch just like it is in [joy2twist](https://github.com/husarion/joy2twist/blob/ros2/joy2twist/src/main.cpp)

panther_battery/src/roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+#include <panther_utils/moving_average.hpp>
+
+namespace panther_battery
+{
+using std::placeholders::_1;
+
+RoboteqRepublisherNode::RoboteqRepublisherNode() : Node("roboteq_republisher_node")
+{
+  battery_timeout_ = 1.0;
```
Kotochleb: If specified this way, `battery_timeout_` should be either `const` and specified in `.hpp` file or an ROS parameter, and I would opt for this option TBH

panther_battery/src/roboteq_republisher_node.cpp
Kotochleb: Optionally, [documentation](http://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/BatteryState.html) specifies `location` as a field and this could be set to `user_compartment`

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include <rclcpp/rclcpp.hpp>
+
+#include <sensor_msgs/msg/battery_state.hpp>
+
+#include <panther_msgs/msg/driver_state.hpp>
+
+#include <panther_battery/roboteq_republisher_node.hpp>
+#include <panther_utils/test/test_utils.hpp>
+
+using namespace std::chrono_literals;
```
Kotochleb: As discussed, everything should be as `std::chrono::milliseconds(my_time)`

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
+using namespace panther_utils::test_utils;
```
Kotochleb:
```suggestion
```
All functions comming from panther_utils should be placed with namespace

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
+TestRoboteqRepublisherNode::~TestRoboteqRepublisherNode()
+{
+  roboteq_republisher_node_.reset();
+  driver_state_pub_.reset();
+  battery_state_sub_.reset();
+}
+
+void TestRoboteqRepublisherNode::CheckBatteryStateMsg(
+  const float & expected_voltage, const float & expected_current,
+  const uint8_t & power_supply_status, const uint8_t & power_supply_health)
```
Kotochleb: I think `expected_percentage` should be moved as a parameter of this function and be manually calculated on each test. This way we will test percentage calculation. The current approach is using exactly the same function to generate both normal and test data, hence it is not testing the function itself. It could be either separate test checking if the output of this function is indeed correct or the percentage should be calculated manually in all cases

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
+  EXPECT_FLOAT_EQ(expected_current, battery_state_->current);
+  EXPECT_FLOAT_EQ(expected_percentage, battery_state_->percentage);
+  EXPECT_FLOAT_EQ(expected_percentage * 20.0, battery_state_->charge);
+
+  if (power_supply_status == BatteryStateMsg::POWER_SUPPLY_STATUS_DISCHARGING) {
+    EXPECT_TRUE(battery_state_->present);
+  }
+  EXPECT_EQ(power_supply_status, battery_state_->power_supply_status);
+  EXPECT_EQ(power_supply_health, battery_state_->power_supply_health);
+}
```
Kotochleb: If we decide to add cells and localization of the battery we should test it

panther_battery/src/roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+  auto battery_current = battery_current_ma_->GetAverage();
+
+  if (
+    std::isnan(battery_voltage) || std::isnan(battery_current) ||
+    (this->get_clock()->now() - last_battery_info_time_) >
+      rclcpp::Duration::from_seconds(battery_timeout_)) {
+    battery_voltage_ma_->Reset();
+    battery_current_ma_->Reset();
+    battery_msg.power_supply_status = BatteryStateMsg::POWER_SUPPLY_STATUS_UNKNOWN;
+    battery_msg.power_supply_health = BatteryStateMsg::POWER_SUPPLY_HEALTH_UNKNOWN;
```
Kotochleb: If we do not know the state of the battery voltage, current and charge should not be nan instead of default 0?

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
+  driver_state_msg.front.voltage = 37.0;
+  driver_state_msg.rear.voltage = 39.0;
+  driver_state_msg.front.current = 0.1;
+  driver_state_msg.rear.current = 0.1;
+
+  driver_state_pub_->publish(driver_state_msg);
+  ASSERT_TRUE(
+    WaitForMsg(roboteq_republisher_node_, battery_state_, std::chrono::milliseconds(1000)));
+
+  auto expected_voltage = (driver_state_msg.front.voltage + driver_state_msg.rear.voltage) / 2;
```
Kotochleb:
```suggestion
  auto expected_voltage = (driver_state_msg.front.voltage + driver_state_msg.rear.voltage) / 2.0;
```

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
+  driver_state_msg.front.voltage = V_BAT_FATAL_MIN - 1.0;
+  driver_state_msg.rear.voltage = V_BAT_FATAL_MIN - 1.0;
+  driver_state_msg.front.current = 0.1;
+  driver_state_msg.rear.current = 0.1;
+
+  driver_state_pub_->publish(driver_state_msg);
+  ASSERT_TRUE(
+    WaitForMsg(roboteq_republisher_node_, battery_state_, std::chrono::milliseconds(1000)));
+
+  auto expected_voltage = (driver_state_msg.front.voltage + driver_state_msg.rear.voltage) / 2;
```
Kotochleb:
```suggestion
  auto expected_voltage = (driver_state_msg.front.voltage + driver_state_msg.rear.voltage) / 2.0;
```

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
+  driver_state_msg.front.voltage = V_BAT_FATAL_MAX + 1.0;
+  driver_state_msg.rear.voltage = V_BAT_FATAL_MAX + 1.0;
+  driver_state_msg.front.current = 0.1;
+  driver_state_msg.rear.current = 0.1;
+
+  driver_state_pub_->publish(driver_state_msg);
+  ASSERT_TRUE(
+    WaitForMsg(roboteq_republisher_node_, battery_state_, std::chrono::milliseconds(1000)));
+
+  auto expected_voltage = (driver_state_msg.front.voltage + driver_state_msg.rear.voltage) / 2;
```
Kotochleb:
```suggestion
  auto expected_voltage = (driver_state_msg.front.voltage + driver_state_msg.rear.voltage) / 2.0;
```

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
+  driver_state_msg.rear.voltage = 37.0;
+  driver_state_msg.front.current = 0.1;
+  driver_state_msg.rear.current = 0.1;
+
+  driver_state_pub_->publish(driver_state_msg);
+  ASSERT_TRUE(
+    WaitForMsg(roboteq_republisher_node_, battery_state_, std::chrono::milliseconds(1000)));
+
+  // check if message was sent correctly
+  auto expected_voltage = (driver_state_msg.front.voltage + driver_state_msg.rear.voltage) / 2;
```
Kotochleb:
```suggestion
  auto expected_voltage = (driver_state_msg.front.voltage + driver_state_msg.rear.voltage) / 2.0;
```

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
```
Kotochleb:
```suggestion
  auto expected_voltage = (driver_state_msg.front.voltage + driver_state_msg.rear.voltage) / 2.0;
```

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,220 @@
...
+  driver_state_msg.rear.current = 0.2;
+  driver_state_pub_->publish(driver_state_msg);
+  ASSERT_TRUE(
+    WaitForMsg(roboteq_republisher_node_, battery_state_, std::chrono::milliseconds(1000)));
+
+  // voltage and current values should not be updated
+  CheckBatteryStateMsg(
+    expected_voltage, expected_current, BatteryStateMsg::POWER_SUPPLY_STATUS_DISCHARGING,
+    BatteryStateMsg::POWER_SUPPLY_HEALTH_GOOD);
+}
```
Kotochleb: In case of can net error timeout is not checked

panther_battery/test/test_roboteq_republisher_node.cpp
Kotochleb, line 1: There are  missing test cases for:
- negative current
- smoothing (aka moving average) is not tested
- exact values of edge cases, like what happens when the battery is really close to 42 V 
- there might be one more test for the correct value

Tests could assume no prior knowledge about limits within class. This way, if values inside class are changed, tests checking hardware safety will fail

Kotochleb (changes requested):

panther_battery/src/roboteq_republisher_node.cpp
KmakD: Consider case when get only 1 mgs with can_net_error and the next one is correct. In that case we don't want to reset average, rather keep its value and update it when correct message is received. Average will be reset if can_net_error persist longer that timeout.

panther_battery/test/test_roboteq_republisher_node.cpp
KmakD: Timeout is checked in a different test, there is no purpose for repeating tests.

panther_battery/include/panther_battery/roboteq_republisher_node.hpp
```diff
@@ -0,0 +1,48 @@
...
+public:
+  RoboteqRepublisherNode();
+
+private:
+  static constexpr float bat_capacity_ = 20.0f;
+  static constexpr float bat_designed_capacity_ = 20.0f;
+  static constexpr float V_bat_fatal_min_ = 27.0f;
+  static constexpr float V_bat_fatal_max_ = 43.0f;
+  static constexpr float V_bat_full_ = 41.4f;
+  static constexpr float V_bat_min_ = 32.0f;
```
Kotochleb:
```suggestion
  static constexpr float v_bat_fatal_min_ = 27.0f;
  static constexpr float v_bat_fatal_max_ = 43.0f;
  static constexpr float v_bat_full_ = 41.4f;
  static constexpr float v_bat_min_ = 32.0f;
```
IMO this will look better

panther_battery/include/panther_battery/roboteq_republisher_node.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+  float battery_timeout_;
```
Kotochleb, line 33: This could be already `rclcpp::Duration`

panther_battery/include/panther_battery/roboteq_republisher_node.hpp
Kotochleb:
```suggestion
  static constexpr float V_bat_min_ = 32.0f;
  static constexpr std::string_view location_ = "user_compartment";
```
This might be a way to add const version of location name

panther_battery/src/roboteq_republisher_node.cpp
Kotochleb: Ok. Makse sene

panther_battery/src/main.cpp
Kotochleb: As discussed we are going to use `using`...

panther_battery/src/roboteq_republisher_node.cpp
Kotochleb: You forgot about this one

panther_battery/src/roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+RoboteqRepublisherNode::RoboteqRepublisherNode() : Node("roboteq_republisher_node")
+{
+  declare_parameter("battery_timeout", 1.0f);
+  declare_parameter("batery_voltage_window_len", 10);
+  declare_parameter("batery_current_window_len", 10);
+
+  get_parameter("battery_timeout", battery_timeout_);
+  const auto batery_voltage_window_len = get_parameter("batery_voltage_window_len").as_int();
+  const auto batery_current_window_len = get_parameter("batery_current_window_len").as_int();
```
Kotochleb:
```suggestion
  this->declare_parameter("battery_timeout", 1.0f);
  this->declare_parameter("batery_voltage_window_len", 10);
  this->declare_parameter("batery_current_window_len", 10);

  this->get_parameter("battery_timeout", battery_timeout_);
  const auto batery_voltage_window_len = this->get_parameter("batery_voltage_window_len").as_int();
  const auto batery_current_window_len = this->get_parameter("batery_current_window_len").as_int();
```

panther_battery/src/roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,108 @@
...
+    battery_current_ma_->Reset();
+    battery_msg.power_supply_status = BatteryStateMsg::POWER_SUPPLY_STATUS_UNKNOWN;
+    battery_msg.power_supply_health = BatteryStateMsg::POWER_SUPPLY_HEALTH_UNKNOWN;
+  } else {
+    battery_msg.voltage = battery_voltage;
+    battery_msg.current = battery_current;
+    battery_msg.percentage =
+      std::clamp((battery_voltage - V_bat_min_) / (V_bat_full_ - V_bat_min_), 0.0, 1.0);
+    battery_msg.charge = battery_msg.percentage * battery_msg.design_capacity;
+    battery_msg.present = true;
```
Kotochleb: Haven't we discussed that no matter what battery is always present, but it's state is unknown in case of error? And in case of unknown state values does not go to zeros but rather NaN

panther_battery/include/panther_battery/roboteq_republisher_node.hpp
Kotochleb: Once you use `f`  after float and once you do not. I think if should be either always or never

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,344 @@
...
+  driver_state_pub_->publish(driver_state_msg);
+  ASSERT_TRUE(panther_utils::test_utils::WaitForMsg(
+    roboteq_republisher_node_, battery_state_, std::chrono::milliseconds(1000)));
+
+  expected_voltage = 36.0;
+  expected_current = -0.1;
+  expected_percentage = 0.425531915;
+  CheckBatteryStateMsg(
+    expected_voltage, expected_current, expected_percentage,
+    BatteryStateMsg::POWER_SUPPLY_STATUS_DISCHARGING, BatteryStateMsg::POWER_SUPPLY_HEALTH_GOOD);
```
Kotochleb, line 151: I am not sure if this shouldn't be a separate test case for "manual" battery percentage. Also the negative current might be a separate test since it is independent to percentage, hence it should be detected by separate test case

panther_battery/test/test_roboteq_republisher_node.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+  // voltage and current values should not be updated
+  CheckBatteryStateMsg(
+    expected_voltage, expected_current, expected_percentage,
+    BatteryStateMsg::POWER_SUPPLY_STATUS_DISCHARGING, BatteryStateMsg::POWER_SUPPLY_HEALTH_GOOD);
+
+  // publish driver state until timeout is reached
+  rclcpp::Time start_time = roboteq_republisher_node_->now();
+  while (rclcpp::ok() &&
+         roboteq_republisher_node_->now() - start_time < std::chrono::milliseconds(2000)) {
```
Kotochleb:
```suggestion
         roboteq_republisher_node_->now() - start_time < std::chrono::milliseconds(2500)) {
```
Can't is be a bit more? 2s will in worst case publish at most 1 time before reaching timeout

Kotochleb (changes requested):
General test if moving average on simple 2 or 3 values is still missing

panther_battery/include/panther_battery/roboteq_republisher_node.hpp
KmakD: rclcpp::Duration doesn't have default constructor, it's easier using float

panther_battery/test/test_roboteq_republisher_node.cpp
KmakD: As discussed, we leave it in 1 test case

panther_battery/include/panther_battery/roboteq_republisher_node.hpp
Kotochleb: Ok

panther_battery/test/test_roboteq_republisher_node.cpp
Kotochleb: Ok

Kotochleb (approved):

PR #135, opened 2023-06-29
ROS 2 adc node

KmakD:


panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,50 @@
...
+  float GetADCMeasurement(const std::string & file, const float & offset, const float & LSB);
+
+private:
+  const std::string device_path_;
+
+  int ReadFile(std::string file_path);
+};
+
+inline float ADCDataReader::GetADCMeasurement(
+  const std::string & file, const float & offset, const float & LSB)
```
Kotochleb:
```suggestion
  const std::string & file, const float offset, const float LSB)
```
I don't think passing basic types as reference is a good idea. For objects it makes sense, but for float I do not second this

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace panther_battery
+{
+
+class ADCDataReader
+{
+public:
+  ADCDataReader(const std::string & device_path) : device_path_(device_path) {}
+
+  float GetADCMeasurement(const std::string & file, const float & offset, const float & LSB);
```
Kotochleb: Instead passing the file it could just take channel number and then combine strings itself. Same goes for LSB and offset. Offset and LSB should be computed in more explicit way

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <stdexcept>
+#include <string>
+
+namespace panther_battery
+{
+
+class ADCDataReader
+{
+public:
+  ADCDataReader(const std::string & device_path) : device_path_(device_path) {}
```
Kotochleb: Constructor should check if path exists and raise exception if it doesn't

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+  float GetADCMeasurement(const std::string & file, const float & offset, const float & LSB);
+
+private:
+  const std::string device_path_;
+
+  int ReadFile(std::string file_path);
```
Kotochleb:
```suggestion
  int ReadChannel(std::string file_path);
```

panther_battery/include/panther_battery/adc_data_reader.hpp
Kotochleb: Those values hard-coded as `LSB` are in face transformation from ADC reading directly to the real voltage value. `LSB` is defined as the smallest value ADC can read and should be specified affording to this table:
image

According to [documentation](https://www.kernel.org/doc/html/v5.3/hwmon/ads1015.html) `FSR`, from which `LSB` can be computed, can be obtained from gain parameter. Gain of a given ADC can be on the other hand obtained from kernel parameter from file `/sys/bus/iio/devices/iio:deviceXXX/in_voltageXXX_gainXXX`

`LSB` used right now was computed with this formula
```
LFS = 2*gain*(Rg+Rd)/(Rd*rawfs)
```
But I think moving the `Rg` and `Rd`  to the battery class since those are resistor values will be more suitable. This ADC class will give raw voltage it is measuring. The class itself will be more general

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,50 @@
...
+};
+
+inline float ADCDataReader::GetADCMeasurement(
```
Kotochleb: Is there any particular reason why this is `inline`?

panther_battery/include/panther_battery/adc_data_reader.hpp
Kotochleb: Device name could be resolved from a full path with all this `/sys/bus/...` or just as `device0`. This way parameters would not have to hold the whole device path, but only name of the device. At the same time testing with full path would be possible

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+class ADCDataReader
+{
+public:
+  ADCDataReader(const std::string & device_path) : device_path_(device_path) {}
+
+  float GetADCMeasurement(const std::string & file, const float & offset, const float & LSB);
+
+private:
+  const std::string device_path_;
```
Kotochleb:
```suggestion
  const std::filesystem::path device_path_;
```

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,50 @@
+#ifndef PANTHER_BATTERY_ADC_DATA_READER_HPP_
+#define PANTHER_BATTERY_ADC_DATA_READER_HPP_
+
+#include <math.h>
```
Kotochleb: Unused import, also in future use `cmath`

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include <filesystem>
+#include <fstream>
+#include <map>
```
Kotochleb: Unused include

panther_battery/include/panther_battery/battery_publisher.hpp
Kotochleb, line 1: This whole class could actually be shared between this node and roboteq republisher node

panther_battery/include/panther_battery/battery_publisher.hpp
```diff
@@ -0,0 +1,182 @@
...
+#include <sensor_msgs/msg/battery_state.hpp>
+
+#include <panther_utils/moving_average.hpp>
+
+namespace panther_battery
+{
+
+using BatteryStateMsg = sensor_msgs::msg::BatteryState;
+
+class BatteryPublisher
```
Kotochleb, line 18: It will be better suited to use inherence from `rclcpp::PublisherBase::SharedPtr` instead passing the publisher as an argument

panther_battery/include/panther_battery/battery_publisher.hpp
```diff
@@ -0,0 +1,182 @@
...
+{
+public:
+  BatteryPublisher(
+    rclcpp::Publisher<BatteryStateMsg>::SharedPtr battery_pub, const float & high_bat_temp,
+    const float & battery_charging_current_tresh, const int & batery_voltage_window_len,
+    const int & batery_temp_window_len, const int & batery_current_window_len,
+    const int & batery_charge_window_len, const float & battery_designed_capacity);
```
Kotochleb: Passing primitive types as reference is unnecessary and it would be better to simplu use it as an argument

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+  auto header_stamp = this->get_clock()->now();
+
+  if (
+    (this->get_clock()->now() - last_battery_info_time_) >
+    rclcpp::Duration::from_seconds(battery_timeout_)) {
+    battery_pub_->PublishUnknown(header_stamp);
+    if (battery_count_ == 2) {
+      battery_1_pub_->PublishUnknown(header_stamp);
+      battery_2_pub_->PublishUnknown(header_stamp);
+    }
```
Kotochleb: Timeout made sense if battery relied on another topic. In this scenario in case of fault on reading we can right away publish unknown state, because we know it won't make any difference. Also the unknown state in case of two batteries should affect only the battery that is in fact broken plus the "merged" one

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+    V_bat_2_ = adc1_reader_->GetADCMeasurement("in_voltage3_raw", 0.0, 0.02504255);
+    I_bat_1_ = adc1_reader_->GetADCMeasurement("in_voltage2_raw", 625.0, 0.04);
+    I_bat_2_ = adc1_reader_->GetADCMeasurement("in_voltage1_raw", 625.0, 0.04);
+    V_temp_bat_1_ = adc0_reader_->GetADCMeasurement("in_voltage1_raw", 0.0, 0.002);
+    V_temp_bat_2_ = adc0_reader_->GetADCMeasurement("in_voltage0_raw", 0.0, 0.002);
+    I_charge_bat_1_ = adc0_reader_->GetADCMeasurement("in_voltage3_raw", 0.0, 0.005);
+    I_charge_bat_2_ = adc0_reader_->GetADCMeasurement("in_voltage2_raw", 0.0, 0.005);
+    last_battery_info_time_ = this->get_clock()->now();
+  } catch (const std::runtime_error & e) {
+    RCLCPP_ERROR(this->get_logger(), "Error reading ADC data: %s. ", e.what());
```
Kotochleb: This should be more general exception I think. There could happen iostream exception. In case of no file on the other hand the node should stop imminently since this means broken hardware

Kotochleb:
```suggestion
    RCLCPP_ERROR(this->get_logger(), "Error reading battery data: %s. ", e.what());
```

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+using std::placeholders::_1;
+
+ADCNode::ADCNode(const std::string & node_name, const rclcpp::NodeOptions & options)
+: Node(node_name, options)
+{
+  this->declare_parameter<std::string>("adc0_device", "/sys/bus/iio/devices/iio:device0");
+  this->declare_parameter<std::string>("adc1_device", "/sys/bus/iio/devices/iio:device1");
+  this->declare_parameter<float>("battery_timeout", 1.0);
+  this->declare_parameter<float>("high_bat_temp", 55.0);
```
Kotochleb:
```suggestion
  this->declare_parameter<double>("battery_timeout", 1.0);
  this->declare_parameter<double>("high_bat_temp", 55.0);
```
It is first specified as `float` and later read as `double`. This should be more consistent

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+  RCLCPP_INFO(this->get_logger(), "Battery count: %d", battery_count_);
+  RCLCPP_INFO(this->get_logger(), "Node started");
+}
+
+int ADCNode::CheckBatteryCount()
+{
+  auto trials_num = 10;
+  auto V_temp_sum = 0.0;
+  auto V_temp_bat_2 = 0.0;
```
Kotochleb: Compiler will assume `double` for those values. Either replace `auto` with regular values or add `f` at the end of floats

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    V_temp_bat_2 = V_temp_sum / trials_num;
+
+  } catch (const std::runtime_error & e) {
+    RCLCPP_ERROR(
+      this->get_logger(),
+      "Error reading ADC data: %s. "
+      "\nThe number of batteries cannot be determined. The single battery was adopted",
+      e.what());
```
Kotochleb: This error will only appear if adc0 is broken, meaning the hardware itself is broken, since adc0 is used by both batteries. In the end the node will not work properly no matter of battery count

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+    std::chrono::milliseconds(100), std::bind(&ADCNode::BatteryPubTimerCB, this));
+
+  RCLCPP_INFO(this->get_logger(), "Battery count: %d", battery_count_);
+  RCLCPP_INFO(this->get_logger(), "Node started");
+}
+
+int ADCNode::CheckBatteryCount()
+{
+  auto trials_num = 10;
+  auto V_temp_sum = 0.0;
```
Kotochleb: This value could be declared right before the `for` loop

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+  battery_pub_timer_ = this->create_wall_timer(
+    std::chrono::milliseconds(100), std::bind(&ADCNode::BatteryPubTimerCB, this));
+
+  RCLCPP_INFO(this->get_logger(), "Battery count: %d", battery_count_);
+  RCLCPP_INFO(this->get_logger(), "Node started");
+}
+
+int ADCNode::CheckBatteryCount()
+{
+  auto trials_num = 10;
```
Kotochleb:
```suggestion
  const auto get_temp_attempts = 10;
```

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+  } catch (const std::runtime_error & e) {
+    RCLCPP_ERROR(
+      this->get_logger(),
+      "Error reading ADC data: %s. "
+      "\nThe number of batteries cannot be determined. The single battery was adopted",
+      e.what());
+    return 1;
+  }
+
+  return V_temp_bat_2 > bat02_detect_thresh_ ? 1 : 2;
```
Kotochleb:
```suggestion
   const float V_temp_bat_2 = V_temp_sum / trials_num;
  return V_temp_bat_2 > bat02_detect_thresh_ ? 1 : 2;
```
This will make everything more compact

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+  const auto battery_charge_window_len = this->get_parameter("batery_charge_window_len").as_int();
+
+  adc0_reader_ = std::make_unique<ADCDataReader>(adc0_device_);
+  adc1_reader_ = std::make_unique<ADCDataReader>(adc1_device_);
+  last_battery_info_time_ = rclcpp::Time(int64_t(0), RCL_ROS_TIME);
+
+  battery_count_ = CheckBatteryCount();
+
+  // create battery publisher based on battery count
+  if (battery_count_ == 2) {
```
Kotochleb:
```suggestion
  // create battery publisher based on battery count
  battery_count_ = CheckBatteryCount();
  if (battery_count_ == 2) {
```

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+float ADCNode::VoltageTempToDeg(const float & V_temp)
+{
+  if (V_temp == 0 || V_temp >= u_supply_) {
+    return std::numeric_limits<float>::quiet_NaN();
+  }
+
+  auto R_therm = (V_temp * R1_) / (u_supply_ - V_temp);
+  return (A_ * B_ / (A_ * log(R_therm / R0_) + B_)) - 273.15;
+}
```
Kotochleb: There could be a battery class encapsualting such computations

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+float ADCNode::VoltageTempToDeg(const float & V_temp)
+{
+  if (V_temp == 0 || V_temp >= u_supply_) {
+    return std::numeric_limits<float>::quiet_NaN();
+  }
+
+  auto R_therm = (V_temp * R1_) / (u_supply_ - V_temp);
+  return (A_ * B_ / (A_ * log(R_therm / R0_) + B_)) - 273.15;
```
Kotochleb: `273.15` should be const value in class

panther_battery/include/panther_battery/battery_publisher.hpp
```diff
@@ -0,0 +1,182 @@
...
+    battery_msg.power_supply_health = BatteryStateMsg::POWER_SUPPLY_HEALTH_GOOD;
+    error_msg_ = "";
+  }
+
+  battery_pub_->publish(battery_msg);
+}
+
+inline bool BatteryPublisher::HasErrorMsg() const
+{
+  if (error_msg_ == "") {
```
Kotochleb:
```suggestion
  if (error_msg_.empty()) {
```

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+        charger_connected_);
+    } else {
+      battery_pub_->Publish(
+        header_stamp, V_bat_1_, temp_bat_1, -(I_bat_1_ + I_bat_2_) + I_charge_bat_1_,
+        I_charge_bat_1_, charger_connected_);
+    }
+  }
+
+  // print error messages
+  if (battery_pub_->HasErrorMsg()) {
```
Kotochleb:
```suggestion
  if (battery_count_ == 1 && battery_pub_->HasErrorMsg()) {
```
Since all possible fault flags will come from the average of the two batteries the errors will be later logged separately on each battery. This will just repeat with shlight delay the error for the "abstract, mean battery"

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+  try {
+    V_bat_1_ = adc1_reader_->GetADCMeasurement("in_voltage0_raw", 0.0, 0.02504255);
+    V_bat_2_ = adc1_reader_->GetADCMeasurement("in_voltage3_raw", 0.0, 0.02504255);
+    I_bat_1_ = adc1_reader_->GetADCMeasurement("in_voltage2_raw", 625.0, 0.04);
+    I_bat_2_ = adc1_reader_->GetADCMeasurement("in_voltage1_raw", 625.0, 0.04);
+    V_temp_bat_1_ = adc0_reader_->GetADCMeasurement("in_voltage1_raw", 0.0, 0.002);
+    V_temp_bat_2_ = adc0_reader_->GetADCMeasurement("in_voltage0_raw", 0.0, 0.002);
+    I_charge_bat_1_ = adc0_reader_->GetADCMeasurement("in_voltage3_raw", 0.0, 0.005);
+    I_charge_bat_2_ = adc0_reader_->GetADCMeasurement("in_voltage2_raw", 0.0, 0.005);
```
Kotochleb: If there was "battery object this would be replaced by more abstract getters

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,191 @@
...
+      "Error reading ADC data: %s. "
+      "\nThe number of batteries cannot be determined. The single battery was adopted",
+      e.what());
+    return 1;
+  }
+
+  return V_temp_bat_2 > bat02_detect_thresh_ ? 1 : 2;
+}
+
+void ADCNode::IOStateCB(const IOStateMsg & msg) { charger_connected_ = msg.charger_connected; }
```
Kotochleb: This could be replaced with a lambda

panther_battery/src/adc_node.cpp
Kotochleb: As mentioned before. The `LSB` should be more explicitly computed, or just coded in a way a "voltage_divisor" value

panther_battery/include/panther_battery/adc_node.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+private:
+  static constexpr float bat_charging_curr_thresh_ = 0.1;
+  static constexpr float bat02_detect_thresh_ = 3.03;
+  static constexpr float bat_designed_capacity_ = 20.0;
+  static constexpr float A_ = 298.15;
+  static constexpr float B_ = 3977.0;
+  static constexpr float R1_ = 10000.0;
+  static constexpr float R0_ = 10000.0;
+  static constexpr float u_supply_ = 3.28;
```
Kotochleb: Unless you read the code related to the temperatures, those constants are unclear. Also they might collide with computations of  what is currently called `LSB`

panther_battery/src/adc_node.cpp
Kotochleb: There is mix of `flaot` and `double` precision. Either change all values to be `double` or use `std::logf`

panther_battery/include/panther_battery/adc_node.hpp
```diff
@@ -0,0 +1,73 @@
...
+  float battery_timeout_;
+  float high_bat_temp_;
+  float V_bat_1_;
+  float V_bat_2_;
+  float V_temp_bat_1_;
+  float V_temp_bat_2_;
+  float I_charge_bat_1_;
+  float I_charge_bat_2_;
+  float I_bat_1_;
+  float I_bat_2_;
```
Kotochleb: Since the code will fail if ADC is broken and the `try.. catch` is not really required, in this case those values might as well be defined in the callback

panther_battery/include/panther_battery/adc_node.hpp
```diff
@@ -0,0 +1,73 @@
...
+  std::string adc0_device_;
+  std::string adc1_device_;
```
Kotochleb: Since this string is only used in constructor, it does not have to be class member

panther_battery/README.md
```diff
@@ -18,6 +43,6 @@ Node publishing Panther battery state read from motor controllers. Used in Panth
 
 #### Parameters
 
+- `~battery_current_window_len` [*int*, default: **10**]: moving average window length used to smooth out battery current readings.
```
Kotochleb:
```suggestion
- `~battery_current_window_len` [*int*, default: **10**]: window length of a moving average, used to smooth out battery current readings.
```

panther_battery/README.md
```diff
@@ -18,6 +43,6 @@ Node publishing Panther battery state read from motor controllers. Used in Panth
...
 - `~battery_timeout` [*float*, default: **1.0**]: specifies the timeout in seconds. If no new battery messages are received within this duration, the node will publish an unknown battery state.
-- `~batery_voltage_window_len` [*int*, default: **10**]: moving average window length used to smooth out battery voltage readings.
-- `~batery_current_window_len` [*int*, default: **10**]: moving average window length used to smooth out battery current readings.
\ No newline at end of file
+- `~batery_voltage_window_len` [*int*, default: **10**]: moving average window length used to smooth out battery voltage readings.
```
Kotochleb:
```suggestion
- `~batery_voltage_window_len` [*int*, default: **10**]: window length of a moving average, used to smooth out battery voltage readings.
```

panther_battery/README.md
```diff
@@ -4,7 +4,32 @@ Package containing nodes monitoring and publishing internal battery state of the
...
+
+#### Subscribes
+
+- `/panther/driver/motor_controllers_state` [*panther_msgs/DriverState*]: current motor controllers' state and error flags.
+
+#### Parameters
+
+- `~adc0_device` [*string*, default: **/sys/bus/iio/devices/iio:device0**]: ADC nr 0 device IIO path.
+- `~adc1_device` [*string*, default: **/sys/bus/iio/devices/iio:device0**]: ADC nr 0 device IIO path.
+- `~battery_charge_window_len` [*int*, default: **10**]: moving average window length used to smooth out battery charge readings.
```
Kotochleb:
```suggestion
- `~battery_charge_window_len` [*int*, default: **10**]: window length of a moving average, used to smooth out battery charge readings.
```

panther_battery/README.md
```diff
@@ -4,7 +4,32 @@ Package containing nodes monitoring and publishing internal battery state of the
...
+- `~battery_current_window_len` [*int*, default: **10**]: moving average window length used to smooth out battery current readings.
```
Kotochleb:
```suggestion
- `~battery_current_window_len` [*int*, default: **10**]: window length of a moving average, used to smooth out battery current readings.
```

panther_battery/README.md
```diff
@@ -4,7 +4,32 @@ Package containing nodes monitoring and publishing internal battery state of the
...
+- `~battery_temp_window_len` [*int*, default: **10**]: moving average window length used to smooth out battery temperature readings.
```
Kotochleb:
```suggestion
- `~battery_temp_window_len` [*int*, default: **10**]: window length of a moving average, used to smooth out battery temperature readings.
```

panther_battery/README.md
```diff
@@ -4,7 +4,32 @@ Package containing nodes monitoring and publishing internal battery state of the
...
+- `~battery_timeout` [*float*, default: **1.0**]: specifies the timeout in seconds. If node fails to read ADC battery data exceeding this duration, the node will publish an unknown battery state.
```
Kotochleb, line 25: As mentioned later. Not sure it this is necessary

panther_battery/README.md
```diff
@@ -4,7 +4,32 @@ Package containing nodes monitoring and publishing internal battery state of the
...
+- `~battery_voltage_window_len` [*int*, default: **10**]: moving average window length used to smooth out battery voltage readings.
```
Kotochleb:
```suggestion
- `~battery_voltage_window_len` [*int*, default: **10**]: window length of a moving average, used to smooth out battery voltage readings.
```

panther_battery/README.md
```diff
@@ -4,7 +4,32 @@ Package containing nodes monitoring and publishing internal battery state of the
...
+- `~high_bat_temp` [*float*, default: **55.0**]: The temperature of the battery at which the battery health state is incorrect.
```
Kotochleb:
```suggestion
- `~high_bat_temp` [*float*, default: **55.0**]: The temperature of the battery at which is ic considered to overheat.
```

panther_battery/README.md
```diff
@@ -4,7 +4,32 @@ Package containing nodes monitoring and publishing internal battery state of the
...
 ## ROS Nodes
 
-### roboteq_republisher_node.py
+### adc_node
+
+Publishes battery state read from ADC unit. Available from Panther version 1.2.
+
+#### Publishes
+
+- `/panther/battery` [*sensor_msgs/BatteryState*]: average values of both batteries if panther has two batteries. In case of single battery values only for the single one.
```
Kotochleb:
```suggestion
- `/panther/battery` [*sensor_msgs/BatteryState*]: mean values of both batteries if panther has two batteries. Otherwise state of the single battery will be published.
```

panther_battery/include/panther_battery/battery_publisher.hpp
```diff
@@ -0,0 +1,182 @@
...
+  battery_msg.power_supply_status = BatteryStateMsg::POWER_SUPPLY_STATUS_UNKNOWN;
+  battery_msg.power_supply_health = BatteryStateMsg::POWER_SUPPLY_HEALTH_UNKNOWN;
+  battery_msg.power_supply_technology = BatteryStateMsg::POWER_SUPPLY_TECHNOLOGY_LIPO;
+  battery_msg.present = true;
+  battery_msg.location = location_;
+
+  battery_pub_->publish(battery_msg);
+}
+
+inline void BatteryPublisher::Publish(
```
Kotochleb: There is something I don't like about that many parameters in the function. I might look better if there were for example setters for this. But it is up to discussion

panther_battery/include/panther_battery/battery_publisher.hpp
```diff
@@ -0,0 +1,182 @@
...
+  battery_msg.charge = battery_msg.percentage * battery_msg.design_capacity;
+  battery_msg.cell_voltage = std::vector<float>(10, std::numeric_limits<float>::quiet_NaN());
+  battery_msg.cell_temperature = std::vector<float>(10, std::numeric_limits<float>::quiet_NaN());
+  battery_msg.power_supply_technology = BatteryStateMsg::POWER_SUPPLY_TECHNOLOGY_LIPO;
+  battery_msg.present = true;
+  battery_msg.location = location_;
+
+  // check battery status
+  if (charger_connected) {
+    if (battery_msg.percentage == 1.0) {
```
Kotochleb: This will never be true. In c++ comparing floats has to be done this way:
```suggestion
    if (fabs(battery_msg.percentage - 1.0f) < std::numeric_limits<float>::epsilon()) {
```

panther_battery/include/panther_battery/battery_publisher.hpp
Kotochleb, line 1: Relating to internal discussions there may occur some changes in constants and treatment of some fields of the message

panther_battery/test/test_adc_data_reader.cpp
```diff
@@ -0,0 +1,61 @@
...
+  std::filesystem::path file_path_;
+  std::ofstream file_;
+
+  void WriteNumberToFile(const int & number);
+};
+
+TestADCDataReader::TestADCDataReader()
+{
+  current_path_ = std::filesystem::current_path();
+  file_path_ = current_path_ / "test_file";
```
Kotochleb: This should not be regular file, but rather [tmpfile](https://en.cppreference.com/w/cpp/io/c/tmpfile), stored in `/tmp` directory

panther_battery/test/test_adc_data_reader.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  data_reader_ = std::make_shared<panther_battery::ADCDataReader>(current_path_);
+}
+
+TestADCDataReader::~TestADCDataReader(){
+  std::filesystem::remove(file_path_);
+}
+
+void TestADCDataReader::WriteNumberToFile(const int & number)
```
Kotochleb:
```suggestion
void TestADCDataReader::WriteNumberToFile(const int number)
```

Kotochleb (changes requested):
I left test not checked since they will most likely change a lot after applying review changes.
Anyway, good idea with mocking the ADC files. I am only afraid there will be still many changes due to still shaped battery message.

Please remove reference to basic types.
One more thing, the way battery publisher is so general it might be used for both nodes.

Kotochleb (changes requested):
I left test not checked since they will most likely change a lot after applying review changes.
Anyway, good idea with mocking the ADC files. I am only afraid there will be still many changes due to still shaped battery message.

Please remove reference to basic types.
One more thing, the way battery publisher is so general it might be used for both nodes.

panther_battery/include/panther_battery/adc_data_reader.hpp
KmakD: Short explanation: without it code won't compile
Long one: when you define a function in a C++ header file, you need to use the inline keyword to allow the function to be defined in multiple translation units (source files) without causing a linker error.

panther_battery/src/adc_node.cpp
KmakD: Timeout was added for cases when we have short temporary fault in readings. This way single reading error won't cause publishing random unknown battery message. 
Checking witch battery failed would require additional logic and storing failures. Also, each battery requires readings from both ADC and most likely error on one will mean failure for both batteries. This raises question is it worth to implement this logic?

panther_battery/src/adc_node.cpp
KmakD: There is no method `as_float()`. However, if you declare variable as float and use `as_double()` it will cast it to float.

panther_battery/src/adc_node.cpp
KmakD: Problem with battery object is that you need `i_bat_2` to calculate battery current when you have single battery. So then you will have logic like if battery_count==1, then use data from battery_1 and battery_2, witch looks super 
counter-intuitive.

panther_battery/include/panther_battery/adc_data_reader.hpp
KmakD: Done

panther_battery/include/panther_battery/adc_data_reader.hpp
KmakD: Done

panther_battery/include/panther_battery/adc_data_reader.hpp
KmakD: Done

panther_battery/include/panther_battery/adc_data_reader.hpp
KmakD: Done

panther_battery/include/panther_battery/adc_data_reader.hpp
KmakD: Done

panther_battery/include/panther_battery/adc_data_reader.hpp
KmakD: Done

panther_battery/include/panther_battery/battery_publisher.hpp
KmakD: Done

panther_battery/src/adc_node.cpp
KmakD: Done

panther_battery/src/adc_node.cpp
KmakD: Done

panther_battery/src/adc_node.cpp
KmakD: Not relevant any more

panther_battery/src/adc_node.cpp
KmakD: Done

panther_battery/src/adc_node.cpp
KmakD: Done

panther_battery/src/adc_node.cpp
KmakD: Done

panther_battery/src/adc_node.cpp
KmakD: Done

panther_battery/include/panther_battery/battery_publisher.hpp
KmakD: Done

panther_battery/src/adc_node.cpp
KmakD: Done

panther_battery/src/adc_node.cpp
KmakD: Done

panther_battery/include/panther_battery/adc_node.hpp
KmakD: Done

panther_battery/README.md
KmakD: Done

panther_battery/README.md
KmakD: Done

panther_battery/include/panther_battery/battery_publisher.hpp
KmakD: Done

panther_battery/include/panther_battery/battery_publisher.hpp
KmakD: I'm not exactly sure if this can be applied as not all battery states/health can be computed with roboteq data

panther_battery/include/panther_battery/battery_publisher.hpp
KmakD: I've changed this class to Battery with method to get battery msg. battery state is published in adc_node

panther_battery/src/adc_node.cpp
KmakD: Done

panther_battery/include/panther_battery/battery_publisher.hpp
KmakD: Had the same feeling. I replaced it with struct with params

panther_battery/include/panther_battery/battery_publisher.hpp
KmakD: Done

panther_battery/include/panther_battery/adc_data_reader.hpp
Kotochleb, line 1: Move the implementation to be within class body

panther_battery/include/panther_battery/battery.hpp
```diff
@@ -0,0 +1,293 @@
...
+  static constexpr float bat_charge_factor_ = 2.5;
+  static constexpr float bat_temp_factor_ = 1.0;
+
+  // Temperature conversion parameters
+  static constexpr double temp_coeff_A_ = 298.15;
+  static constexpr double temp_coeff_B_ = 3977.0;
+  static constexpr double R1_ = 10000.0;
+  static constexpr double R0_ = 10000.0;
+  static constexpr double u_supply_ = 3.28;
+  static constexpr double kelvin_to_celcius_offset_ = 273.15;
```
Kotochleb: I would check how much error if we have if we use floating point

panther_battery/include/panther_battery/battery.hpp
```diff
@@ -0,0 +1,293 @@
...
+  std::function<float()> ReadTemp;
+  std::function<float()> ReadCharge;
+
+  BatteryStateMsg battery_state_;
+  BatteryStateMsg battery_state_raw_;
+
+  std::unique_ptr<panther_utils::MovingAverage<double>> voltage_ma_;
+  std::unique_ptr<panther_utils::MovingAverage<double>> temp_ma_;
+  std::unique_ptr<panther_utils::MovingAverage<double>> current_ma_;
+  std::unique_ptr<panther_utils::MovingAverage<double>> charge_ma_;
```
Kotochleb: For most of the time float is used. Is there a reason why this is double?

panther_battery/include/panther_battery/battery.hpp
Kotochleb, line 1: Move function bodies to sepeare .cpp file

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,213 @@
...
+    std::bind(&ADCDataReader::GetADCMeasurement, *adc0_reader_, 2, 0), default_battery_params);
+
+  battery_count_ = battery_2_->Present() ? 2 : 1;
+  if (battery_count_ == 2) {
+    RCLCPP_INFO(this->get_logger(), "Second battery detected");
+    battery_1_ = std::make_unique<Battery>(
+      std::bind(&ADCDataReader::GetADCMeasurement, *adc1_reader_, 0, 0),
+      std::bind(&ADCDataReader::GetADCMeasurement, *adc1_reader_, 2, adc_current_offset_),
+      std::bind(&ADCDataReader::GetADCMeasurement, *adc0_reader_, 1, 0),
+      std::bind(&ADCDataReader::GetADCMeasurement, *adc0_reader_, 3, 0), default_battery_params);
```
Kotochleb: This dereferences `unique_ptr` breaking it being unique

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+  if (battery_count_ == 2) {
+    battery_pub_->publish(
+      MergeBatteryMsgs(battery_1_->GetBatteryMsg(), battery_2_->GetBatteryMsg()));
+    battery_1_pub_->publish(battery_1_->GetBatteryMsgRaw());
+    battery_2_pub_->publish(battery_2_->GetBatteryMsgRaw());
+  } else {
+    battery_pub_->publish(battery_1_->GetBatteryMsg());
+    battery_1_pub_->publish(battery_1_->GetBatteryMsgRaw());
+  }
```
Kotochleb: Battery message object could be created only once and later passed to the publishers

Kotochleb (changes requested):
This is only draft review, yet some changes are still required

panther_battery/include/panther_battery/battery.hpp
KmakD: Input values are float, I used double to reduce the error when doing calculations.

panther_battery/src/adc_node.cpp
KmakD: What's the point in creating the battery msg object that is used only once?

panther_battery/include/panther_battery/battery.hpp
KmakD: Err is around 0.00001

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,55 @@
...
+  float GetADCMeasurement(const int channel, const int offset)
+  {
+    auto LSB = ReadChannel<float>(channel, "scale") / 1000;
+    auto raw_value = ReadChannel<int>(channel, "raw");
+    return (raw_value - offset) * LSB;
+  }
+
+private:
+  template <typename T>
+  T ReadChannel(const int channel, const std::string data_type)
```
Kotochleb:
```suggestion
  T ReadChannel(const int channel, const std::string & data_type)
```

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,55 @@
...
+    T data;
+    auto data_file = "in_voltage" + std::to_string(channel) + "_" + data_type;
+    auto file_path = device_path_ / data_file;
+
+    std::fstream file(file_path, std::ios_base::in);
+    if (!file) {
+      throw std::runtime_error("Failed to open file: " + std::string(file_path));
+    }
+
+    file >> data;
```
Kotochleb:
```suggestion
    auto data_file = "in_voltage" + std::to_string(channel) + "_" + data_type;
    auto file_path = device_path_ / data_file;

    std::fstream file(file_path, std::ios_base::in);
    if (!file) {
      throw std::runtime_error("Failed to open file: " + std::string(file_path));
    }

    T data;
    file >> data;
```

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,55 @@
...
+namespace panther_battery
+{
+
+class ADCDataReader
+{
+public:
+  ADCDataReader(const std::string & device_path) : device_path_(device_path)
+  {
+    if (!std::filesystem::exists(device_path_)) {
+      throw std::runtime_error("Failed to open device path: " + std::string(device_path_));
```
Kotochleb: I think "Failed to open device is a bit confusing. This error might occur while reading, but on initialization it might be better to say that the device does not exist.

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,55 @@
...
+      throw std::logic_error("Invalid data type: " + data_type);
+    }
+
+    T data;
+    auto data_file = "in_voltage" + std::to_string(channel) + "_" + data_type;
+    auto file_path = device_path_ / data_file;
+
+    std::fstream file(file_path, std::ios_base::in);
+    if (!file) {
+      throw std::runtime_error("Failed to open file: " + std::string(file_path));
```
Kotochleb, line 49:
```suggestion
      throw std::runtime_error("Failed to open device file: " + std::string(file_path));
```

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,55 @@
...
+  template <typename T>
+  T ReadChannel(const int channel, const std::string data_type)
+  {
+    if (data_type != "raw" && data_type != "scale" && data_type != "sampling_frequency") {
+      throw std::logic_error("Invalid data type: " + data_type);
+    }
+
+    T data;
+    auto data_file = "in_voltage" + std::to_string(channel) + "_" + data_type;
+    auto file_path = device_path_ / data_file;
```
Kotochleb:
```suggestion
    const auto data_file = "in_voltage" + std::to_string(channel) + "_" + data_type;
    const auto file_path = device_path_ / data_file;
```

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,55 @@
...
+  {
+    if (!std::filesystem::exists(device_path_)) {
+      throw std::runtime_error("Failed to open device path: " + std::string(device_path_));
+    }
+  }
+
+  float GetADCMeasurement(const int channel, const int offset)
+  {
+    auto LSB = ReadChannel<float>(channel, "scale") / 1000;
+    auto raw_value = ReadChannel<int>(channel, "raw");
```
Kotochleb:
```suggestion
    const auto LSB = ReadChannel<float>(channel, "scale") / 1000;
    const auto raw_value = ReadChannel<int>(channel, "raw");
```

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,55 @@
...
+{
+public:
+  ADCDataReader(const std::string & device_path) : device_path_(device_path)
+  {
+    if (!std::filesystem::exists(device_path_)) {
+      throw std::runtime_error("Failed to open device path: " + std::string(device_path_));
+    }
+  }
+
+  float GetADCMeasurement(const int channel, const int offset)
```
Kotochleb:
```suggestion
  float GetADCMeasurement(const int channel, const int offset) const
```

panther_battery/include/panther_battery/adc_data_reader.hpp
Kotochleb:
```suggestion
  T ReadChannel(const int channel, const std::string data_type) const
```

panther_battery/include/panther_battery/battery.hpp
```diff
@@ -0,0 +1,105 @@
...
+  float voltage_raw_;
+  float current_raw_;
+  float temp_raw_;
+  float charge_raw_;
+  std::string error_msg_;
+
+  std::function<float()> ReadVoltage;
+  std::function<float()> ReadCurrent;
+  std::function<float()> ReadTemp;
+  std::function<float()> ReadCharge;
```
Kotochleb:
```suggestion
  const std::function<float()> ReadVoltage;
  const std::function<float()> ReadCurrent;
  const std::function<float()> ReadTemp;
  const std::function<float()> ReadCharge;
```

panther_battery/include/panther_battery/battery.hpp
```diff
@@ -0,0 +1,105 @@
...
+  const int64_t charge_window_len;
+};
+
+class Battery
+{
+public:
+  Battery(
+    std::function<float()> read_voltage, std::function<float()> read_current,
+    std::function<float()> read_temp, std::function<float()> read_charge,
+    const BatteryParams & params);
```
Kotochleb:
```suggestion
  Battery(
  const std::function<float()> & read_voltage, const std::function<float()> & read_current,
  const std::function<float()> & read_temp, const std::function<float()> & read_charge,
  const BatteryParams & params);
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include <sensor_msgs/msg/battery_state.hpp>
+
+#include <panther_utils/moving_average.hpp>
+
+namespace panther_battery
+{
+
+Battery::Battery(
+  std::function<float()> read_voltage, std::function<float()> read_current,
+  std::function<float()> read_temp, std::function<float()> read_charge,
```
Kotochleb:
```suggestion
  const std::function<float()> & read_voltage, const std::function<float()> & read_current,
  const std::function<float()> & read_temp, const std::function<float()> & read_charge,
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+    params.temp_window_len, std::numeric_limits<double>::quiet_NaN());
+  current_ma_ = std::make_unique<panther_utils::MovingAverage<double>>(
+    params.current_window_len, std::numeric_limits<double>::quiet_NaN());
+  charge_ma_ = std::make_unique<panther_utils::MovingAverage<double>>(
+    params.charge_window_len, std::numeric_limits<double>::quiet_NaN());
+}
+
+bool Battery::Present()
+{
+  const auto get_temp_attempts = 10;
```
Kotochleb: This should be either a parameter or static class constant, not something declared in the function

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+  auto V_temp_sum = 0.0f;
+
+  for (int i = 0; i < get_temp_attempts; i++) {
+    V_temp_sum += ADCToBatteryVoltageTemp(ReadTemp());
+    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+  }
+
+  const auto V_temp_bat = V_temp_sum / get_temp_attempts;
```
Kotochleb:
```suggestion
  const auto V_temp_bat = V_temp_sum / static_cast<float>(get_temp_attempts);
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+  return V_temp_bat < bat_detect_thresh_;
+}
+
+void Battery::Update(rclcpp::Time & header_stamp, const bool charger_connected)
```
Kotochleb:
```suggestion
void Battery::Update(const rclcpp::Time & header_stamp, const bool charger_connected)
```

panther_battery/include/panther_battery/battery.hpp
```diff
@@ -0,0 +1,105 @@
...
+  BatteryStateMsg GetBatteryMsgRaw() const;
+
+private:
+  float ADCToBatteryVoltage(const float adc_data);
+  float ADCToBatteryCurrent(const float adc_data);
+  float ADCToBatteryCharge(const float adc_data);
+  float ADCToBatteryTemp(const float adc_data);
+  float ADCToBatteryVoltageTemp(const float adc_data);
+  void UpdateBatteryMsg(rclcpp::Time & header_stamp, const bool charger_connected);
+  void ResetBatteryMsg(rclcpp::Time & header_stamp);
```
Kotochleb:
```suggestion
  void UpdateBatteryMsg(const rclcpp::Time & header_stamp, const bool charger_connected);
  void ResetBatteryMsg(const rclcpp::Time & header_stamp);
```

panther_battery/include/panther_battery/battery.hpp
```diff
@@ -0,0 +1,105 @@
...
+  Battery(
+    std::function<float()> read_voltage, std::function<float()> read_current,
+    std::function<float()> read_temp, std::function<float()> read_charge,
+    const BatteryParams & params);
+
+  ~Battery() {}
+
+  bool Present();
+  void Update(rclcpp::Time & header_stamp, const bool charger_connected);
+  void Reset(rclcpp::Time & header_stamp);
```
Kotochleb:
```suggestion
  void Update(const rclcpp::Time & header_stamp, const bool charger_connected);
  void Reset(const rclcpp::Time & header_stamp);
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+  ResetBatteryMsg(header_stamp);
+}
+
+float Battery::ADCToBatteryVoltage(const float adc_data) { return adc_data * bat_voltage_factor_; }
+
+float Battery::ADCToBatteryCurrent(const float adc_data) { return adc_data * bat_current_factor_; }
+
+float Battery::ADCToBatteryCharge(const float adc_data) { return adc_data * bat_charge_factor_; }
+
+float Battery::ADCToBatteryTemp(const float adc_data)
```
Kotochleb:
```suggestion
float Battery::ADCToBatteryVoltage(const float adc_data) const { return adc_data * bat_voltage_factor_; }

float Battery::ADCToBatteryCurrent(const float adc_data) const { return adc_data * bat_current_factor_; }

float Battery::ADCToBatteryCharge(const float adc_data) const { return adc_data * bat_charge_factor_; }

float Battery::ADCToBatteryTemp(const float adc_data) const
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+{
+  auto V_temp = ADCToBatteryVoltageTemp(adc_data);
```
Kotochleb:
```suggestion
  const auto V_temp = ADCToBatteryVoltageTemp(adc_data);
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+  if (V_temp == 0 || V_temp >= u_supply_) {
+    return std::numeric_limits<float>::quiet_NaN();
+  }
+
+  auto R_therm = (V_temp * R1_) / (u_supply_ - V_temp);
```
Kotochleb:
```suggestion
  const auto R_therm = (V_temp * R1_) / (u_supply_ - V_temp);
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+  return (temp_coeff_A_ * temp_coeff_B_ / (temp_coeff_A_ * log(R_therm / R0_) + temp_coeff_B_)) -
+         kelvin_to_celcius_offset_;
+}
+
+float Battery::ADCToBatteryVoltageTemp(const float adc_data) { return adc_data * bat_temp_factor_; }
```
Kotochleb:
```suggestion
float Battery::ADCToBatteryVoltageTemp(const float adc_data) const { return adc_data * bat_temp_factor_; }
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+float Battery::ADCToBatteryVoltage(const float adc_data) { return adc_data * bat_voltage_factor_; }
+
+float Battery::ADCToBatteryCurrent(const float adc_data) { return adc_data * bat_current_factor_; }
+
+float Battery::ADCToBatteryCharge(const float adc_data) { return adc_data * bat_charge_factor_; }
+
+float Battery::ADCToBatteryTemp(const float adc_data)
+{
+  auto V_temp = ADCToBatteryVoltageTemp(adc_data);
+  if (V_temp == 0 || V_temp >= u_supply_) {
```
Kotochleb: Since `V_temp` is float, `==` comparison is incorrect. It should check if the values are numerically close. Also `u_supply_` is type of `u_supply_` leading to quiet promotion of `float` to `double` to achieve unused precision

panther_battery/src/battery.cpp
Kotochleb, line 1: I think those function that multiply something and then return it could be header-only if they will format there into single line

panther_battery/include/panther_battery/battery.hpp
```diff
@@ -0,0 +1,105 @@
...
+  std::string GetErrorMsg() const;
+  BatteryStateMsg GetBatteryMsg() const;
+  BatteryStateMsg GetBatteryMsgRaw() const;
+
+private:
+  float ADCToBatteryVoltage(const float adc_data);
+  float ADCToBatteryCurrent(const float adc_data);
+  float ADCToBatteryCharge(const float adc_data);
+  float ADCToBatteryTemp(const float adc_data);
+  float ADCToBatteryVoltageTemp(const float adc_data);
```
Kotochleb:
```suggestion
  inline float ADCToBatteryVoltage(const float adc_data);
  inline float ADCToBatteryCurrent(const float adc_data);
  inline float ADCToBatteryCharge(const float adc_data);
  float ADCToBatteryTemp(const float adc_data);
  inline float ADCToBatteryVoltageTemp(const float adc_data);
```
`ADCToBatteryTemp` was left not inline intentionally

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+  charge_ma_->Reset();
+
+  ResetBatteryMsg(header_stamp);
+}
+
+float Battery::ADCToBatteryVoltage(const float adc_data) { return adc_data * bat_voltage_factor_; }
+
+float Battery::ADCToBatteryCurrent(const float adc_data) { return adc_data * bat_current_factor_; }
+
+float Battery::ADCToBatteryCharge(const float adc_data) { return adc_data * bat_charge_factor_; }
```
Kotochleb:
```suggestion
inline float Battery::ADCToBatteryVoltage(const float adc_data) { return adc_data * bat_voltage_factor_; }

inline float Battery::ADCToBatteryCurrent(const float adc_data) { return adc_data * bat_current_factor_; }

inline float Battery::ADCToBatteryCharge(const float adc_data) { return adc_data * bat_charge_factor_; }
```

panther_battery/src/battery.cpp
Kotochleb:
```suggestion
inline float Battery::ADCToBatteryVoltageTemp(const float adc_data) { return adc_data * bat_temp_factor_; }
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  auto R_therm = (V_temp * R1_) / (u_supply_ - V_temp);
+  return (temp_coeff_A_ * temp_coeff_B_ / (temp_coeff_A_ * log(R_therm / R0_) + temp_coeff_B_)) -
+         kelvin_to_celcius_offset_;
+}
+
+float Battery::ADCToBatteryVoltageTemp(const float adc_data) { return adc_data * bat_temp_factor_; }
+
+void Battery::UpdateBatteryMsg(rclcpp::Time & header_stamp, const bool charger_connected)
```
Kotochleb:
```suggestion
void Battery::UpdateBatteryMsg(const rclcpp::Time & header_stamp, const bool charger_connected)
```

panther_battery/include/panther_battery/adc_data_reader.hpp
```diff
@@ -0,0 +1,55 @@
...
+  {
+    auto LSB = ReadChannel<float>(channel, "scale") / 1000;
```
Kotochleb:
```suggestion
    auto LSB = ReadChannel<float>(channel, "scale") / 1000.0f;
```

panther_battery/include/panther_battery/battery.hpp
Kotochleb: I plotted the results and I think numerical error lower than 1e-4 is much lower than measurement error and we could use floats everywhere. You can see plots below:

image

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+{
+  auto V_bat = voltage_ma_->GetAverage();
+  auto I_bat = current_ma_->GetAverage();
+  auto temp_bat = temp_ma_->GetAverage();
+  auto I_charge = charge_ma_->GetAverage();
```
Kotochleb:
```suggestion
  const auto V_bat = voltage_ma_->GetAverage();
  const auto I_bat = current_ma_->GetAverage();
  const auto temp_bat = temp_ma_->GetAverage();
  const auto I_charge = charge_ma_->GetAverage();
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+  battery_state_.header.stamp = header_stamp;
+  battery_state_.voltage = V_bat;
+  battery_state_.temperature = temp_bat;
+  battery_state_.current = -I_bat + I_charge;
+  battery_state_.percentage =
+    std::clamp((V_bat - V_bat_min_) / (V_bat_full_ - V_bat_min_), 0.0, 1.0);
```
Kotochleb:
```suggestion
    std::clamp((V_bat - V_bat_min_) / (V_bat_full_ - V_bat_min_), 0.0f, 1.0f);
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+  battery_state_.capacity = std::numeric_limits<double>::quiet_NaN();
```
Kotochleb:
```suggestion
  battery_state_.capacity = std::numeric_limits<float>::quiet_NaN();
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+void Battery::ResetBatteryMsg(rclcpp::Time & header_stamp)
+{
+  battery_state_.header.stamp = header_stamp;
+  battery_state_.voltage = std::numeric_limits<double>::quiet_NaN();
+  battery_state_.temperature = std::numeric_limits<double>::quiet_NaN();
+  battery_state_.current = std::numeric_limits<double>::quiet_NaN();
+  battery_state_.percentage = std::numeric_limits<double>::quiet_NaN();
+  battery_state_.capacity = std::numeric_limits<double>::quiet_NaN();
+  battery_state_.design_capacity = designed_capacity_;
+  battery_state_.charge = std::numeric_limits<double>::quiet_NaN();
```
Kotochleb:
```suggestion
  battery_state_.voltage = std::numeric_limits<float>::quiet_NaN();
  battery_state_.temperature = std::numeric_limits<float>::quiet_NaN();
  battery_state_.current = std::numeric_limits<float>::quiet_NaN();
  battery_state_.percentage = std::numeric_limits<float>::quiet_NaN();
  battery_state_.capacity = std::numeric_limits<float>::quiet_NaN();
  battery_state_.design_capacity = designed_capacity_;
  battery_state_.charge = std::numeric_limits<float>::quiet_NaN();
```

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+  battery_state_.location = location_;
+}
+
+bool Battery::HasErrorMsg() const
+{
+  if (error_msg_.empty()) {
+    return false;
+  }
+  return true;
+}
```
Kotochleb:
```suggestion
bool Battery::HasErrorMsg() const { return !error_msg_.empty(); }
```

panther_battery/src/battery.cpp
Kotochleb, line 1: I think the number of cells which is everywhere hard-coded as 10 could be a constant within the class

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+: ReadVoltage(read_voltage), ReadCurrent(read_current), ReadTemp(read_temp), ReadCharge(read_charge)
+{
+  voltage_ma_ = std::make_unique<panther_utils::MovingAverage<double>>(
+    params.voltage_window_len, std::numeric_limits<double>::quiet_NaN());
+  temp_ma_ = std::make_unique<panther_utils::MovingAverage<double>>(
+    params.temp_window_len, std::numeric_limits<double>::quiet_NaN());
+  current_ma_ = std::make_unique<panther_utils::MovingAverage<double>>(
+    params.current_window_len, std::numeric_limits<double>::quiet_NaN());
+  charge_ma_ = std::make_unique<panther_utils::MovingAverage<double>>(
+    params.charge_window_len, std::numeric_limits<double>::quiet_NaN());
```
Kotochleb:
```suggestion
  voltage_ma_ = std::make_unique<panther_utils::MovingAverage<float>>(
    params.voltage_window_len, std::numeric_limits<float>::quiet_NaN());
  temp_ma_ = std::make_unique<panther_utils::MovingAverage<float>>(
    params.temp_window_len, std::numeric_limits<float>::quiet_NaN());
  current_ma_ = std::make_unique<panther_utils::MovingAverage<float>>(
    params.current_window_len, std::numeric_limits<float>::quiet_NaN());
  charge_ma_ = std::make_unique<panther_utils::MovingAverage<float>>(
    params.charge_window_len, std::numeric_limits<float>::quiet_NaN());
```

panther_battery/include/panther_battery/battery.hpp
Kotochleb:
```suggestion
  std::unique_ptr<panther_utils::MovingAverage<float>> voltage_ma_;
  std::unique_ptr<panther_utils::MovingAverage<float>> temp_ma_;
  std::unique_ptr<panther_utils::MovingAverage<float>> current_ma_;
  std::unique_ptr<panther_utils::MovingAverage<float>> charge_ma_;
```

I will still opt for the floats. Those are linear functions and since numerical error in temperature is lower than measurement error I will opt for the floats. For other measurements, functions are linear hence the numerical error will be even lower

panther_battery/src/battery.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+float Battery::ADCToBatteryTemp(const float adc_data)
+{
+  auto V_temp = ADCToBatteryVoltageTemp(adc_data);
+  if (V_temp == 0 || V_temp >= u_supply_) {
+    return std::numeric_limits<float>::quiet_NaN();
+  }
+
+  auto R_therm = (V_temp * R1_) / (u_supply_ - V_temp);
+  return (temp_coeff_A_ * temp_coeff_B_ / (temp_coeff_A_ * log(R_therm / R0_) + temp_coeff_B_)) -
```
Kotochleb:
```suggestion
  return (temp_coeff_A_ * temp_coeff_B_ / (temp_coeff_A_ * logf(R_therm / R0_) + temp_coeff_B_)) -
```

panther_battery/include/panther_battery/battery.hpp
```diff
@@ -0,0 +1,105 @@
...
+{
+
+using BatteryStateMsg = sensor_msgs::msg::BatteryState;
+
+struct BatteryParams
+{
+  const int64_t voltage_window_len;
+  const int64_t temp_window_len;
+  const int64_t current_window_len;
+  const int64_t charge_window_len;
```
Kotochleb:
```suggestion
  const std::size_t voltage_window_len;
  const std::size_t temp_window_len;
  const std::size_t current_window_len;
  const std::size_t charge_window_len;
```

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,213 @@
...
+  adc0_reader_ = std::make_shared<ADCDataReader>(adc0_device);
+  adc1_reader_ = std::make_shared<ADCDataReader>(adc1_device);
+  last_battery_info_time_ = rclcpp::Time(int64_t(0), RCL_ROS_TIME);
+
+  BatteryParams default_battery_params = {
+    battery_voltage_window_len,
+    battery_temp_window_len,
+    battery_current_window_len,
+    battery_charge_window_len,
+  };
```
Kotochleb:
```suggestion
  const BatteryParams default_battery_params = {
    static_cast<std::size_t>(battery_voltage_window_len),
    static_cast<std::size_t>(battery_temp_window_len),
    static_cast<std::size_t>(battery_current_window_len),
    static_cast<std::size_t>(battery_charge_window_len),
  };
```

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,213 @@
...
+      std::bind(&ADCDataReader::GetADCMeasurement, *adc0_reader_, 3, 0), default_battery_params);
+    battery_2_.reset();
+  }
+
+  battery_pub_ = this->create_publisher<BatteryStateMsg>("battery", 10);
+  battery_1_pub_ = this->create_publisher<BatteryStateMsg>("battery_1_raw", 10);
+
+  io_state_sub_ = this->create_subscription<IOStateMsg>(
+    "hardware/io_state", 10,
+    [&](const IOStateMsg::SharedPtr msg) { charger_connected_ = msg->charger_connected; });
```
Kotochleb:
```suggestion
    [&](const IOStateMsg & msg) { charger_connected_ = msg.charger_connected; });
```

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,213 @@
...
+  // running at 10 Hz
+  battery_pub_timer_ = this->create_wall_timer(
+    std::chrono::milliseconds(100), std::bind(&ADCNode::BatteryPubTimerCB, this));
+
+  RCLCPP_INFO(this->get_logger(), "Node started");
+}
+
+void ADCNode::BatteryPubTimerCB()
+{
+  auto header_stamp = this->get_clock()->now();
```
Kotochleb:
```suggestion
  const auto header_stamp = this->get_clock()->now();
```

panther_battery/src/adc_node.cpp
```diff
@@ -0,0 +1,213 @@
...
+    if (battery_2_->HasErrorMsg()) {
+      RCLCPP_ERROR_THROTTLE(
+        this->get_logger(), *this->get_clock(), 10000, "Battery nr 2 error: %s",
+        battery_2_->GetErrorMsg().c_str());
+    }
+  }
+}
+
+BatteryStateMsg ADCNode::MergeBatteryMsgs(
+  const BatteryStateMsg & battery_msg_1, const BatteryStateMsg & battery_msg_2)
```
Kotochleb:
```suggestion
  const BatteryStateMsg & battery_msg_1, const BatteryStateMsg & battery_msg_2) const
```

panther_battery/include/panther_battery/adc_node.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+class ADCNode : public rclcpp::Node
+{
+public:
+  ADCNode(
+    const std::string & node_name, const rclcpp::NodeOptions & options = rclcpp::NodeOptions());
+
+protected:
+  BatteryStateMsg MergeBatteryMsgs(
+    const BatteryStateMsg & battery_msg_1, const BatteryStateMsg & battery_msg_2);
```
Kotochleb:
```suggestion
    const BatteryStateMsg & battery_msg_1, const BatteryStateMsg & battery_msg_2) const;
```

Kotochleb (changes requested):
Please consider directly specifying types like `float` and `int` directly instead of `auto`. At least in cases of primitive types.

I haven't looked into the logic of filling the battery message, since it might still change. The same applies to tests. Otherwise, style-wise, the code starts looking fine.

panther_battery/src/adc_node.cpp
KmakD: As this variables are now `std::size_t` there is no point doing static cast. I'll add const

Kotochleb (approved):

PR #140, opened 2023-07-05
Add motor enable 1.0x

Kotochleb:
bump::patch

Port motors enable service behavior to Panther 1.06 and older


panther_power_control/src/relays_node.py
```diff
@@ -134,19 +149,50 @@ def _e_stop_reset_cb(self, req: TriggerRequest) -> TriggerResponse:
...
+
     def _set_motor_state_timer_cb(self, *args) -> None:
-        motor_state = self._lines['STAGE2_INPUT'].get_value()
-        self._lines['MOTOR_ON'].set_value(motor_state)
-        if self._io_state.motor_on != motor_state:
-            self._io_state.motor_on = motor_state
+        with self._motors_lock:
+            motor_state = self._lines['STAGE2_INPUT'].get_value()
+            # if switch changes from off to on overwrite service value
+            if not self._last_motor_state and motor_state:
```
rafal-gorecki, line 184: `self._lines['STAGE2_INPUT'].get_value()` bardzo często się pojawia w kodzie proponuje stworzyć lambde `is_stage2()`
```suggestion
            # if switch changes from off to on overwrite service value
            if not self._last_motor_state and self.is_stage2():
```

panther_power_control/src/relays_node.py
```diff
@@ -134,19 +149,50 @@ def _e_stop_reset_cb(self, req: TriggerRequest) -> TriggerResponse:
...
+                self._motor_enabled = True
+
+            self._last_motor_state = motor_state
```
rafal-gorecki, line 187: Taught with C++ practice, I think that it would not be a stupid approach to use static variables in such situations. The `_last_motor_state` variable is really only needed in this function, so you can encapsulate such things so as not to clutter the whole class.

Example
```python
def my_function(self):
    if not hasattr(self.my_function, "counter"):
        self.my_function.counter = 0

    self.my_function.counter += 1
    print(f"Counter: {self.my_function.counter}")
```

panther_power_control/src/relays_node.py
```diff
@@ -134,19 +149,50 @@ def _e_stop_reset_cb(self, req: TriggerRequest) -> TriggerResponse:
...
+
+            state_to_set = motor_state and self._motor_enabled
+            self._lines['MOTOR_ON'].set_value(state_to_set)
+            self._publish_motor_state(state_to_set)
+
+    def _publish_motor_state(self, val: bool) -> None:
+        if self._io_state.motor_on != val:
+            self._io_state.motor_on = val
```
rafal-gorecki: More readable
```suggestion
    def _publish_motor_state(self, desired_state: bool) -> None:
        if self._io_state.motor_on != desired_state:
            self._io_state.motor_on = desired_state
```

panther_power_control/src/relays_node.py
Kotochleb: It appears 4 times. If we had to add lambda, we would have to do this for both pins and also port it to `power_borad_node.py` to keep the same coding style

panther_power_control/src/relays_node.py
Kotochleb: I don't see the point. This makes the code less readable. We would have to make an strong assumption the class is a singleton. Even though it would work I do not think assigning variables to a function rather than object is a good idea

rafal-gorecki (approved):

PR #145, opened 2023-07-17
ROS 1 Panther manager BehaviorTree

KmakD:
bump::patch

Rewrite Panther Manager to base it on a behavior tree for better scalability


panther/panther.repos
```diff
@@ -6,4 +6,8 @@ repositories:
   ros_components_description:
     type: git
     url: https://github.com/husarion/ros_components_description.git
-    version: ros1
\ No newline at end of file
+    version: ros1
+  BehaviorTree.CPP:
+    type: git
+    url: https://github.com/BehaviorTree/BehaviorTree.CPP.git
+    version: 4.1.0
```
Kotochleb: The behavior tree was released to rosdep so we no longer have to build it from source
https://github.com/ros/rosdistro/blob/21d33ea60e8dc1e568b00102dcf767a29b4745bf/noetic/distribution.yaml#L509C8-L509C8

Kotochleb (changes requested):

panther/panther.repos
KmakD: Done

Kotochleb (approved):

PR #165, opened 2023-08-01
use imu standalone nodes

KmakD:
bump::patch

Use IMU standalone nodes instead of nodelet

### DIFF
--- a/panther_bringup/README.md
+++ b/panther_bringup/README.md
@@ -5,11 +5,12 @@ The package containing default configuration and launch files necessary to start
 ## Default nodes launched
 
 - `battery_node` - node responsible for monitoring and publishing internal battery state of the Husarion Panther robot. For more information refer to: [panther_battery](../panther_battery/README.md).
-- `imu_manager` - nodelet managing Phidget Spatial IMU sensor.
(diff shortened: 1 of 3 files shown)


panther_bringup/README.md
```diff
@@ -5,11 +5,12 @@ The package containing default configuration and launch files necessary to start
 ## Default nodes launched
 
 - `battery_node` - node responsible for monitoring and publishing internal battery state of the Husarion Panther robot. For more information refer to: [panther_battery](../panther_battery/README.md).
-- `imu_manager` - nodelet managing Phidget Spatial IMU sensor.
+- `imu_filter_node` - node responsible for filtering and fusing raw data from IMU.
```
Kotochleb, line 8: There should be two nodes. One for `phidgets_spatial_node` and one for `imu_filter_node`

Kotochleb (changes requested):

panther_bringup/README.md
KmakD: `phidgets_spatial_node` is described later in the list to keep alphabetical order

Kotochleb: Sorry. My bad

Kotochleb (approved):

PR #175, opened 2023-08-14
add battery charging logs

pkowalsk1:
`bump::patch`

### DIFF
--- a/panther_battery/src/adc_node.py
+++ b/panther_battery/src/adc_node.py
@@ -238,8 +238,8 @@ class ADCNode:
                     rospy.logwarn_throttle(
                         5.0,
                         f'[{rospy.get_name()}] The charger has been plugged in, '
-                        f'but the charging process has not started. ',
-                        f'Verify whether the charger is connected to a power source.'
+                        f'but the charging process has not started. '
+                        f'Verify whether the charger is connected to a power source.',
                     )
             else:
                 battery_msg.power_supply_status = BatteryState.POWER_SUPPLY_STATUS_DISCHARGING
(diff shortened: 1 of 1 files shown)


panther_battery/src/adc_node.py
```diff
@@ -221,10 +221,25 @@ def _publish_battery_msg(
...
+                        f'[{rospy.get_name()}] The battery is fully charged. '
+                        f'The robot can be disconnected from the charger.',
+                    )
                 elif I_bat_mean > self._I_bat_charging_thresh[bat_pub]:
                     battery_msg.power_supply_status = BatteryState.POWER_SUPPLY_STATUS_CHARGING
+                    rospy.loginfo_throttle(
+                        1800.0,
+                        f'[{rospy.get_name()}] Robot charging process update: \n',
+                        f'    - Battery Percentage: {battery_msg.percentage * 100}%',
+                    )
```
KmakD:
```suggestion
                    rospy.loginfo_throttle(
                        1800.0,
                        f'[{rospy.get_name()}] Robot charging process update. '
                        f'Battery Percentage: {battery_msg.percentage * 100}%',
                    )
```
As I think about right now, there is no need to log this as a list, one-liner should look good. Also, comma after the first part of the string will make the code fail.

panther_battery/src/adc_node.py
```diff
@@ -221,10 +221,25 @@ def _publish_battery_msg(
...
                 else:
                     battery_msg.power_supply_status = BatteryState.POWER_SUPPLY_STATUS_NOT_CHARGING
+                    rospy.logerr_throttle(
+                        5.0,
+                        f'[{rospy.get_name()}] The charger has been plugged in, '
+                        f'but the charging process has not started. ',
+                    )
```
KmakD: I think this should be a warn, with information to check if the charger is connected to a power source.

KmakD (changes requested):

KmakD (approved):

PR #180, opened 2023-08-21
ROS 1 improve err logging for pth 1.0x

KmakD:
bump::patch

Fix publishing errors when enabling/disabling motors on Panther v1.0x

### DIFF
--- a/panther_driver/src/driver_node.py
+++ b/panther_driver/src/driver_node.py
@@ -118,6 +118,8 @@ class PantherDriverNode:
         self._motor_off_last_time = rospy.Time.now()
         self._cmd_vel_command_last_time = rospy.Time.now()
         self._cmd_vel_timeout = 0.2
+        self._motor_power_on_timeout = 2.3
+        self._motor_power_on_error_timeout = 2.5
 
         self._robot_pos = [0.0, 0.0, 0.0]  # x,  y,  yaw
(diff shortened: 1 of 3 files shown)


panther_driver/src/driver_node.py
```diff
@@ -360,6 +360,13 @@ def _main_timer_cb(self, *args) -> None:
             self._publish_tf_cb()
 
     def _driver_state_timer_cb(self, *args) -> None:
+        # wait for motor drivers to power on before publishing their state
+        if rospy.Time.now() - self._motor_off_last_time < rospy.Duration(2.3):
```
Kotochleb: Move the 2.3 to be const value

panther_driver/src/driver_node.py
```diff
@@ -415,7 +422,7 @@ def _safety_timer_cb(self, *args) -> None:
                 self._motor_off_last_time = rospy.Time.now()
             else:
                 # wait for motor drivers to power on before logging an error
-                if rospy.Time.now() - self._motor_off_last_time < rospy.Duration(2.0):
+                if rospy.Time.now() - self._motor_off_last_time < rospy.Duration(2.5):
```
Kotochleb: Move the 2.5 to be const value

Kotochleb (changes requested):
Update readme and write that the message is not published when motors are disabled

Kotochleb (approved):

PR #188, opened 2023-09-06
ROS 1 - Panther CAN class hierarchy changes

pkowalsk1:
`bump::patch`

## Description

This pull request introduces changes to the hierarchy and organization of classes used for handling communication over the CAN bus. These changes enable seamless utilization of the functionality provided by the CANopen library.

Additionally, the Network Management object has been employed to detect anomalies in communication. It monitors the last timestamp of the heartbeat frame, and if its timeout is reached, the network transitions into an error state. This triggers the activation of the emergency stop.

### DIFF
--- a/panther_driver/src/driver_node.py
+++ b/panther_driver/src/driver_node.py
@@ -15,7 +15,7 @@ from std_msgs.msg import Bool
 from std_srvs.srv import Trigger, TriggerRequest, TriggerResponse
 
 from panther_msgs.msg import DriverState, FaultFlag, IOState, RuntimeError, ScriptFlag
-from panther_can import PantherCANSDO, PantherCANPDO
+from panther_can import PantherCAN
 from panther_kinematics import PantherDifferential, PantherMecanum
 from velocity_smoother import VelocitySmoother
 
@@ -69,9 +69,6 @@ class PantherDriverNode:
 
         self._lock = Lock()
 
-        self._eds_file = rospy.get_param('~eds_file')
-        self._use_pdo = rospy.get_param('~use_pdo', False)
-        self._can_interface = rospy.get_param('~can_interface', 'panther_can')
         self._kinematics_type = rospy.get_param('~kinematics', 'differential')
         self._motor_torque_constant = rospy.get_param('~motor_torque_constant', 2.6149)
         self._gear_ratio = rospy.get_param('~gear_ratio', 30.08)
@@ -86,6 +83,10 @@ class PantherDriverNode:
         self._publish_joints = rospy.get_param('~publish_joints', True)
         self._odom_frame = rospy.get_param('~odom_frame', 'odom')
         self._base_link_frame = rospy.get_param('~base_link_frame', 'base_link')
+
+        eds_file = rospy.get_param('~eds_file')
+        use_pdo = rospy.get_param('~use_pdo', False)
+        can_interface = rospy.get_param('~can_interface', 'panther_can')
 
         robot_width = rospy.get_param('~wheel_separation', 0.697)
         robot_length = rospy.get_param('~robot_length', 0.44)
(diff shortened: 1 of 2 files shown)


panther_driver/src/driver_node.py
```diff
@@ -371,6 +366,16 @@ def _driver_state_timer_cb(self, *args) -> None:
...
+        can_net_errors = list(self._panther_can.can_connection_error())
+        if any(can_net_errors):
+            [
+                self._driver_state_msg.front.fault_flag.can_net_err,
+                self._driver_state_msg.rear.fault_flag.can_net_err,
+            ] = can_net_errors
+
+            self._driver_state_pub.publish(self._driver_state_msg)
+            return
+
```
KmakD:
```suggestion
        can_net_errors = list(self._panther_can.can_connection_error())
        [
            self._driver_state_msg.front.fault_flag.can_net_err,
            self._driver_state_msg.rear.fault_flag.can_net_err,
        ] = can_net_errors

        if any(can_net_errors):
            self._driver_state_pub.publish(self._driver_state_msg)
            return

```
This part of the code is repeated in lines 396-399. It can be moved outside if block and removed from later lines.

panther_driver/src/panther_can.py
```diff
@@ -18,327 +18,337 @@ def __setattr__(self, name, value):
         raise AttributeError(f'can\'t reassign constant {name}')
 
 
-@dataclass
 class MotorController:
-    def __init__(self, can_node_id: int, eds_file: str) -> None:
-        self._last_time_callback = {}
+    HEARTBEAT_TIMEOUT_S = 0.3
```
KmakD:
```suggestion
    HEARTBEAT_TIMEOUT = 0.3
```
I think it is generally assumed that time is specified in seconds

panther_driver/src/panther_can.py
```diff
@@ -18,327 +18,337 @@ def __setattr__(self, name, value):
...
+        with self._lock:
+            for i, wheel in enumerate(self._wheels):
+                try:
+                    self._can_node.sdo['Cmd_CANGO'][wheel].raw = vel[i]
+                except:
+                    pass
 
-        self._network.add_node(self._motor_controllers[0].can_node)
-        self._network.add_node(self._motor_controllers[1].can_node)
+    def _heartbeat_cb(self, nmt_state: int) -> None:
```
KmakD:
```suggestion
    def _heartbeat_cb(self, _) -> None:
```
We can indicate that this value wont be used inside callback

panther_driver/src/panther_can.py
```diff
@@ -18,327 +18,337 @@ def __setattr__(self, name, value):
...
-        self._robot_driver_initialized = False
-        self._can_init_time = rospy.Time.now()
+class MotorControllerPDO(MotorController):
+    def __init__(self, can_node_id: int, eds_file: str, lock: Lock) -> None:
+        super().__init__(can_node_id, eds_file, lock)
 
-        rospy.loginfo(f'[{rospy.get_name()}] Connected to the CAN bus.')
+    @property
+    def wheel_pos(self) -> List[float]:
+        return self._wheel_pos
```
KmakD: What do you think about moving this general properties to MotorController class? They can be later overwritten in MotorControllerSDO.  Properties that just returns private variables are pretty standard and I think they can be defined inside parent class
script_flags can be moved for sure as it is the same for both controller classes

panther_driver/src/panther_can.py
```diff
@@ -18,327 +18,337 @@ def __setattr__(self, name, value):
...
+        return True
 
-    def _update_wheels_current_cb(
-        self, message: canopen.pdo.Map, motor_controller: MotorController
-    ) -> None:
-        motor_controller.wheel_curr = [m.raw for m in reversed(message)]
+    def set_wheels_enc_velocity(self, velocity: List[float]) -> None:
+        vel = [velocity[:2], velocity[2:]]
+        for controller, v in zip(self._motor_controllers, vel):
+            controller.write_wheels_enc_velocity(v)
```
KmakD:
```suggestion
    def set_wheels_enc_velocity(self, velocity: List[float]) -> None:
        for controller, vel in zip(self._motor_controllers, [velocity[:2], velocity[2:]]):
            controller.write_wheels_enc_velocity(vel)
```
Also, for some reason I dont like this method being called `set_wheels_enc_velocity` - `write` made more sense. This methods could be named the same for can interface and controller but I don't know how we feel about it

KmakD (changes requested):

panther_driver/src/driver_node.py
pkowalsk1: The code repeated in these lines should be removed; that was my mistake. In my opinion, it's better to place the fault_flag.can_net_err assignment within an if statement because the for loop on line 392 reassigns it. It's not critical, but I'll sleep much better at night without it.

KmakD (approved):

PR #205, opened 2023-11-02
Ros2 gpiod

pkowalsk1:


panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_driver.hpp
```diff
@@ -1,33 +1,101 @@
...
   GPIOController();
   ~GPIOController();
 
-  void Start();
+  bool start() { return watchdog_on(); };
+  bool motors_enable(bool enable) { return set_pin_value(GPIOpins::VMOT_ON, enable); }
+  bool fan_enable(bool enable) { return set_pin_value(GPIOpins::FAN_SW, enable); }
+  bool led_enable(bool enable) { return set_pin_value(GPIOpins::LED, enable); }
+  OperationResult e_stop_trigger();
+  OperationResult e_stop_reset();
```
macstepien: I think that it would be better to use bool, and exceptions for msgs in case of failure.

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_driver.hpp
```diff
@@ -1,33 +1,101 @@
...
+
+  boost::function<void(const GPIOinfo & gpio_info)> publish_gpio_state_callback;
```
macstepien: why not std::function? Also I'm not sure if it is the best approach, I'm refactoring ros2 control driver, and if anything better comes to my mind I will let you know.

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_driver.hpp
```diff
@@ -1,33 +1,101 @@
...
+    {GPIOpins::MOTOR_DRIVER_EN, GPIOinfo{"GPIO23", gpiod::line::direction::OUTPUT}},
+    {GPIOpins::VMOT_ON, GPIOinfo{"GPIO6", gpiod::line::direction::OUTPUT}},
+    {GPIOpins::FAN_SW, GPIOinfo{"GPIO5", gpiod::line::direction::OUTPUT}},
+    {GPIOpins::TEST_IN, GPIOinfo{"SDA1", gpiod::line::direction::INPUT}},
+  };
+
+  std::unique_ptr<gpiod::line_request> watchdog_request_;
+  std::unique_ptr<gpiod::line_request> controll_request_;
+  std::unique_ptr<std::thread> gpio_monitor_thread_;
   std::unique_ptr<std::thread> watchdog_thread_;
```
macstepien, line 83: It would be good to add priority to this threads

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_driver.hpp
```diff
@@ -1,33 +1,101 @@
...
+  bool watchdog_on();
+  bool watchdog_off();
+  void watchdog_thread();
+  bool is_watchdog_thread_running() const;
+  void gpio_monitor_on();
+  void gpio_monitor_off();
+  void monitor_async_events();
+  void handle_edge_event(const gpiod::edge_event & event);
 
+  std::map<GPIOpins, GPIOinfo> gpio_info_{
```
macstepien: It would be better not to use map - we are targeting RT, and it isn't really safe to use maps. We can use vector instead.

panther_hardware_interfaces/src/gpio_driver.cpp
```diff
@@ -5,56 +5,282 @@ namespace panther_hardware_interfaces
...
+
+  std::vector<GPIOpins> controll_pins;
+  for (GPIOpins pin = GPIOpins::WATCHDOG + 1; pin < GPIOpins::UNKNOWN; pin = pin + 1) {
+    controll_pins.push_back(pin);
+  }
+
+  controll_request_ = create_line_request(gpio_chip, controll_pins);
+  watchdog_request_ = create_line_request(gpio_chip, GPIOpins::WATCHDOG);
+
+  gpio_monitor_on();
```
macstepien: I think that it would make sense to split it into initialize and activate (activate - `gpio_monitor_on`)

panther_hardware_interfaces/src/gpio_driver.cpp
```diff
@@ -5,56 +5,282 @@ namespace panther_hardware_interfaces
...
 }
 
 GPIOController::~GPIOController()
 {
-  watchdog_thread_->~thread();
+  watchdog_off();
+  gpio_monitor_off();
+  watchdog_request_->release();
+  controll_request_->release();
```
macstepien, line 35: here also we could split it into deinitialize and deactivate

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_driver.hpp
```diff
@@ -1,33 +1,101 @@
...
+};
 
-enum class GPIOPins { VMOT_ON = 6, MOTOR_DRIVER_EN = 23, WATCHDOG = 14, E_STOP_RESET = 27 };
+struct OperationResult
+{
+  bool success;
+  std::string message = "";
+};
 
 class GPIOController
```
macstepien: I think that it may make sense to split it into two classes - one more generic and lower level (set_pin_value, get_pin_value, etc) and one more specific for panther with watchdog thread, etc.

panther_hardware_interfaces/src/gpio_driver.cpp
```diff
@@ -5,56 +5,282 @@ namespace panther_hardware_interfaces
...
+  auto line_config = gpiod::line_config();
+  line_config.add_line_settings(pin_info.offset, settings);
+
+  controll_request_->reconfigure_lines(line_config);
+}
+
+bool GPIOController::get_pin_value(GPIOpins pin)
+{
+  return gpio_info_[pin].value == gpiod::line::value::ACTIVE;
+}
```
macstepien: maybe is_pin_active?

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_driver.hpp
```diff
@@ -1,33 +1,101 @@
...
+  E_STOP_RESET,
+  MOTOR_DRIVER_EN,
+  VMOT_ON,
+  FAN_SW,
+  LED,
+  TEST_IN,
+  UNKNOWN
+};
+
+GPIOpins operator+(const GPIOpins & pin, int int_val);
```
macstepien: probably won't be necessary if we switch from map to vector

macstepien (changes requested):

panther_hardware_interfaces/src/gpio_driver.cpp
```diff
@@ -5,56 +5,282 @@ namespace panther_hardware_interfaces
...
+  pollfd.fd = controll_request_->fd();
+  pollfd.events = POLLIN;
+
+  while (gpio_monitor_thread_enabled_) {
+    auto ret = poll(&pollfd, 1, -1);
+
+    if (ret == -1) {
+      std::cerr << "Error waiting for edge events" << std::endl;
+      return;
+    }
```
macstepien, line 189: here we can set some flag, so that it will be possible to check if error occurred

macstepien (changes requested):

PR #206, opened 2023-11-06
ROS 2 add Roboteq Battery

KmakD:


panther_battery/README.md
```diff
@@ -1,48 +1,68 @@
+[//]: # (ROS_API_PACKAGE_START)
+[//]: # (ROS_API_PACKAGE_NAME_START)
+
 # panther_battery
 
+[//]: # (ROS_API_PACKAGE_NAME_END)
+[//]: # (ROS_API_PACKAGE_DESCRIPTION_START)
+
 Package containing nodes monitoring and publishing internal battery state of the Husarion Panther robot.
```
Kotochleb:
```suggestion
Package containing nodes monitoring and publishing the internal battery state of the Husarion Panther robot.
```

panther_battery/README.md
```diff
@@ -1,48 +1,68 @@
...
+[//]: # (ROS_API_NODE_SUBSCRIBERS_END)
 
-### roboteq_republisher_node
+#### Parameters
 
-Node publishing Panther battery state read from motor controllers. Used in Panther versions 1.06 and below.
+[//]: # (ROS_API_NODE_PARAMETERS_START)
 
-#### Publishes
+- `~battery_timeout` [*float*, default: **1.0**]: specifies the timeout in seconds. If node fails to read battery data exceeding this duration, the node will publish an unknown battery state.
```
Kotochleb:
```suggestion
- `~battery_timeout` [*float*, default: **1.0**]: specifies the timeout in seconds. If the node fails to read battery data exceeding this duration, the node will publish an unknown battery state.
```

panther_battery/README.md
```diff
@@ -1,48 +1,68 @@
...
+[//]: # (ROS_API_NODE_COMPATIBLE_1_2)
+[//]: # (ROS_API_NODE_NAME_START)
+
+### battery_node
 
-Publishes battery state read from ADC unit. Available from Panther version 1.2.
+[//]: # (ROS_API_NODE_NAME_END)
+[//]: # (ROS_API_NODE_DESCRIPTION_START)
+
+Publishes battery state read from ADC unit for Panther version 1.2 and above, or based on Roboteq motor controllers data for earlier versions of the robot.
```
Kotochleb:
```suggestion
Publishes battery state read from ADC unit for Panther version 1.2 and above, or based on Roboteq motor controllers' data for earlier versions of the robot.
```

panther_battery/README.md
```diff
@@ -1,48 +1,68 @@
...
+- `~adc0_device` [*string*, default: **/sys/bus/iio/devices/iio:device0**]: ADC nr 0 device IIO path.
+- `~adc1_device` [*string*, default: **/sys/bus/iio/devices/iio:device1**]: ADC nr 1 device IIO path.
+- `~ma_window_len/charge` [*int*, default: **10**]: window length of a moving average, used to smooth out battery charge readings.
+- `~ma_window_len/temp` [*int*, default: **10**]: window length of a moving average, used to smooth out battery temperature readings.
 
-- `/panther/driver/motor_controllers_state` [*panther_msgs/DriverState*]: current motor controllers' state and error flags.
+**Roboteq Battery Additional Parameters**
 
-#### Parameters
+- `~driver_state_timeout` [*float*, default: **0.2**]: specifies timeout in seconds after which driver state messages will be considered old.
```
Kotochleb: Adding a namespace for ROS parameters would simplify this section of documentation and make them more intuitive.

```
- `~adc/device_0` [*string*, default: **/sys/bus/iio/devices/iio:device0**]: ADC nr 0 device IIO path.
- `~adc/device_1` [*string*, default: **/sys/bus/iio/devices/iio:device1**]: ADC nr 1 device IIO path.
- `~adc/ma_window_len/charge` [*int*, default: **10**]: window length of a moving average, used to smooth out battery charge readings.
- `~adc/ma_window_len/temp` [*int*, default: **10**]: window length of a moving average, used to smooth out battery temperature readings.
- `~roboteq/driver_state_timeout` [*float*, default: **0.2**]: specifies timeout in seconds after which driver state messages will be considered old.
```

panther_battery/src/battery_node.cpp
```diff
@@ -26,31 +28,52 @@
...
+    std::chrono::milliseconds(100), std::bind(&BatteryNode::BatteryPubTimerCB, this));
 
   RCLCPP_INFO(this->get_logger(), "Node started");
 }
 
-void ADCNode::Initialize()
+void BatteryNode::Initialize()
 {
+  auto panther_version = this->get_parameter("panther_version").as_double();
+  if (panther_version < 1.2 - std::numeric_limits<float>::epsilon()) {
```
Kotochleb:
```suggestion
```suggestion
  const float panther_version = this->get_parameter("panther_version").as_double();
  if (panther_version < 1.2f - std::numeric_limits<float>::epsilon()) {
```
This way, the double will be automatically cast to float and the computation will not make any pointless recasting.

One question though, why do you subtract the epsilon?

panther_battery/src/battery_node.cpp
```diff
@@ -26,31 +28,52 @@
...
+    try {
+      InitializeWithADCBattery();
+    } catch (std::runtime_error & err) {
+      RCLCPP_WARN(this->get_logger(), "Failed to initialize ADC Battery: %s", err.what());
+      RCLCPP_INFO(this->get_logger(), "Using Roboteq drivers to publish battery data.");
+      InitializeWithRoboteqBattery();
+    }
+  }
+  RCLCPP_INFO(this->get_logger(), "Battery publishers initialized");
+}
```
Kotochleb: I last log will be moved to the timer callback the function can be rewritten to call `InitializeWithRoboteqBattery()` only once as a form of fallback when ADC is not available

panther_battery/src/battery_node.cpp
```diff
@@ -26,31 +28,52 @@
...
+
+void BatteryNode::InitializeWithADCBattery()
+{
+  RCLCPP_INFO(this->get_logger(), "Initializing battery node using ADC data");
+
+  this->declare_parameter<std::string>("adc0_device", "/home/ros/ros2_ws/src/device0");
+  this->declare_parameter<std::string>("adc1_device", "/home/ros/ros2_ws/src/device1");
```
Kotochleb: This path is oddly specific...
Not only that, it does not match the documentation which I believe is correct

panther_battery/include/panther_battery/roboteq_battery.hpp
```diff
@@ -0,0 +1,76 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef PANTHER_BATTERY_ROBOTEQ_BATTERY_HPP_
+#define PANTHER_BATTERY_ROBOTEQ_BATTERY_HPP_
+
+#include <cstdint>
```
Kotochleb, line 18: Unnecessary import

panther_battery/src/roboteq_battery.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+  voltage_ma_ = std::make_unique<panther_utils::MovingAverage<float>>(
+    params.voltage_window_len, std::numeric_limits<float>::quiet_NaN());
+  current_ma_ = std::make_unique<panther_utils::MovingAverage<float>>(
+    params.current_window_len, std::numeric_limits<float>::quiet_NaN());
+}
+
+bool RoboteqBattery::Present() { return true; }
+
+void RoboteqBattery::Update(const rclcpp::Time & header_stamp, const bool)
```
Kotochleb:
```suggestion
void Update(const rclcpp::Time & header_stamp, const bool /* charger_connected */)
```
This shows what was the original use for the boolean variable and not cause unused error

panther_battery/src/roboteq_battery.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  if (driver_state_->front.fault_flag.can_net_err || driver_state_->rear.fault_flag.can_net_err) {
+    throw std::runtime_error("Motor controller CAN network error");
+  }
+}
+
+void RoboteqBattery::UpdateBatteryMsgs(const rclcpp::Time & header_stamp)
+{
+  UpdateBatteryState(header_stamp);
+  UpdateBatteryStateRaw();
```
Kotochleb, line 85: I am not entirely convinced about the readability of those functions, but I will not strongly opt to change those

panther_battery/test/test_battery_node_roboteq.cpp
```diff
@@ -0,0 +1,124 @@
...
+  driver_state_pub_->publish(driver_state);
+
+  ASSERT_TRUE(panther_utils::test_utils::WaitForMsg(
+    battery_node_, battery_state_, std::chrono::milliseconds(1000)));
+
+  // battery state msg should have some values
+  EXPECT_FALSE(std::isnan(battery_state_->voltage));
+  EXPECT_FALSE(std::isnan(battery_state_->current));
+  EXPECT_FALSE(std::isnan(battery_state_->charge));
+  EXPECT_FALSE(std::isnan(battery_state_->percentage));
```
Kotochleb, line 99: Also at this stage the state should not be unknown I think

panther_battery/test/test_roboteq_battery.cpp
```diff
@@ -0,0 +1,258 @@
...
+    BatteryStateMsg::POWER_SUPPLY_STATUS_UNKNOWN, BatteryStateMsg::POWER_SUPPLY_HEALTH_UNKNOWN);
+}
+
+TEST_F(TestRoboteqBattery, BatteryMsgValues)
+{
+  const auto V_bat_full = 41.4;
+  const auto V_bat_min = 32.0;
+
+  const float voltage_1 = 35.0;
+  const float current_1 = 0.1;
```
Kotochleb:
```suggestion
  const float V_bat_full = 41.4;
  const float V_bat_min = 32.0;

  const float voltage_1 = 35.0;
  const float current_1 = 0.1;
```
Why one was auto and one was float?

Kotochleb (changes requested):
Some comments start with upper case and some with lower case. I think starting all comments that start at the beginning of a sentence with an upper case letter would be a good idea

panther_battery/src/battery_node.cpp
KmakD: To ensure condition is not fulfilled when panthet_version equals precisely 1.2

panther_battery/include/panther_battery/roboteq_battery.hpp
KmakD: `uint8_t` is used in line 61

panther_battery/test/test_battery_node_roboteq.cpp
KmakD: At this stage the battery state is not unknown and these tests checks exactly that.

panther_battery/src/battery_node.cpp
Kotochleb: Adding brackets surrounding the subtraction will make the order more clear in this case

Kotochleb (changes requested):
Comments in lines 64-66 in `adc_battery.hpp` could be better formatted. Other comments are also not beginning with upper case letters.

Kotochleb (approved):

PR #207, opened 2023-11-08
Add pre-commit, clang-format and license to files

Kotochleb:
Adds pre-commit configuration to the repository


.pre-commit-config.yaml
```diff
@@ -0,0 +1,93 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.5.0
+    hooks:
+      - id: check-added-large-files
```
macstepien, line 5: Does it work with defaults? I checked and max size should be 500kB, and our largest mesh has 1.4MB.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - id: check-xml
+      - id: check-yaml
+      - id: debug-statements
+      - id: destroyed-symlinks
+      - id: detect-private-key
+      - id: end-of-file-fixer
+      - id: fix-byte-order-marker
+      - id: name-tests-test
+      - id: mixed-line-ending
+      - id: sort-simple-yaml
```
macstepien: I'm not sure about `sort-simple-yaml`

panther_battery/launch/battery.launch.py
```diff
@@ -6,21 +22,21 @@
 
 
 def generate_launch_description():
-    panther_version = LaunchConfiguration('panther_version')
-    declare_panther_version_arg = DeclareLaunchArgument('panther_version')
+    panther_version = LaunchConfiguration("panther_version")
```
macstepien, line 25: I thought that we were supposed to change this setting from black and use single quotes (I prefer double though).

.pre-commit-config.yaml
Kotochleb: I made it to 2 MB

.pre-commit-config.yaml
Kotochleb: I removed it

panther_battery/launch/battery.launch.py
Kotochleb: As we discussed in ROS 2 we will used double quotes

macstepien (approved):

PR #210, opened 2023-11-17
ROS 2 panther lights

KmakD:
Parts with usage of the libgpiod will change when separate class for GPIO handling is introduced. Don't pay too much attention to this parts of the code.
I have no idea why `panther_battery/package.xml` is shown as it was modified - it didn't change.


panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+void APA102::SetPanel(const std::vector<std::uint8_t> & frame) const
+{
+  std::uint8_t * buffer;
+  auto buffer_size = RGBAFrameToBGRBuffer(frame, buffer);
+
+  struct spi_ioc_transfer tr;
+  memset(&tr, 0, sizeof(tr));
+  tr.tx_buf = (unsigned long long)buffer;
```
Kotochleb:
```suggestion
  tr.tx_buf = static_cast<uint64_t>(buffer);
```
Even though most compilers define `unsigned long long` as a 64 bits, the standard says it has to be "at least" 64 bits. In most cases this would be fine, but here I would opt for clear standard int, which precisely defines the size since it makes it clear, plus it matches the definition from the structure defined in the system

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,124 @@
...
+  tr.rx_buf = 0;
+  tr.len = (unsigned int)buffer_size;
```
Kotochleb:
```suggestion
  tr.len = static_cast<uint32_t>(buffer_size);
```
Same as above

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,124 @@
...
+  int ret = ioctl(fd_, SPI_IOC_MESSAGE(1), &tr);
+  delete[] buffer;
+
+  if (ret < 1) {
+    throw std::ios_base::failure("Failed to send data over SPI " + device_);
+  }
+}
+
+std::size_t APA102::RGBAFrameToBGRBuffer(
+  const std::vector<std::uint8_t> & frame, std::uint8_t *& buffer) const
```
Kotochleb:
```suggestion
  const std::vector<std::uint8_t> & frame, std::uint8_t buffer[]) const
```
Passing by reference has the same overhead as passing with a pointer. Additionally, the array can be passed like this

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,124 @@
...
+    std::uint8_t brightness = (std::uint16_t(frame[padding + 3]) * global_brightness_) / 255;
+    buffer[4 + padding] = 0xE0 | brightness;
+    // Convert rgb to bgr with collor correction
+    buffer[4 + padding + 1] = std::uint8_t((std::uint16_t(frame[padding + 2]) * kCorrBlue) / 255);
+    buffer[4 + padding + 2] = std::uint8_t((std::uint16_t(frame[padding + 1]) * kCorrGreen) / 255);
+    buffer[4 + padding + 3] = std::uint8_t((std::uint16_t(frame[padding + 0]) * kCorrRed) / 255);
+  }
+
+  return buffer_size;
+}
```
Kotochleb: When separating functions responsible for setting the panel and populating the frame I think it is better to add third, control function that will first allocate the memory and invoke `RGBAFrameToBGRBuffer` and later pass raw data pointer, or a `std::vector` to the `SetPanel` that will be directly sent to the panels

Kotochleb: Buffers might be implemented with `std::vector`. This will simplify the logic, but requires checking if it works. In case it will, feel free to discard other comments that will interfeere with this change

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,124 @@
...
+  struct spi_ioc_transfer tr;
+  memset(&tr, 0, sizeof(tr));
+  tr.tx_buf = (unsigned long long)buffer;
+  tr.rx_buf = 0;
+  tr.len = (unsigned int)buffer_size;
+  tr.speed_hz = speed_;
+  tr.delay_usecs = 0;
+  tr.bits_per_word = kBits;
+
+  int ret = ioctl(fd_, SPI_IOC_MESSAGE(1), &tr);
```
Kotochleb:
```suggestion
  const int ret = ioctl(fd_, SPI_IOC_MESSAGE(1), &tr);
```

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  // Init start and end frames
+  for (std::size_t i = 0; i < 4; i++) {
+    buffer[i] = 0x00;
+    buffer[buffer_size - i - 1] = 0xFF;
+  }
+
+  // Copy frame from vector to sending buffer
+  for (std::size_t i = 0; i < frame.size() / 4; i++) {
+    std::size_t padding = i * 4;
```
Kotochleb:
```suggestion
    const std::size_t padding = i * 4;
```

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,124 @@
...
+    // Header with brightness
+    std::uint8_t brightness = (std::uint16_t(frame[padding + 3]) * global_brightness_) / 255;
```
Kotochleb:
```suggestion
    const std::uint8_t brightness = (std::uint16_t(frame[padding + 3]) * global_brightness_) / 255;
```

panther_lights/src/driver_node.cpp
```diff
@@ -0,0 +1,176 @@
...
+  this->declare_parameter<int>("num_led", 46);
+
+  RCLCPP_INFO(this->get_logger(), "Node started");
+}
+
+void DriverNode::Initialize()
+{
+  it_ = std::make_shared<image_transport::ImageTransport>(this->shared_from_this());
+
+  const double global_brightness = this->get_parameter("global_brightness").as_double();
```
Kotochleb:
```suggestion
  const flaot global_brightness = static_cast<float>(this->get_parameter("global_brightness").as_double());
```
The service uses float so passing a double does not make much sense

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  if (ioctl(fd_, SPI_IOC_WR_MAX_SPEED_HZ, &speed_) == -1) {
+    close(fd_);
+    throw std::ios_base::failure("Can't set speed for " + device_);
+  }
+}
+
+APA102::~APA102() { close(fd_); }
+
+void APA102::SetGlobalBrightness(const double brightness)
```
Kotochleb:
```suggestion
void APA102::SetGlobalBrightness(const float brightness)
```
`panther_msgs/srvs/SetLEDBrightness.data` if float so I don't see a point of using double in here

panther_lights/src/driver_node.cpp
```diff
@@ -0,0 +1,176 @@
...
+  request.release();
+}
+
+void DriverNode::SetBrightnessCB(
+  const SetLEDBrightnessSrv::Request::SharedPtr & req, SetLEDBrightnessSrv::Response::SharedPtr res)
+{
+  const float brightness = req->data;
+  if (
+    brightness < 0.0f - std::numeric_limits<float>::epsilon() ||
+    brightness > 1.0f + std::numeric_limits<float>::epsilon()) {
```
Kotochleb: I think that epsilon is a bit of an overshoot in this case...

panther_lights/include/panther_lights/apa102.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+class APA102
+{
+public:
+  APA102(
+    const std::string & device, const std::uint32_t speed = 800000, const bool cs_high = false);
+  ~APA102();
+
+  void SetGlobalBrightness(const std::uint8_t brightness);
+  void SetGlobalBrightness(const double brightness);
```
Kotochleb:
```suggestion
  void SetGlobalBrightness(const float brightness);
```
As stated everywhere else, `panther_msgs/srvs/SetLEDBrightness.data` is `float`

Kotochleb (changes requested):
Still missing tests checking whether the frame is correctly parsed

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,126 @@
...
+    close(fd_);
+    throw std::ios_base::failure("Can't set speed for " + device_);
+  }
+}
+
+APA102::~APA102() { close(fd_); }
+
+void APA102::SetGlobalBrightness(const float brightness)
+{
+  std::uint8_t val = brightness > 0.0f ? ceil(brightness * 31.0f) : 0;
```
Kotochleb:
```suggestion
  const std::uint8_t val = brightness > 0.0f ? ceil(brightness * 31.0f) : 0;
```

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,126 @@
...
+  SetGlobalBrightness(val);
+}
+
+void APA102::SetGlobalBrightness(const std::uint8_t brightness)
+{
+  // Clamp values to be at max 31
+  global_brightness_ = std::uint16_t(brightness) & 0x1F;
```
Kotochleb:
```suggestion
  global_brightness_ = static_cast<std::uint16_t>(brightness) & 0x1F;
```
This suggestion can be omitted. Generally, either we keep it as it is or we do static casts. This can be omitted since in `RGBAFrameToBGRBuffer` this would generate loads of pointless lines of code

panther_lights/include/panther_lights/apa102.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+class APA102
+{
+public:
+  APA102(
+    const std::string & device, const std::uint32_t speed = 800000, const bool cs_high = false);
+  ~APA102();
+
+  void SetGlobalBrightness(const std::uint8_t brightness);
+  void SetGlobalBrightness(const float brightness);
```
Kotochleb: Missing description for those function

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,126 @@
...
+  }
+}
+
+APA102::~APA102() { close(fd_); }
+
+void APA102::SetGlobalBrightness(const float brightness)
+{
+  std::uint8_t val = brightness > 0.0f ? ceil(brightness * 31.0f) : 0;
+  SetGlobalBrightness(val);
+}
```
Kotochleb: Depending on how strong is our choice about not throwing exceptions, this function might in fact throw exception if values are out of scope. Or the function with `std::uint8_t` input might do so. This way there is no need to check size of range of brightness in the service callback

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+std::vector<std::uint8_t> APA102::RGBAFrameToBGRBuffer(
+  const std::vector<std::uint8_t> & frame) const
+{
+  if (frame.size() % 4 != 0) {
+    throw std::runtime_error("Incorrect number of bytes to convert frame");
+  }
+
+  std::size_t buffer_size = (4 * sizeof(std::uint8_t)) + frame.size() + (4 * sizeof(std::uint8_t));
+  std::vector<std::uint8_t> buffer(buffer_size);
```
Kotochleb:
```suggestion
  // Create buffer with length of RGBA frame with leading and tailing 4 bytes
  std::vector<std::uint8_t> buffer(2 * (4 * sizeof(std::uint8_t)) + frame.size());
```

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+  // Init start and end frames
+  for (std::size_t i = 0; i < 4; i++) {
+    buffer[i] = 0x00;
+    buffer[buffer_size - i - 1] = 0xFF;
+  }
```
Kotochleb:
```suggestion
  // Init start and end frames
  std::fill(buffer.begin(), buffer.begin() + 4, 0x00);
  std::fill(buffer.end() - 4, buffer.end(), 0xFF);
```

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+void APA102::SetGlobalBrightness(const std::uint8_t brightness)
+{
+  // Clamp values to be at max 31
+  global_brightness_ = std::uint16_t(brightness) & 0x1F;
+}
+
+void APA102::SetPanel(const std::vector<std::uint8_t> & frame) const
+{
+  std::vector<std::uint8_t> buffer = RGBAFrameToBGRBuffer(frame);
```
Kotochleb:
```suggestion
  const auto buffer = RGBAFrameToBGRBuffer(frame);
```

panther_lights/src/driver_node.cpp
```diff
@@ -0,0 +1,174 @@
...
+  rclcpp::on_shutdown(std::bind(&DriverNode::OnShutdown, this));
+
+  this->declare_parameter<double>("global_brightness", 1.0);
+  this->declare_parameter<double>("frame_timeout", 0.1);
+  this->declare_parameter<int>("num_led", 46);
+
+  RCLCPP_INFO(this->get_logger(), "Node started");
+}
+
+void DriverNode::Initialize()
```
Kotochleb, line 52: Is there any particular reason why the initialize function is being used and not to do all of this work in the constructor?

Kotochleb (changes requested):

panther_lights/src/apa102.cpp
KmakD: It make sense to throw exceptions and check them inside service callback. I'll change that

panther_lights/src/apa102.cpp
KmakD: According to our latest course:
If a comment is intended to explain a complex expression, the expression should be split into understandable subexpressions using Extract Variable

panther_lights/src/driver_node.cpp
KmakD: Creating ImageTransport requires `shared_from_this` method which can't be used inside constructor

panther_lights/src/apa102.cpp
KmakD: We leave it as it according to discussion

panther_lights/src/apa102.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+std::vector<std::uint8_t> APA102::RGBAFrameToBGRBuffer(
+  const std::vector<std::uint8_t> & frame) const
+{
+  if (frame.size() % 4 != 0) {
+    throw std::runtime_error("Incorrect number of bytes to convert frame");
+  }
+
+  std::size_t buffer_size = (4 * sizeof(std::uint8_t)) + frame.size() + (4 * sizeof(std::uint8_t));
```
Kotochleb:
```suggestion
  const std::size_t buffer_size = (4 * sizeof(std::uint8_t)) + frame.size() + (4 * sizeof(std::uint8_t));
```

Kotochleb (changes requested):

Kotochleb (approved):

PR #212, opened 2023-11-23
ROS 2 imu node

KmakD:


panther_bringup/config/imu_config.yaml
```diff
@@ -0,0 +1,44 @@
+imu_filter_node:
```
pkowalsk1: `imu_config.yaml` or `imu.yaml`? Considering other config in this pkg and other Husarion robots utilize `ekf.yaml`, etc., which one would you prefer?

panther_bringup/launch/imu.launch.py
```diff
@@ -0,0 +1,57 @@
...
+def generate_launch_description():
+    imu_config_file = LaunchConfiguration("imu_config_file")
+    declare_imu_config_file_arg = DeclareLaunchArgument(
+        "imu_config_file",
+        description="Path to IMU configuration file",
+    )
+
+    imu_container = ComposableNodeContainer(
+        name="imu_container",
+        namespace="",
```
pkowalsk1: Is it necessary? Won't they conflict if we add a namespace in `bringup.launch` later?

panther_bringup/config/imu_config.yaml
KmakD: Good question. in ROS 1 all configs are named with `_config`. I assume we switch to not using it.

pkowalsk1 (approved):

PR #213, opened 2023-12-01
ROS 2 lights gpio handle

KmakD:


panther_lights/src/driver_node.cpp
```diff
@@ -57,6 +58,9 @@ void DriverNode::Initialize()
   frame_timeout_ = this->get_parameter("frame_timeout").as_double();
   num_led_ = this->get_parameter("num_led").as_int();
 
+  std::vector<panther_gpiod::GPIOInfo> gpio_info_storage = {panther_gpiod::GPIOInfo{
+    panther_gpiod::GPIOPin::LED_SBC_SEL, gpiod::line::direction::OUTPUT, true}};
+  gpio_driver_ = std::make_unique<panther_gpiod::GPIODriver>(gpio_info_storage);
   SetPowerPin(false);
```
pkowalsk1: Is it still necessary? The GPIODriver constructor establishes the initial pin value, which in this case is set to INACTIVE.

panther_lights/src/driver_node.cpp
```diff
@@ -134,22 +138,7 @@ void DriverNode::FrameCB(
 
 void DriverNode::SetPowerPin(const bool value) const
```
pkowalsk1: Does this method remain const?

KmakD: The code compiled but I guess it make more sense to make int non const in this case

pkowalsk1 (approved):

PR #217, opened 2023-12-18
Add welcome msg

KmakD:
The output will look like this:

Screenshot from 2023-12-18 11-12-38

### DIFF
--- a/panther_bringup/launch/bringup.launch.py
+++ b/panther_bringup/launch/bringup.launch.py
@@ -14,12 +14,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import textwrap
+
+import click
 from ament_index_python.packages import get_package_share_directory
 from launch import LaunchDescription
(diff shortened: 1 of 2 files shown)


panther_bringup/launch/bringup.launch.py
```diff
@@ -14,12 +14,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import textwrap
+
+import click
```
pkowalsk1, line 19: Is there a reason why this isn't alphabetically ordered?

panther_bringup/package.xml
```diff
@@ -22,6 +22,9 @@
   <exec_depend>phidgets_spatial</exec_depend>
   <exec_depend>robot_localization</exec_depend>
 
+  <!-- Python dependencies -->
+  <depend>python3-click</depend>
```
pkowalsk1: maybe just exec_depend?

pkowalsk1 (approved):

PR #218, opened 2024-01-16
Manuall merge of ros2-prealpha to ros2-dev

KmakD:
Manual merge of prealpha branch. Changes regarding `ros2_control` and `panther_hardware_interfaces` were omitted.


panther_gpiod/include/panther_gpiod/gpio_driver.hpp
```diff
@@ -245,9 +254,10 @@ class GPIODriver
     {GPIOPin::GPIN2, "GPIN2"},
     {GPIOPin::LED_SBC_SEL, "LED_SBC_SEL"},
     {GPIOPin::SHDN_INIT, "SHDN_INIT"},
+    {GPIOPin::STAGE2_INPUT, "STAGE2_INPUT"},
     {GPIOPin::VDIG_OFF, "VDIG_OFF"},
     {GPIOPin::VMOT_ON, "VMOT_ON"},
-  };
+    {GPIOPin::MOTOR_ON, "MOTOR_ON"}};
```
macstepien: is it formatted correctly? shouldn't `};` be in the next line?

panther_gpiod/src/gpio_driver.cpp
```diff
@@ -105,9 +105,18 @@ gpiod::line_settings GPIODriver::GenerateLineSettings(const GPIOInfo & gpio_info
 {
   auto settings = gpiod::line_settings();
   settings.set_direction(gpio_info.direction);
-  settings.set_output_value(gpio_info.init_value);
   settings.set_active_low(gpio_info.active_low);
 
+  // Set the initial value only when the line is configured for the first time;
+  // otherwise, set the last known value
+  gpiod::line::value new_outpu_value;
```
macstepien:
```suggestion
  gpiod::line::value new_output_value;
```

panther_gpiod/src/gpio_driver.cpp
```diff
@@ -217,7 +236,7 @@ void GPIODriver::MonitorAsyncEvents()
   pollfd.events = POLLIN;
 
   {
-    std::lock_guard<std::mutex> lck(monitor_init_mtx_);
+    std::unique_lock<std::mutex> lck(monitor_init_mtx_);
```
macstepien: I think that lock_guard is better in this case.

panther_gpiod/src/gpio_driver.cpp
```diff
@@ -192,11 +211,11 @@ void GPIODriver::GPIOMonitorOn()
     }
   }
 
+  std::unique_lock<std::mutex> lck(monitor_init_mtx_);
```
macstepien, line 207: also here, and above (I can't add comment there), lock_guard will be enough

macstepien (changes requested):

panther_gpiod/include/panther_gpiod/gpio_driver.hpp
KmakD: Depends in there is `,` after last element. I copied it as it was in prealpha but it make sense to change that

panther_gpiod/src/gpio_driver.cpp
KmakD: I don't know the reason this was changed for prealpha. @pidpawel ?

pkowalsk1: something went no yes with wild merging. In this simple case lock_guard should be a better fit.

panther_bringup/launch/bringup.launch.py
```diff
@@ -14,20 +14,69 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import textwrap
+
+import click
```
pkowalsk1, line 19: is it pre-commit?

panther_gpiod/src/gpio_driver.cpp
```diff
@@ -105,9 +105,18 @@ gpiod::line_settings GPIODriver::GenerateLineSettings(const GPIOInfo & gpio_info
...
 
+  // Set the initial value only when the line is configured for the first time;
+  // otherwise, set the last known value
+  gpiod::line::value new_output_value;
+  if (!line_request_) {
+    new_output_value = gpio_info.init_value;
+  } else {
+    new_output_value = gpio_info.value;
+  }
+  settings.set_output_value(new_output_value);
```
pkowalsk1:
```suggestion
  gpiod::line::value new_output_value = line_request_ ? gpio_info.value : gpio_info.init_value;
  settings.set_output_value(new_output_value);
```
just a suggestion

panther_gpiod/src/gpio_driver.cpp
```diff
@@ -138,6 +147,16 @@ void GPIODriver::ChangePinDirection(const GPIOPin pin, const gpiod::line::direct
...
+bool GPIODriver::IsPinAvaible(const GPIOPin pin) const
+{
+  for (auto & info : gpio_info_storage_) {
+    if (info.pin == pin) {
+      return true;
+    }
+  }
+  return false;
+}
+
```
pkowalsk1:
```suggestion
bool GPIODriver::IsPinAvaible(const GPIOPin pin) const
{
  return std::any_of(gpio_info_storage_.begin(), gpio_info_storage_.end(),
                     [&](const auto &info) { return info.pin == pin; });
}

```
It is considered more idiomatic in modern C++. Just another suggestion :p 

With such changes, we still need to include <algorithm>.

panther_bringup/launch/bringup.launch.py
KmakD: yes

panther_gpiod/src/gpio_driver.cpp
KmakD: I think all locks in file can be made `lock_guard` correct me if I'm wrong

macstepien (approved):

panther_gpiod/src/gpio_driver.cpp
pkowalsk1: I'm sorry @KmakD. There is a typo in my comment caused by parsing `<>`. With such changes, we still need to include `<algorithm>`.

pkowalsk1 (approved):

PR #219, opened 2024-01-17
ros2_control PDO commands

macstepien:
New capabilities were introduced in fw 2.1a for Roboteq, which allowed us, among others to use PDO commands, and move some other information also to PDO. Additional required changes to `panther_msgs`: https://github.com/husarion/panther_msgs/pull/18


panther_description/urdf/panther_macro.urdf.xacro
```diff
@@ -97,19 +97,22 @@
           <param name="master_can_id">3</param>
           <param name="front_driver_can_id">1</param>
           <param name="rear_driver_can_id">2</param>
-          <!-- Shouldn't but can depend on frequencies of publishing PDO (set on Roboteq) and
-            frequency of controller (frequency of publishing commands) -->
-          <param name="sdo_operation_timeout_ms">3</param>
-          <!-- Depends on frequency of the controller -->
-          <param name="pdo_feedback_timeout_ms">30</param>
+          <param name="sdo_operation_timeout_ms">100</param>
+          <!-- Depends on frequency of the controller, more critical motors state are sent with higher frequency, other parameters are sent with lower, here allowed time is set to be expected period +50% margin -->
```
KmakD: Its a very long comment, can you split it into at least 2 lines?

panther_hardware_interfaces/src/roboteq_driver.cpp
```diff
@@ -100,110 +98,81 @@ bool RoboteqDriver::WaitForBoot()
...
+      rpdo_mapped[RoboteqCANObjects::flags.id][RoboteqCANObjects::flags.subid]),
+    1);
+  state.runtime_stat_flag_motor_2 = GetByte(
+    static_cast<std::int32_t>(
+      rpdo_mapped[RoboteqCANObjects::flags.id][RoboteqCANObjects::flags.subid]),
+    2);
+  state.script_flags = GetByte(
+    static_cast<std::int32_t>(
+      rpdo_mapped[RoboteqCANObjects::flags.id][RoboteqCANObjects::flags.subid]),
+    3);
```
KmakD: Is there a point to create constants representing byte for each flag instead of using 0,1,2,3?

panther_hardware_interfaces/src/roboteq_driver.cpp
```diff
@@ -100,110 +98,81 @@ bool RoboteqDriver::WaitForBoot()
...
 {
-  try {
-    SyncSDOWrite<std::int32_t>(RoboteqCANObjects::cmd_2.id, RoboteqCANObjects::cmd_2.subid, cmd);
-  } catch (const std::runtime_error & e) {
-    throw std::runtime_error(
-      "Error when trying to send channel 2 Roboteq command: " + std::string(e.what()));
-  }
+  tpdo_mapped[RoboteqCANObjects::cmd_1.id][RoboteqCANObjects::cmd_1.subid] = cmd_channel_1;
+  tpdo_mapped[RoboteqCANObjects::cmd_2.id][RoboteqCANObjects::cmd_2.subid] = cmd_channel_2;
+  tpdo_mapped[RoboteqCANObjects::cmd_2.id][RoboteqCANObjects::cmd_2.subid].WriteEvent();
```
KmakD, line 160: Maybe there is something I don't know, but why there is WriteEvent only for `cmd_2`?

panther_hardware_interfaces/src/roboteq_driver.cpp
```diff
@@ -261,38 +230,16 @@ void RoboteqDriver::TurnOnSafetyStopChannel2()
 template <typename T>
 T RoboteqDriver::SyncSDORead(const std::uint16_t index, const std::uint8_t subindex)
```
KmakD: Is this method still needed? If we want to keep it as "it may be useful someday", then at least add comment explaining that

panther_hardware_interfaces/src/motors_controller.cpp
```diff
@@ -81,127 +82,96 @@ void MotorsController::Activate()
...
+    canopen_controller_.GetRearDriver()->SendRoboteqCmd(0, 0);
   } catch (const std::runtime_error & e) {
     throw std::runtime_error("Rear driver send 0 command exception: " + std::string(e.what()));
   }
 
   std::this_thread::sleep_for(std::chrono::milliseconds(1000));
 }
 
-void MotorsController::UpdateSystemFeedback()
+void MotorsController::SetMotorsStates(
```
KmakD: Should be after `UpdateMotorsState()`

panther_hardware_interfaces/src/motors_controller.cpp
```diff
@@ -81,127 +82,96 @@ void MotorsController::Activate()
...
-
-  if (front_can_error || rear_can_error) {
-    throw std::runtime_error("CAN error detected when trying to read Roboteq feedback");
+  if (front_data_.IsCANNetErr() || rear_data_.IsCANNetErr()) {
+    throw std::runtime_error("CAN error detected when trying to read motors states");
   }
 }
 
-bool MotorsController::UpdateDriversState()
+void MotorsController::SetDriverState(
```
KmakD: Should be after `UpdateDriverState()`

panther_hardware_interfaces/src/roboteq_data_converters.cpp
```diff
@@ -147,8 +147,11 @@ RuntimeError::RuntimeError()
       "amps_trigger_activated",
     },
     {
-      "safety_stop_active",
       "amps_limit_active",
+      "safety_stop_active",
+      "forward_limit_triggered",
+      "reverse_limit_triggered",
+      "amps_trigger_activated",
```
KmakD, line 154: Why this are now suppressed flags? 
BTW there is a typo in `FlagError` - should be suppressed not surpressed

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -379,45 +415,32 @@ void PantherSystem::UpdateSystemFeedback()
...
-  can_errors.read_sdo_error = roboteq_error_filter_->IsError(ErrorsFilterIds::READ_SDO);
-  can_errors.read_pdo_error = roboteq_error_filter_->IsError(ErrorsFilterIds::READ_PDO);
-  can_errors.front_data_timed_out = motors_controller_->GetFrontData().IsDataTimedOut();
-  can_errors.rear_data_timed_out = motors_controller_->GetRearData().IsDataTimedOut();
-  can_errors.front_can_net_err = motors_controller_->GetFrontData().IsCANNetErr();
-  can_errors.rear_can_net_err = motors_controller_->GetRearData().IsCANNetErr();
-
-  panther_system_ros_interface_->UpdateMsgErrors(can_errors);
+  panther_system_ros_interface_->PublishDriverState();
 }
```
KmakD: Why this was removed and its content moved to UpdateDriverState? Why not update this method and use it directly?

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -335,40 +341,70 @@ std::vector<CommandInterface> PantherSystem::export_command_interfaces()
...
   } catch (const std::runtime_error & e) {
-    RCLCPP_ERROR_STREAM(logger_, "Error when trying to read drivers feedback: " << e.what());
-    roboteq_error_filter_->UpdateError(ErrorsFilterIds::READ_SDO, true);
+    roboteq_error_filter_->UpdateError(ErrorsFilterIds::READ_PDO_MOTOR_STATES, true);
+    RCLCPP_WARN_STREAM(logger_, "Error when trying to read motors states: " << e.what());
   }
 }
 
-void PantherSystem::UpdateSystemFeedback()
+void PantherSystem::UpdateDriverState()
```
KmakD:
```suggestion
void PantherSystem::UpdateAndPublishDriverState()
```
just a suggestion

panther_hardware_interfaces/CODE_STRUCTURE.md
```diff
@@ -46,5 +49,5 @@ A class that takes care of additional ROS interface of panther system, such as p
 ## PantherSystem
 
 The main class that implements SystemInterface from ros2_control (for details refer to the [ros2_control documentation](https://control.ros.org/master/index.html)).
-
-<!-- todo: when an exception is thrown it is not RT safe (situation may change when we switch to PDO - on hold) -->
+Handles transitions (initialization, activation, shutdown, error, etc.), provides interfaces for feedback (position, velocity, effort) and commands (velocity).
+In the main loop controller should call read and write functions to communicate with motor drivers.
```
KmakD:
```suggestion
In the main loop controller should call `read` and `write` functions to communicate with motor drivers.
```

panther_hardware_interfaces/README.md
```diff
@@ -62,17 +62,21 @@ CAN settings
...
  - `front_driver_can_id` [*int*, default: 1]: CAN ID defined in the properties of Roboteq (set as in [canopen_configuration.yaml](./config/canopen_configuration.yaml))
  - `rear_driver_can_id` [*int*, default: 2]: CAN ID defined in the properties of Roboteq (set as in [canopen_configuration.yaml](./config/canopen_configuration.yaml))
- - `sdo_operation_timeout_ms` [*int*, default: 3 [ms]]: it is set so that the full controller loop takes up to the required time. Each controller loop contains five SDO operations (four writes and one read). For example, in a 100Hz loop, there is up to 10ms for all operations. This timeout should be set so that in the worst case everything takes 10ms.
- - `pdo_feedback_timeout_ms` [*int*, default: 30 [ms]]  - depends on the frequency at which Roboteq is configured to send PDO data. At 100Hz there should be 10ms between received data if it takes more than `pdo_feedback_timeout_ms`, a PDO read error is triggered
+ - `sdo_operation_timeout_ms` [*int*, default: 100 [ms]]: timeout of the SDO operations, currently no SDO operation is required in RT operation, so this timeout can be set to a higher value
```
KmakD:
```suggestion
 - `sdo_operation_timeout_ms` [*int*, default: 100]: timeout in **[ms]** of the SDO operations, currently no SDO operation is required in RT operation, so this timeout can be set to a higher value
```
I don't know if putting units inside is a good idea. Also, units should be bold.

panther_hardware_interfaces/README.md
```diff
@@ -62,17 +62,21 @@ CAN settings
...
+ - `pdo_motor_states_timeout_ms` [*int*, default: 15 [ms]]: depends on the frequency at which Roboteq is configured to send motor states (PDO 1 and 2) data. By default there should 8ms between received data, if it takes more than `pdo_motor_states_timeout_ms`, a motor states read error is triggered. Default value is set to be expected period +50% margin
```
KmakD:
```suggestion
 - `pdo_motor_states_timeout_ms` [*int*, default: 15]: timeout in **[ms]**, depends on the frequency at which Roboteq is configured to send motor states (PDO 1 and 2) data. By default there should be 8 **[ms]** between received data, if it takes more than `pdo_motor_states_timeout_ms`, a motor states read error is triggered. Default value is set to be expected period +50% margin
```
As above

panther_hardware_interfaces/README.md
```diff
@@ -62,17 +62,21 @@ CAN settings
...
+ - `pdo_driver_state_timeout_ms` [*int*, default: 75 [ms]]: depends on the frequency at which Roboteq is configured to send driver state (PDO 3 and 4) data. By default there should 104ms between received data, if it takes more than `pdo_driver_state_timeout_ms`, a driver state read error is triggered. Default value is set to be expected period +50% margin.
+ - `driver_states_update_frequency` [*float*, default: 20.0 [Hz]]: as by default driver state is published with lower frequency, it also shouldn't be updated with every controller loop iteration. Exact frequency at which driver state is published won't match this value - it will depend also on the frequency of the controller (exact value of period can be calculated with the following formula `controller_frequency / ceil(controller_frequency / driver_states_update_frequency)`)
```
KmakD:
```suggestion
 - `pdo_driver_state_timeout_ms` [*int*, default: 75]: timeout in **[ms]**, depends on the frequency at which Roboteq is configured to send driver state (PDO 3 and 4) data. By default there should 104ms between received data, if it takes more than `pdo_driver_state_timeout_ms`, a driver state read error is triggered. Default value is set to be expected period +50% margin.
 - `driver_states_update_frequency` [*float*, default: 20.0]: as by default driver state is published with lower frequency, it also shouldn't be updated with every controller loop iteration. Exact frequency at which driver state is published won't match this value - it will depend also on the frequency of the controller (exact value of period can be calculated with the following formula `controller_frequency / ceil(controller_frequency / driver_states_update_frequency)`)
```
For frequency unit is pretty obvious

panther_hardware_interfaces/README.md
```diff
@@ -62,17 +62,21 @@ CAN settings
...
+ - `max_write_pdo_cmds_errors_count` [*int*, default: 4]: how many consecutive errors can happen before escalating to general error
+ - `max_read_pdo_motor_states_errors_count` [*int*, default: 4]: how many consecutive errors can happen before escalating to general error
+ - `max_read_pdo_driver_state_errors_count` [*int*, default: 20]: how many consecutive errors can happen before escalating to general error
+
 
 > [!CAUTION]
-> `max_write_sdo_errors_count`, `max_read_sdo_errors_count`, `max_read_pdo_errors_count`, `max_safety_stop_attempts`, `sdo_operation_timeout_ms` and `pdo_feedback_timeout_ms` are safety-critical parameters, they should be changed only in very specific cases, be sure that you know how they work and be really cautious when changing them.
+> `max_write_pdo_cmds_errors_count`, `max_read_pdo_motor_states_errors_count`, `max_read_pdo_driver_state_errors_count`, `max_safety_stop_attempts`. `sdo_operation_timeout`, `pdo_motor_states_timeout_ms` and `pdo_driver_state_timeout_ms` are safety-critical parameters, they should be changed only in very specific cases, be sure that you know how they work and be really cautious when changing them.
```
KmakD:
```suggestion
> `max_write_pdo_cmds_errors_count`, `max_read_pdo_motor_states_errors_count`, `max_read_pdo_driver_state_errors_count`, `max_safety_stop_attempts`, `sdo_operation_timeout`, `pdo_motor_states_timeout_ms` and `pdo_driver_state_timeout_ms` are safety-critical parameters, they should be changed only in very specific cases, be sure that you know how they work and be really cautious when changing them.
```

panther_hardware_interfaces/README.md
KmakD, line 1: something I have missed before - according to our README writing guide default values should be bold. Eg.

- `param` [*type*, default: **value**]: description.

panther_hardware_interfaces/config/roboteq_motor_controllers_v80_21a.eds
KmakD, line 1: What is the reason to rename that?

KmakD (changes requested):
Again, PR without tests

panther_hardware_interfaces/config/roboteq_motor_controllers_v80_21a.eds
macstepien: it indicates firmware version that it is intended for (now 2.1a)

panther_hardware_interfaces/src/roboteq_driver.cpp
macstepien: both commands are in the same TPDO, so only one event should be triggered to send both, i'll add comment about it.

panther_hardware_interfaces/src/roboteq_driver.cpp
macstepien: I would prefer not to do it, certainly it won't fit with the `CANopenObject`. Instead we could move this conversion to roboteq data converters, here we would save the whole `int32_t` value to the flags variable in the `RoboteqDriverState`. I think that it will make more sense.

panther_hardware_interfaces/src/roboteq_driver.cpp
macstepien: It is no longer needed, probably will be better to remove it.

panther_hardware_interfaces/src/motors_controller.cpp
macstepien: Not sure if directly after, I moved to the end of file to better match header file.

panther_hardware_interfaces/src/motors_controller.cpp
macstepien: as above

panther_hardware_interfaces/src/roboteq_data_converters.cpp
macstepien: This change was consulted with @reastdignity, and this flags should be also ignored. I'm not sure if this is the best place for this modification, if you would prefer to make it more clear, I can move to a different PR.

panther_hardware_interfaces/README.md
macstepien: after adding `_ms` to the variable name I'm not sure if it necessary to add the `[ms]`

panther_hardware_interfaces/README.md
macstepien: as above, not sure if `[ms]` is even needed

panther_hardware_interfaces/README.md
macstepien: here I also remove `[ms]`

panther_hardware_interfaces/src/panther_system.cpp
macstepien: I did more general refector of the panther system - after changing logic more methods became bloated IMO, so I separated them (commit 61fc921). I also fixed the order of methods, sorry for that, but it became really messy.

panther_hardware_interfaces/src/roboteq_data_converters.cpp
KmakD: We can leave it in this PR I was just not aware why this change has happened

panther_hardware_interfaces/README.md
```diff
@@ -51,28 +51,32 @@ That said apart from the usual interface provided by the ros2_control, this plug
...
+ - `can_interface_name` [*string*, default: **panther_can**]: name of the CAN interface
+ - `master_can_id` [*int*, default: **3**]: CAN ID of the master device (set as in [canopen_configuration.yaml](./config/canopen_configuration.yaml))
+ - `front_driver_can_id` [*int*, default: **1**]: CAN ID defined in the properties of Roboteq (set as in [canopen_configuration.yaml](./config/canopen_configuration.yaml))
+ - `rear_driver_can_id` [*int*, default: **2**]: CAN ID defined in the properties of Roboteq (set as in [canopen_configuration.yaml](./config/canopen_configuration.yaml))
+ - `sdo_operation_timeout_ms` [*int*, default: **100**]: timeout of the SDO operations, currently no SDO operation is required in RT operation, so this timeout can be set to a higher value
+ - `pdo_motor_states_timeout_ms` [*int*, default: **15**]: depends on the frequency at which Roboteq is configured to send motor states (PDO 1 and 2) data. By default there should 10 **[ms]** between received data, if it takes more than `pdo_motor_states_timeout_ms`, a motor states read error is triggered. Default value is set to be expected period +50% margin.
```
KmakD:
```suggestion
 - `pdo_motor_states_timeout_ms` [*int*, default: **15**]: depends on the frequency at which Roboteq is configured to send motor states (PDO 1 and 2) data. By default there should be 10 **[ms]** between received data, if it takes more than `pdo_motor_states_timeout_ms`, a motor states read error is triggered. Default value is set to be expected period +50% margin.
```

panther_hardware_interfaces/README.md
```diff
@@ -51,28 +51,32 @@ That said apart from the usual interface provided by the ros2_control, this plug
...
+ - `pdo_driver_state_timeout_ms` [*int*, default: **75**]: depends on the frequency at which Roboteq is configured to send driver state (PDO 3 and 4) data. By default there should 50 **[ms]** between received data, if it takes more than `pdo_driver_state_timeout_ms`, a driver state read error is triggered. Default value is set to be expected period +50% margin.
```
KmakD:
```suggestion
 - `pdo_driver_state_timeout_ms` [*int*, default: **75**]: depends on the frequency at which Roboteq is configured to send driver state (PDO 3 and 4) data. By default there should be 50 **[ms]** between received data, if it takes more than `pdo_driver_state_timeout_ms`, a driver state read error is triggered. Default value is set to be expected period +50% margin.
```

panther_hardware_interfaces/src/roboteq_driver.cpp
KmakD: I guess we can keep it this way

KmakD (changes requested):

panther_hardware_interfaces/src/roboteq_driver.cpp
macstepien: okay, let's wait for @pkowalsk1 opinion

panther_hardware_interfaces/README.md
```diff
@@ -51,28 +51,32 @@ That said apart from the usual interface provided by the ros2_control, this plug
...
+ - `max_roboteq_initialization_attempts` [*int*, default: **5**]: in some cases, an SDO error can happen during initialization, it is possible to configure more attempts, before escalating to general error
+ - `max_roboteq_activation_attempts` [*int*, default: **5**]: similar to initialization, it is possible to allow some SDO errors before escalating to error
+ <!-- TODO: with GPIODriver it should no longer be needed -->
+ - `max_safety_stop_attempts` [*int*, default: **20**]: how many attempts to activate safety stop will be taken before failing
+ - `max_write_pdo_cmds_errors_count` [*int*, default: **4**]: how many consecutive errors can happen before escalating to general error
+ - `max_read_pdo_motor_states_errors_count` [*int*, default: **4**]: how many consecutive errors can happen before escalating to general error
+ - `max_read_pdo_driver_state_errors_count` [*int*, default: **20**]: how many consecutive errors can happen before escalating to general error
+
```
pkowalsk1:
```suggestion
 - `can_interface_name` [*string*, default: **panther_can**]: name of the CAN interface
 - `master_can_id` [*int*, default: **3**]: CAN ID of the master device (set as in [canopen_configuration.yaml](./config/canopen_configuration.yaml))
 - `front_driver_can_id` [*int*, default: **1**]: CAN ID defined in the properties of Roboteq (set as in [canopen_configuration.yaml](./config/canopen_configuration.yaml))
 - `rear_driver_can_id` [*int*, default: **2**]: CAN ID defined in the properties of Roboteq (set as in [canopen_configuration.yaml](./config/canopen_configuration.yaml))
 - `sdo_operation_timeout_ms` [*int*, default: **100**]: timeout of the SDO operations, currently no SDO operation is required in RT operation so that this timeout can be set to a higher value
 - `pdo_motor_states_timeout_ms` [*int*, default: **15**]: depends on the frequency at which Roboteq is configured to send motor states (PDO 1 and 2) data. By default, there should be 10 **[ms]** between received data, if it takes more than `pdo_motor_states_timeout_ms`, a motor states read error is triggered. The default value is set to be expected period +50% margin.
 - `pdo_driver_state_timeout_ms` [*int*, default: **75**]: depends on the frequency at which Roboteq is configured to send driver state (PDO 3 and 4) data. By default, there should be 50 **[ms]** between received data, if it takes more than `pdo_driver_state_timeout_ms`, a driver state read error is triggered. The default value is set to be expected period +50% margin.
 - `driver_states_update_frequency` [*float*, default: **20.0**]: as by default, the driver state is published with lower frequency, it also shouldn't be updated with every controller loop iteration. The exact frequency at which driver state is published won't match this value - it will also depend on the frequency of the controller (the exact value of the period can be calculated with the following formula `controller_frequency / ceil(controller_frequency / driver_states_update_frequency)`)
 - `max_roboteq_initialization_attempts` [*int*, default: **5**]: in some cases, an SDO error can happen during initialization, it is possible to configure more attempts, before escalating to a general error
 - `max_roboteq_activation_attempts` [*int*, default: **5**]: similar to initialization, it is possible to allow some SDO errors before escalating to error
 <!-- TODO: with GPIODriver it should no longer be needed -->
 - `max_safety_stop_attempts` [*int*, default: **20**]: how many attempts to activate safety stop will be taken before failing
 - `max_write_pdo_cmds_errors_count` [*int*, default: **4**]: how many consecutive errors can happen before escalating to general error
 - `max_read_pdo_motor_states_errors_count` [*int*, default: **4**]: how many consecutive errors can happen before escalating to general error
 - `max_read_pdo_driver_state_errors_count` [*int*, default: **20**]: how many consecutive errors can happen before escalating to general error

```

panther_hardware_interfaces/include/panther_hardware_interfaces/motors_controller.hpp
```diff
@@ -37,6 +37,8 @@ class MotorsController
   MotorsController(
     const CANopenSettings & canopen_settings, const DrivetrainSettings & drivetrain_settings);
 
+  ~MotorsController() { Deinitialize(); }
```
pkowalsk1, line 40: I can't add a comment to these lines, but there are method names like TurnOnEstop, etc. It should be EStop as you mentioned in RP with GPIO controller.

panther_hardware_interfaces/include/panther_hardware_interfaces/motors_controller.hpp
```diff
@@ -58,26 +60,18 @@ class MotorsController
...
-   * @exception std::runtime_error if there was an error
-   * @return whether all updates were finished - only one is read every iteration.
-   * Once it is ready, driver state values can be accessed
+   * @exception std::runtime_error if CAN error was detected
    */
-  bool UpdateDriversState();
+  void UpdateDriversState();
 
   const RoboteqData & GetFrontData() { return front_data_; }
   const RoboteqData & GetRearData() { return rear_data_; }
```
pkowalsk1, line 77:
```suggestion
  const RoboteqData & GetFrontDataRef() { return front_data_; }
  const RoboteqData & GetRearDataRef() { return rear_data_; }
```
I believe we discussed this ref/copy issue previously, but I just suggest adding `Ref` to the method names as a solution. What are your thoughts?

panther_hardware_interfaces/src/roboteq_driver.cpp
```diff
@@ -68,18 +68,16 @@ struct RoboteqCANObjects
...
 }
 
 bool RoboteqDriver::Boot()
 {
   booted_.store(false);
-  return FiberDriver::Boot();
+  return LoopDriver::Boot();
 }
 
 bool RoboteqDriver::WaitForBoot()
```
pkowalsk1: I'm just curious about what happens when we use `std::future<std::string>` in this boot sequence. In my opinion, it would be clearer, but I am aware of the fact that it may not be worth it. Using `std::async()` in `CANopenController::BootDrivers()` and remove  `WaitForBoot()` could be an option there ;)

panther_hardware_interfaces/src/roboteq_driver.cpp
```diff
@@ -100,110 +98,84 @@ bool RoboteqDriver::WaitForBoot()
...
-  fb.runtime_stat_flag_motor_1 = GetByte(
-    static_cast<int32_t>(
-      rpdo_mapped[RoboteqCANObjects::motor_flags.id][RoboteqCANObjects::motor_flags.subid]),
-    0);
-  fb.runtime_stat_flag_motor_2 = GetByte(
-    static_cast<int32_t>(
-      rpdo_mapped[RoboteqCANObjects::motor_flags.id][RoboteqCANObjects::motor_flags.subid]),
-    1);
+  std::unique_lock<std::mutex> lck_sc(speed_current_timestamp_mtx_);
+  states.vel_current_timestamp = last_speed_current_timestamp_;
```
pkowalsk1: We shouldn't unlock lck_p? Or meaby use scoped_lock for multiple mutexes simultaneously?

panther_hardware_interfaces/src/roboteq_driver.cpp
pkowalsk1: In my opinion, using 0, 1, 2, 3 is okay. However, wouldn't it be better to retrieve the flag value only once at the beginning and store it in a variable when the idx and subidx are the same?

panther_hardware_interfaces/src/roboteq_driver.cpp
```diff
@@ -100,110 +98,84 @@ bool RoboteqDriver::WaitForBoot()
...
+  state.battery_current_1 = rpdo_mapped[RoboteqCANObjects::battery_current_1.id]
+                                       [RoboteqCANObjects::battery_current_1.subid];
+  state.battery_current_2 = rpdo_mapped[RoboteqCANObjects::battery_current_2.id]
+                                       [RoboteqCANObjects::battery_current_2.subid];
+
+  std::unique_lock<std::mutex> lck_fa(flags_current_timestamp_mtx_);
+  state.flags_current_timestamp = flags_current_timestamp_;
+
+  std::unique_lock<std::mutex> lck_vt(voltages_temps_timestamp_mtx_);
+  state.voltages_temps_timestamp = last_voltages_temps_timestamp_;
```
pkowalsk1: Same here. What about unlocking it or even using `scoped_lock`?

panther_hardware_interfaces/test/src/roboteq_mock.cpp
```diff
@@ -21,31 +21,48 @@ namespace panther_hardware_interfaces_test
...
-  (*this)[0x2106][static_cast<std::uint8_t>(channel) + 2] = value;
+  (*this)[0x2107][static_cast<std::uint8_t>(channel)] = value;
 }
 
-void RoboteqSlave::SetCurrent(DriverChannel channel, std::int32_t value)
+void RoboteqSlave::SetCurrent(DriverChannel channel, std::int16_t value)
 {
-  (*this)[0x2106][static_cast<std::uint8_t>(channel) + 4] = value;
+  (*this)[0x2100][static_cast<std::uint8_t>(channel)] = value;
 }
```
pkowalsk1: Perhaps move it to the .hpp file, following the pattern of other methods with `static_cast`?

pkowalsk1 (commented):
I'm satisfied with these changes; they are just a few suggestions. 

How about using `""` instead of `<>` when including certain header files?

panther_hardware_interfaces/README.md
macstepien: I think that this change `timeout of the SDO operations, currently no SDO operation is required in RT operation so that this timeout can be set to a higher value` has slightly different meaning, and I would prefer to leave earlier version.

macstepien:
> I'm satisfied with these changes; they are just a few suggestions.
> 
> How about using `""` instead of `<>` when including certain header files?

I just mentioned it yesterday when reviewing GPIO and I agree. This change will have to be done also in other parts of the code ('battery', 'lights').

panther_hardware_interfaces/include/panther_hardware_interfaces/motors_controller.hpp
macstepien: Yes, and I changed in the GPIO PR, I think that it makes more sense there, as it unifies naming with GPIO.

panther_hardware_interfaces/include/panther_hardware_interfaces/motors_controller.hpp
macstepien: I would prefer the regular names (without `Ref`), how about you @KmakD?

panther_hardware_interfaces/src/roboteq_driver.cpp
macstepien: that's good point, I'll change that

panther_hardware_interfaces/src/roboteq_driver.cpp
macstepien: I'm not sure if `async` can be used in this case - `Boot` method only requests boot operation, once it completes `OnBoot` is called, for which `WaitForBoot` waits. It could be done with `future`, but I wouldn't use `string` for that - IMO that would look worse. Instead bool variable can be used, that will store result if the boot was successful or not (change `booted_` to promise). This makes sense, not sure if it will be that better, but I can change it. What do you think about it?

panther_hardware_interfaces/src/roboteq_driver.cpp
macstepien: I would prefer to leave separate locking, I changed locks to `lock_guard` and used shorter scope.

panther_hardware_interfaces/src/roboteq_driver.cpp
macstepien: as above

panther_hardware_interfaces/src/roboteq_driver.cpp
pkowalsk1: Sounds good. @KmakD ?

panther_hardware_interfaces/include/panther_hardware_interfaces/motors_controller.hpp
KmakD: Don't have a strong opinion about this. I guess with `Ref` we leave less room for confusion later

panther_hardware_interfaces/src/roboteq_driver.cpp
KmakD: Sounds good to me too

panther_hardware_interfaces/include/panther_hardware_interfaces/motors_controller.hpp
macstepien: To be honest, I would even prefer to remove reference and just copy, if that can be confusing, instead of adding `Ref` to method names xd I think that it will look strange, and I haven't seen such convention anywhere.

panther_hardware_interfaces/src/roboteq_driver.cpp
macstepien: Done, 488c578836b976d4438cfc85582ea6590090557b, it wasn't even necessary to use bool promise.

pkowalsk1 (approved):

KmakD (approved):

PR #220, opened 2024-01-22
Ros2 - Add GPIO Tests

pkowalsk1:


panther_gpiod/test/test_gpiod.cpp
KmakD, line 1: should be called `test_gpio_driver` like the file that is tested

panther_gpiod/CMakeLists.txt
```diff
@@ -82,4 +79,18 @@ ament_export_libraries(${PROJECT_NAME})
 
 ament_export_dependencies(realtime_tools)
 
+if(BUILD_TESTING)
+  find_package(ament_cmake_gtest REQUIRED)
+
+  ament_add_gtest(${PROJECT_NAME}_test_gpiod test/test_gpiod.cpp)
+  add_dependencies(${PROJECT_NAME}_test_gpiod ${PROJECT_NAME} libgpiod)
```
KmakD: is this line necessary?

panther_gpiod/CMakeLists.txt
```diff
@@ -82,4 +79,18 @@ ament_export_libraries(${PROJECT_NAME})
...
+
+  target_include_directories(
+    ${PROJECT_NAME}_test_gpiod
+    PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+           $<INSTALL_INTERFACE:include>)
+
+  target_link_libraries(${PROJECT_NAME}_test_gpiod ${PROJECT_NAME} libgpiod)
```
KmakD, line 93:
```suggestion
  target_link_libraries(${PROJECT_NAME}_test_gpiod libgpiod)
```
is this necessary?

panther_gpiod/README.md
```diff
@@ -38,3 +38,10 @@ int main() {
 > **NOTE**
 >
 > Not invoking the `GPIOMonitorEnable()` method will result in the absence of functionality to read pin values. It is not mandatory to call this method, but the `IsPinActive()` method throws a runtime error when the GPIO monitor thread is not running.
+
+## Testing
+
+```bash
+colcon build --packages-select panther_gpiod --symlink-install
+colcon test --event-handlers console_direct+ --packages-select panther_gpiod --parallel-workers 1
+```
```
KmakD: This a general method for testing any package with colcon. Is there a point in adding this here, then we should add this to all README files in each package. It will be better to write general testing method in repo main readme file

panther_gpiod/test/test_gpiod.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include <stdexcept>
+#include <streambuf>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <gtest/gtest.h>
+#include "gpiod.hpp"
+
+#include "panther_gpiod/gpio_driver.hpp"
```
KmakD:
```suggestion
#include <gpiod.hpp>

#include <panther_gpiod/gpio_driver.hpp>
```

panther_gpiod/test/test_gpiod.cpp
```diff
@@ -0,0 +1,208 @@
...
+class TestGPIODriver : public ::testing::Test
+{
+public:
+  TestGPIODriver() = default;
+  virtual ~TestGPIODriver() = default;
+
+  void GPIOEventCallback(const GPIOInfo & gpio_info);
+
+protected:
+  void ReinitializeGPIODriver(std::vector<GPIOInfo> gpio_info_storage);
```
KmakD:
```suggestion
  void ReinitializeGPIODriver(std::vector<GPIOInfo> & gpio_info_storage);
```

panther_gpiod/test/test_gpiod.cpp
```diff
@@ -0,0 +1,208 @@
...
+  void SetAndVerifyPinState(const GPIOPin & pin);
+
+  std::unique_ptr<GPIODriver> gpio_driver_;
+  std::pair<GPIOPin, gpiod::line::value> last_gpio_event_summary_;
+  const std::vector<GPIOInfo> gpio_config_info_{
+    GPIOInfo{GPIOPin::LED_SBC_SEL, gpiod::line::direction::OUTPUT},
+  };
+};
+
+void TestGPIODriver::ReinitializeGPIODriver(std::vector<GPIOInfo> gpio_info_storage)
```
KmakD:
```suggestion
void TestGPIODriver::ReinitializeGPIODriver(std::vector<GPIOInfo> & gpio_info_storage)
```

panther_gpiod/test/test_gpiod.cpp
```diff
@@ -0,0 +1,208 @@
...
+  // There is no OS version that supports simultaneous operation of MOTOR_ON and VMOT_ON pins.
+  EXPECT_THROW(
+    {
+      try {
+        ReinitializeGPIODriver({GPIOInfo{GPIOPin::MOTOR_ON, gpiod::line::direction::OUTPUT}});
+        ReinitializeGPIODriver({GPIOInfo{GPIOPin::VMOT_ON, gpiod::line::direction::OUTPUT}});
+      } catch (const std::invalid_argument & e) {
+        EXPECT_STREQ("error requesting GPIO lines", e.what());
+        throw;
+      }
```
KmakD: Does ReinitializeGPIODriver throws multiple std::invalid_argument exceptions?

panther_gpiod/test/test_gpiod.cpp
```diff
@@ -0,0 +1,208 @@
...
+void TestGPIODriver::SetAndVerifyPinState(const GPIOPin & pin)
+{
+  EXPECT_TRUE(gpio_driver_->SetPinValue(pin, true));
+  EXPECT_TRUE(gpio_driver_->IsPinActive(pin));
+
+  EXPECT_TRUE(gpio_driver_->SetPinValue(pin, false));
+  EXPECT_FALSE(gpio_driver_->IsPinActive(pin));
+}
+
+TEST_F(TestGPIODriver, InitGPIODriver_WrongPinConfigFail)
```
KmakD: For InitGPIODriver missing tests for empty info storage

panther_gpiod/test/test_gpiod.cpp
```diff
@@ -0,0 +1,208 @@
...
+  const std::vector<GPIOInfo> gpio_config_info_{
+    GPIOInfo{GPIOPin::LED_SBC_SEL, gpiod::line::direction::OUTPUT},
+  };
+};
+
+void TestGPIODriver::ReinitializeGPIODriver(std::vector<GPIOInfo> gpio_info_storage)
+{
+  gpio_driver_.reset();
+  gpio_driver_ = std::make_unique<GPIODriver>(gpio_info_storage);
+}
```
KmakD: Not sure about, but if you can move `GPIODriverManipulatePin_WrongPinFail` to a simple TEST using something like: 
`std::shared_ptr<TestGPIODriver> driver = std::make_unique<GPIODriver>(gpio_info_storage);`
Then this can be moved to the constructor

panther_gpiod/test/test_gpiod.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+TEST_F(TestGPIODriver, GPIODriverManipulatePin_WrongPinFail)
+{
+  ReinitializeGPIODriver(gpio_config_info_);
+
+  EXPECT_THROW(
+    {
+      try {
+        gpio_driver_->SetPinValue(static_cast<GPIOPin>(-1), true);
```
KmakD:
```suggestion
        this->gpio_driver_->SetPinValue(static_cast<GPIOPin>(-1), true);
```
applies to other tests that use TestGPIODriver objects

panther_gpiod/test/test_gpiod.cpp
```diff
@@ -0,0 +1,208 @@
...
+        ReinitializeGPIODriver({GPIOInfo{GPIOPin::VMOT_ON, gpiod::line::direction::OUTPUT}});
+      } catch (const std::invalid_argument & e) {
+        EXPECT_STREQ("error requesting GPIO lines", e.what());
+        throw;
+      }
+    },
+    std::invalid_argument);
+}
+
+TEST_F(TestGPIODriver, GPIODriverManipulatePin_WrongPinFail)
```
KmakD:
```suggestion
TEST_F(TestGPIODriver, SetPinValue_WrongPinFail)
```

KmakD (changes requested):
missing test for IsPinAvailable method
missing test for IsPinActive method
Dont we want to test CreateLineRequest? For exceptions like "No name defined for one of pins" etc.
Generally i think some private methods could be made protected and tested like HandleEdgeEvent

panther_gpiod/CMakeLists.txt
pkowalsk1: Yes, if `${PROJECT_NAME}` is omitted, an `undefined reference to 'panther_gpiod::GPIODriver:: ...` error occurs.

panther_gpiod/test/test_gpiod.cpp
pkowalsk1: These changes could result in an `binding reference of type 'std::vector<GPIOInfo>&' to 'const std::vector<GPIOInfo>` discards qualifiers` error. Alternatively, if we modify `gpio_config_info_` to be non-const, it may lead to a `Segmentation fault (core dumped)` due to the use of `std::move` in the constructor and potential issues with ownership transfer. 

IMO, using `std::move` instead of referencing is still justified in this case :p

panther_gpiod/test/test_gpiod.cpp
pkowalsk1: Same as above

panther_gpiod/test/test_gpiod.cpp
pkowalsk1: Yes, it's a better solution, but I suggest using `SetUp`. Google Test does not capture exceptions thrown in constructors. If an exception is thrown in `SetUp`, it will be caught, and the test will be marked as failed.

panther_gpiod/test/test_gpio_driver.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+  this->gpio_driver_->GPIOMonitorEnable();
+  this->gpio_driver_->ConfigureEdgeEventCallback(
+    std::bind(&TestGPIODriver::GPIOEventCallback, this, std::placeholders::_1));
+
+  EXPECT_TRUE(this->gpio_driver_->SetPinValue(tested_pin, true));
+  EXPECT_TRUE(this->gpio_driver_->IsPinActive(tested_pin));
+
+  EXPECT_EQ(last_gpio_event_summary_.first, tested_pin);
+  EXPECT_EQ(last_gpio_event_summary_.second, gpiod::line::value::ACTIVE);
```
KmakD:
```suggestion
  EXPECT_EQ(this->last_gpio_event_summary_.first, tested_pin);
  EXPECT_EQ(this->last_gpio_event_summary_.second, gpiod::line::value::ACTIVE);
```

panther_gpiod/test/test_gpio_driver.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+  EXPECT_TRUE(this->gpio_driver_->SetPinValue(tested_pin, false));
+  EXPECT_FALSE(this->gpio_driver_->IsPinActive(tested_pin));
+
+  EXPECT_EQ(last_gpio_event_summary_.first, tested_pin);
+  EXPECT_EQ(last_gpio_event_summary_.second, gpiod::line::value::INACTIVE);
```
KmakD:
```suggestion
  EXPECT_EQ(this->last_gpio_event_summary_.first, tested_pin);
  EXPECT_EQ(this->last_gpio_event_summary_.second, gpiod::line::value::INACTIVE);
```

panther_gpiod/test/test_gpio_driver.cpp
```diff
@@ -0,0 +1,206 @@
...
+  EXPECT_THROW(
+    {
+      try {
+        this->gpio_driver_->SetPinValue(static_cast<GPIOPin>(-1), true);
+      } catch (const std::invalid_argument & e) {
+        EXPECT_STREQ("Pin not found in GPIO info storage.", e.what());
+        throw;
+      }
+    },
+    std::invalid_argument);
```
KmakD: You could use ExpectThrowWithDescription from panther_utils

panther_gpiod/test/test_gpio_driver.cpp
```diff
@@ -0,0 +1,206 @@
...
+    {
+      try {
+        this->gpio_driver_->ConfigureEdgeEventCallback(
+          std::bind(&TestGPIODriver::GPIOEventCallback, this, std::placeholders::_1));
+      } catch (const std::runtime_error & e) {
+        EXPECT_STREQ("GPIO monitor thread is not running!", e.what());
+        throw;
+      }
+    },
+    std::runtime_error);
```
KmakD: Same as above

panther_gpiod/test/test_gpio_driver.cpp
```diff
@@ -0,0 +1,206 @@
...
+  EXPECT_THROW(
+    {
+      try {
+        this->gpio_driver_->SetPinValue(GPIOPin::LED_SBC_SEL, true);
+      } catch (const std::invalid_argument & e) {
+        EXPECT_STREQ("Cannot set value for INPUT pin.", e.what());
+        throw;
+      }
+    },
+    std::invalid_argument);
```
KmakD: As above

panther_gpiod/test/test_gpio_driver.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+  SetAndVerifyPinState(GPIOPin::LED_SBC_SEL);
+}
+
+int main(int argc, char ** argv)
+{
+  testing::InitGoogleTest(&argc, argv);
+  auto run_tests = RUN_ALL_TESTS();
+
+  return run_tests;
```
KmakD:
```suggestion
  testing::InitGoogleTest(&argc, argv);
  return RUN_ALL_TESTS();
```

KmakD (changes requested):

KmakD (approved):

PR #222, opened 2024-02-05
Add GPIO controller

macstepien:


panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+
+#include <atomic>
+#include <functional>
+#include <memory>
+#include <thread>
+#include <vector>
+
+#include <gpiod.hpp>
+
+#include "panther_gpiod/gpio_driver.hpp"
```
macstepien, line 32: GPIO is the only place where we use `#include ""`, I think that it is correct, so it should be fixed in other parts.

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+   *   }
+   * };
+   *
+   * MyClass my_obj;
+   * GPIOControllerPTH12X gpio_controller;
+   * gpio_controller.ConfigureGpioStateCallback(
+   *     std::bind(&MyClass::HandleGPIOEvent, &my_obj, std::placeholders::_1));
+   * @endcode
+   */
+  void ConfigureGpioStateCallback(
```
macstepien:
```suggestion
  void ConfigureGPIOStateCallback(
```

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+   * @return true if the Watchdog thread is successfully disabled.
+   * @exception std::runtime_error when the Watchdog thread fails to stop.
+   */
+  bool EStopTrigger() override;
+
+  /**
+   * @brief Resets the E-Stop.
+   *
+   * This method verifies the status of the E_STOP_RESET pin, which is configured as an input.
+   * If the pin is active, it attempts to reset the E-Stop by momentarily setting it to an inactive
```
macstepien:
```suggestion
   * If the pin is active, it attempts to reset the E-stop by momentarily setting it to an inactive
```
Also should be changed in other places (`E-stop` is used in manual).

panther_hardware_interfaces/src/gpio_controller.cpp
```diff
@@ -0,0 +1,229 @@
...
+  watchdog_thread_->join();
+
+  return !IsWatchdogEnabled();
+}
+
+void Watchdog::WatchdogThread()
+{
+  while (watchdog_thread_enabled_) {
+    bool value = gpio_driver_->IsPinActive(watchdog_pin_);
+
```
macstepien:
```suggestion
    const bool value = gpio_driver_->IsPinActive(watchdog_pin_);
```

panther_hardware_interfaces/src/gpio_controller.cpp
```diff
@@ -0,0 +1,229 @@
...
+  gpio_driver_->GPIOMonitorEnable(true, 60);
+
+  gpio_driver_->SetPinValue(panther_gpiod::GPIOPin::VMOT_ON, true);
+  MotorsEnable(true);
+
+  watchdog_ = std::make_unique<Watchdog>(gpio_driver_);
+}
+
+// TODO: void
+bool GPIOControllerPTH12X::EStopTrigger()
```
macstepien: I think that it will be better to change both `EStopTrigger` and `EStopReset` to void. We could also return false instead of throwing, but I would prefer void with throw (it can provide more information through message).

panther_hardware_interfaces/src/gpio_controller.cpp
```diff
@@ -0,0 +1,229 @@
...
+{
+  if (!watchdog_->TurnOff()) {
+    throw std::runtime_error("Can't stop watchdg thread");
```
macstepien:
```suggestion
    throw std::runtime_error("Can't stop watchdog thread");
```

panther_hardware_interfaces/src/gpio_controller.cpp
```diff
@@ -0,0 +1,229 @@
...
+  }
+  return true;
+}
+
+bool GPIOControllerPTH12X::EStopReset()
+{
+  auto e_stop_pin = panther_gpiod::GPIOPin::E_STOP_RESET;
```
macstepien:
```suggestion
  using panther_gpiod::GPIOPin;
```
and change e_stop_pin to panther_gpiod::E_STOP_RESET. This is just a suggestion, if you prefer variable, then add `const`.

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -70,6 +69,15 @@ CallbackReturn PantherSystem::on_configure(const rclcpp_lifecycle::State &)
...
   RCLCPP_INFO(logger_, "Configuring Panther System");
 
+  RCLCPP_INFO_STREAM(logger_, "Creating GPIO controller for Panther version: " << panther_version_);
+  if (panther_version_ >= 1.2 - std::numeric_limits<float>::epsilon()) {
+    gpio_controller_ = std::make_shared<GPIOControllerPTH12X>();
+  } else {
+    gpio_controller_ = std::make_shared<GPIOControllerPTH10X>();
+  }
+
+  gpio_controller_->Start();
```
macstepien: `Start` should be inside try catch

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -109,9 +119,38 @@ CallbackReturn PantherSystem::on_activate(const rclcpp_lifecycle::State &)
...
+  panther_system_ros_interface_ =
+    std::make_shared<PantherSystemRosInterface>("panther_system_node");
+
+  gpio_controller_->ConfigureGpioStateCallback(std::bind(
+    &PantherSystemRosInterface::PublishGPIOState, panther_system_ros_interface_,
+    std::placeholders::_1));
+
+  panther_system_ros_interface_->AddSetBoolService(
+    "~/motor_power_enable",
+    std::bind(&GPIOControllerInterface::MotorsEnable, gpio_controller_, std::placeholders::_1));
```
macstepien:
```suggestion
    std::bind(&GPIOControllerInterface::MotorPowerEnable, gpio_controller_, std::placeholders::_1));
```

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -109,9 +119,38 @@ CallbackReturn PantherSystem::on_activate(const rclcpp_lifecycle::State &)
...
+  panther_system_ros_interface_->AddSetBoolService(
+    "~/fan_enable",
+    std::bind(&GPIOControllerInterface::FanEnable, gpio_controller_, std::placeholders::_1));
+  panther_system_ros_interface_->AddSetBoolService(
+    "~/aux_power_enable",
+    std::bind(&GPIOControllerInterface::AUXEnable, gpio_controller_, std::placeholders::_1));
```
macstepien:
```suggestion
    std::bind(&GPIOControllerInterface::AUXPowerEnable, gpio_controller_, std::placeholders::_1));
```

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -109,9 +119,38 @@ CallbackReturn PantherSystem::on_activate(const rclcpp_lifecycle::State &)
...
+  panther_system_ros_interface_->AddSetBoolService(
+    "~/digital_power_enable",
+    std::bind(&GPIOControllerInterface::VDIGEnable, gpio_controller_, std::placeholders::_1));
```
macstepien:
```suggestion
    std::bind(&GPIOControllerInterface::DigitalPowerEnable, gpio_controller_, std::placeholders::_1));
```

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -70,6 +69,15 @@ CallbackReturn PantherSystem::on_configure(const rclcpp_lifecycle::State &)
 {
   RCLCPP_INFO(logger_, "Configuring Panther System");
 
+  RCLCPP_INFO_STREAM(logger_, "Creating GPIO controller for Panther version: " << panther_version_);
+  if (panther_version_ >= 1.2 - std::numeric_limits<float>::epsilon()) {
```
macstepien, line 73: Probably better to move it into a separate method (as it is used also in other parts).

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -548,6 +606,82 @@ bool PantherSystem::CheckIfSafetyStopActive()
...
+
+  roboteq_error_filter_->SetClearErrorsFlag();
+  e_stop_ = false;
+}
+
+bool PantherSystem::ReadEStop()
+{
+  if (panther_version_ >= 1.2 - std::numeric_limits<float>::epsilon()) {
+    // TODO: it has reversed logic
+    return !gpio_controller_->IsPinActive(panther_gpiod::GPIOPin::E_STOP_RESET);
```
macstepien: I thought about changing the `active_low` setting, but when I reconsidered, It makes more sense that EStop is true when EStop reset is false, so we can drop it.

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -28,34 +28,75 @@
...
+  try {
+    callback_();
+    response->success = true;
+  } catch (const std::exception & err) {
+    response->success = false;
+    response->message = err.what();
+
+    RCLCPP_INFO(
+      rclcpp::get_logger("PantherSystem"), "Trigger service response: %s",
+      response->message.c_str());
```
macstepien:
```suggestion
    RCLCPP_WARN_STREAM(
      rclcpp::get_logger("PantherSystem"), "Trigger service failed: " << response->message);
```

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -28,34 +28,75 @@
...
+  try {
+    callback_(request->data);
+    response->success = true;
+  } catch (const std::exception & err) {
+    response->success = false;
+    response->message = err.what();
+
+    RCLCPP_INFO(
+      rclcpp::get_logger("PantherSystem"), "SetBool service response: %s",
+      response->message.c_str());
```
macstepien:
```suggestion
    RCLCPP_WARN_STREAM(
      rclcpp::get_logger("PantherSystem"), "SetBool service response: " << response->message);
```

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -28,34 +28,75 @@
...
   node_ = std::make_shared<rclcpp::Node>(node_name, node_options);
   executor_ = std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
   executor_->add_node(node_);
 
   executor_thread_ = std::make_unique<std::thread>([this]() { executor_->spin(); });
 
-  clear_errors_ = clear_errors;
-
-  driver_state_publisher_ = node_->create_publisher<DriverStateMsg>(
+  driver_state_publisher_ = node_->create_publisher<panther_msgs::msg::DriverState>(
```
macstepien:
```suggestion
  driver_state_publisher_ = node_->create_publisher<DriverStateMsg>(
```
Also should be chnaged in case of other services/msgs

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -117,24 +184,93 @@ void PantherSystemRosInterface::UpdateMsgErrors(const CANErrors & can_errors)
   driver_state.rear.can_net_err = can_errors.rear_can_net_err;
 }
 
+void PantherSystemRosInterface::InitializeAndPublishIOStateMsg(
+  std::shared_ptr<GPIOControllerInterface> gpio_controller, const float panther_version)
+{
+  auto & io_state = realtime_io_state_publisher_->msg_;
+
+  if (panther_version >= 1.2 - std::numeric_limits<float>::epsilon()) {
```
macstepien: This should be moved to some function.

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -117,24 +184,93 @@ void PantherSystemRosInterface::UpdateMsgErrors(const CANErrors & can_errors)
...
+    realtime_io_state_publisher_->unlockAndPublish();
+  }
+}
+
+void PantherSystemRosInterface::InitializeAndPublishEStopStateMsg(const bool e_stop)
+{
+  realtime_e_stop_state_publisher_->msg_.data = e_stop;
+  if (realtime_e_stop_state_publisher_->trylock()) {
+    realtime_e_stop_state_publisher_->unlockAndPublish();
+  }
```
macstepien, line 182: should be moved to another method

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -117,24 +184,93 @@ void PantherSystemRosInterface::UpdateMsgErrors(const CANErrors & can_errors)
...
   }
 }
 
-void PantherSystemRosInterface::ClearErrorsCb(
-  TriggerSrv::Request::ConstSharedPtr /* request */, TriggerSrv::Response::SharedPtr response)
+void PantherSystemRosInterface::PublishGPIOState(const panther_gpiod::GPIOInfo & gpio_info)
 {
-  RCLCPP_INFO(node_->get_logger(), "Clearing errors");
+  auto & io_state = realtime_io_state_publisher_->msg_;
+  bool pin_value = (gpio_info.value == gpiod::line::value::ACTIVE);
```
macstepien:
```suggestion
  const bool pin_value = (gpio_info.value == gpiod::line::value::ACTIVE);
```

macstepien (commented):
@pkowalsk1 Please also update `CODE_STRUCTURE.md`

panther_hardware_interfaces/src/gpio_controller.cpp
pkowalsk1: I think `const` is the clearest option, okay?

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+   * MyClass my_obj;
+   * GPIOControllerPTH12X gpio_controller;
+   * gpio_controller.ConfigureGpioStateCallback(
+   *     std::bind(&MyClass::HandleGPIOEvent, &my_obj, std::placeholders::_1));
+   * @endcode
+   */
+  void ConfigureGpioStateCallback(
+    const std::function<void(const panther_gpiod::GPIOInfo &)> & callback);
+
+  bool IsPinActive(const panther_gpiod::GPIOPin pin) const;
```
KmakD, line 124:
```suggestion
  bool IsPinActive(const panther_gpiod::GPIOPin & pin) const;
```
Not sure if we should add reference for enum class tho

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+{
+public:
+  /**
+   * @brief Constructor for Watchdog class.
+   *
+   * @param gpio_driver Pointer to the GPIODriver object.
+   * @exception std::runtime_error if the Watchdog pin is not configured by GPIODriver or not
+   * described in GPIOController gpio_info storage
+   */
+  Watchdog(std::shared_ptr<panther_gpiod::GPIODriver> gpio_driver);
```
KmakD, line 47:
```suggestion
  Watchdog(const std::shared_ptr<panther_gpiod::GPIODriver> & gpio_driver);
```

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+   */
+  bool MotorsEnable(const bool enable) override;
+
+  /**
+   * @brief Controls the fan based on the 'enable' parameter.
+   *
+   * @param enable Set to 'true' to enable the motors, 'false' to disable.
+   * @return 'true' if the motor control pin value is successfully set, 'false' otherwise.
+   */
+  bool FanEnable(const bool enable) override;
```
KmakD:
```suggestion
   * @param enable Set to 'true' to enable the fan, 'false' to disable.
   * @return 'true' if the fan control pin value is successfully set, 'false' otherwise.
   */
  bool FanEnable(const bool enable) override;
```

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+
+  /**
+   * @brief Controls AUX power source based on the 'enable' parameter.
+   *
+   * @param enable Set to 'true' to enable the motors, 'false' to disable.
+   * @return 'true' if the motor control pin value is successfully set, 'false' otherwise.
+   */
+  bool AUXEnable(const bool enable) override;
```
KmakD:
```suggestion
   * @param enable Set to 'true' to enable the AUX, 'false' to disable.
   * @return 'true' if the AUX control pin value is successfully set, 'false' otherwise.
   */
  bool AUXEnable(const bool enable) override;
```

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+
+  /**
+   * @brief Controls the digital power source based on the 'enable' parameter.
+   *
+   * @param enable Set to 'true' to enable the motors, 'false' to disable.
+   * @return 'true' if the motor control pin value is successfully set, 'false' otherwise.
+   */
+  bool VDIGEnable(const bool enable) override;
```
KmakD:
```suggestion
   * @param enable Set to 'true' to enable the VDIG 'false' to disable.
   * @return 'true' if the VDIG control pin value is successfully set, 'false' otherwise.
   */
  bool VDIGEnable(const bool enable) override;
```

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+
+  bool ChargerEnable(const bool enable) override;
```
KmakD, line 199: Other enable methods have briefs this doesn't

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+   * @brief Placeholder method indicating lack of hardware E-Stop support for the robot in this
+   * version.
+   *
+   * @return Always returns true.
+   */
+  bool EStopTrigger() override;
+
+  /**
+   * @brief Checks if the motors are powered up (when STAGE2 is active) without controlling any
+   * GPIO.
```
KmakD:
```suggestion
   * @brief Checks if the motors are powered up (when STAGE2_INPUT is active) without controlling any
   * GPIO.
```
or STAGE2_INPUT/main switch is set to STAGE2 position

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,294 @@
...
+
+  /**
+   * @brief Placeholder method indicating lack of support for controlling the fan in this robot
+   * version.
+   *
+   * @param enable Ignored parameter in this version.
+   * @exception std::runtime_error Always throws a runtime error due to lack of support for fan
+   * control.
+   */
+  bool FanEnable(const bool /* enable */) override;
```
KmakD, line 277:
```suggestion
   * @exception std::runtime_error Always throws a runtime error due to lack of support for fan
   * control.
   */
  bool FanEnable(const bool /* enable */) override;
```
Is there a point in putting param here? method definition indicates it is not used. Applies to other methods too

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,296 @@
...
+   * @brief Placeholder method indicating lack of support for controlling Digital Power in this
+   * robot version.
+   *
+   * @param enable Ignored parameter in this version.
+   * @exception std::runtime_error Always throws a runtime error due to lack of support for fan
+   * control.
+   */
+  bool DigitalPowerEnable(const bool /* enable */) override;
+
+  bool ChargerEnable(const bool /* enable */) override;
```
KmakD, line 307: same as earlier

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system.hpp
```diff
@@ -124,15 +132,19 @@ class PantherSystem : public hardware_interface::SystemInterface
...
   std::shared_ptr<RoboteqErrorFilter> roboteq_error_filter_;
 
+  float panther_version_;
+
+  std::atomic_bool e_stop_ = true;
+
+  std::atomic_bool last_commands_zero_ = false;
+
+  std::mutex motor_controller_write_mtx_;
+
```
KmakD:
```suggestion
  std::atomic_bool e_stop_ = true;
  std::atomic_bool last_commands_zero_ = false;
  std::mutex motor_controller_write_mtx_;

```
I think this can be grupped

panther_hardware_interfaces/package.xml
```diff
@@ -17,6 +17,7 @@
 
   <depend>controller_interface</depend>
   <depend>hardware_interface</depend>
+  <depend>panther_gpiod</depend>
```
KmakD, line 21: Add also `std_msgs`

panther_hardware_interfaces/package.xml
KmakD, line 1: You can update authors :wink:

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
```diff
@@ -54,6 +59,36 @@ struct CANErrors
   bool rear_can_net_err;
 };
 
+class TriggerServiceWrapper
```
KmakD, line 67: I'm not sure if this wrappers are not an overkill.
If not then it can be definitely moved to a template class, sth like:
```cpp
template <typename SrvT, typename CallbackT>
class ServiceWrapper(const std::function<CallbackT> & callback)
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
```diff
@@ -54,6 +59,36 @@ struct CANErrors
...
+{
+public:
+  TriggerServiceWrapper(const std::function<void()> & callback) : callback_(callback){};
+
+  void CallbackWrapper(
+    std_srvs::srv::Trigger::Request::ConstSharedPtr /* request */,
+    std_srvs::srv::Trigger::Response::SharedPtr response);
```
KmakD:
```suggestion
  void CallbackWrapper(
    const TriggerSrv::Request::ConstSharedPtr /* request */,
    TriggerSrv::Response::SharedPtr response);
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
```diff
@@ -54,6 +59,36 @@ struct CANErrors
...
+};
+
+class SetBoolServiceWrapper
+{
+public:
+  SetBoolServiceWrapper(const std::function<void(const bool)> & callback) : callback_(callback){};
+
+  void CallbackWrapper(
+    std_srvs::srv::SetBool::Request::ConstSharedPtr request,
+    std_srvs::srv::SetBool::Response::SharedPtr response);
```
KmakD:
```suggestion
  void CallbackWrapper(
    const SetBoolSrv::Request::ConstSharedPtr request,
    SetBoolSrv::Response::SharedPtr response);
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
```diff
@@ -54,6 +59,36 @@ struct CANErrors
...
+
+  rclcpp::Service<std_srvs::srv::SetBool>::SharedPtr service;
+
+private:
+  std::function<void(bool enable)> callback_;
```
KmakD:
```suggestion
  std::function<void(const bool)> callback_;
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
```diff
@@ -70,10 +105,17 @@ class PantherSystemRosInterface
...
     const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions());
   ~PantherSystemRosInterface();
 
+  /**
+   * @brief Adds new service server to node
+   */
+  void AddTriggerService(const std::string service_name, const std::function<void()> & callback);
+  void AddSetBoolService(
+    const std::string service_name, const std::function<void(const bool)> & callback);
+
```
KmakD, line 122: This could also be a template method

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
```diff
@@ -24,17 +24,22 @@
...
+#include <panther_hardware_interfaces/gpio_controller.hpp>
 #include <panther_hardware_interfaces/roboteq_data_converters.hpp>
 
 namespace panther_hardware_interfaces
 {
 
-using TriggerSrv = std_srvs::srv::Trigger;
 using DriverStateMsg = panther_msgs::msg::DriverState;
+using IOStateMsg = panther_msgs::msg::IOState;
+using TriggerSrv = std_srvs::srv::Trigger;
```
KmakD:
```suggestion
using DriverStateMsg = panther_msgs::msg::DriverState;
using IOStateMsg = panther_msgs::msg::IOState;
using BoolMsg = std_msgs::msg::Bool;
using SetBoolSrv = std_srvs::srv::SetBoolSrv;
using TriggerSrv = std_srvs::srv::Trigger;
```
and this can be applied later

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -117,24 +181,93 @@ void PantherSystemRosInterface::UpdateMsgErrors(const CANErrors & can_errors)
...
+    io_state.motor_on = gpio_controller->IsPinActive(panther_gpiod::GPIOPin::VMOT_ON);
+  } else {
+    io_state.aux_power = true;
+    io_state.charger_connected = false;
+    io_state.charger_enabled = false;
+    io_state.digital_power = true;
+    io_state.fan = false;
+    io_state.power_button = false;
+    io_state.motor_on = gpio_controller->IsPinActive(panther_gpiod::GPIOPin::MOTOR_ON);
+  }
```
KmakD: I wonder if this make sense. Using any conditional blocks if we have an abstraction seems a bad practice.
Maybe controller abstraction should have methods like `GetAUXPowerState` etc. and return some default value for older panther? What do you think?

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -117,24 +181,93 @@ void PantherSystemRosInterface::UpdateMsgErrors(const CANErrors & can_errors)
   driver_state.rear.can_net_err = can_errors.rear_can_net_err;
 }
 
+void PantherSystemRosInterface::InitializeAndPublishIOStateMsg(
```
KmakD:
```suggestion
void PantherSystemRosInterface::PublishIOStateMsg(
```

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -117,24 +181,93 @@ void PantherSystemRosInterface::UpdateMsgErrors(const CANErrors & can_errors)
...
+    io_state.power_button = false;
+    io_state.motor_on = gpio_controller->IsPinActive(panther_gpiod::GPIOPin::MOTOR_ON);
+  }
+
+  if (realtime_io_state_publisher_->trylock()) {
+    realtime_io_state_publisher_->unlockAndPublish();
+  }
+}
+
+void PantherSystemRosInterface::InitializeAndPublishEStopStateMsg(const bool e_stop)
```
KmakD:
```suggestion
void PantherSystemRosInterface::PublishEStopStateMsg(const bool e_stop)
```
Is this method needed? Won't PublishEStopStateIfChanged be enough?

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -117,24 +181,93 @@ void PantherSystemRosInterface::UpdateMsgErrors(const CANErrors & can_errors)
...
 void PantherSystemRosInterface::PublishDriverState()
 {
   if (realtime_driver_state_publisher_->trylock()) {
     realtime_driver_state_publisher_->unlockAndPublish();
   }
 }
 
-void PantherSystemRosInterface::ClearErrorsCb(
-  TriggerSrv::Request::ConstSharedPtr /* request */, TriggerSrv::Response::SharedPtr response)
+void PantherSystemRosInterface::PublishGPIOState(const panther_gpiod::GPIOInfo & gpio_info)
```
KmakD: I'm confused, another method for publishing IO states?

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -68,6 +69,20 @@ CallbackReturn PantherSystem::on_configure(const rclcpp_lifecycle::State &)
...
+  if (panther_version_ >= 1.2 - std::numeric_limits<float>::epsilon()) {
+    gpio_controller_ = std::make_shared<GPIOControllerPTH12X>();
+  } else {
+    gpio_controller_ = std::make_shared<GPIOControllerPTH10X>();
+  }
+
+  try {
+    gpio_controller_->Start();
+  } catch (const std::runtime_error & e) {
+    RCLCPP_FATAL(logger_, "GPIO controller initialization failed");
```
KmakD:
```suggestion
    RCLCPP_FATAL(logger_, "GPIO controller initialization failed. Error: " + std::string(e.what());
```

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -546,6 +612,81 @@ bool PantherSystem::CheckIfSafetyStopActive()
...
+
+bool PantherSystem::ReadEStop()
+{
+  if (panther_version_ >= 1.2 - std::numeric_limits<float>::epsilon()) {
+    return !gpio_controller_->IsPinActive(panther_gpiod::GPIOPin::E_STOP_RESET);
+  } else {
+    // For older panther versions there is no hardware E-stop
+    return e_stop_;
+  }
+}
```
KmakD: Again this could be moved to gpio controller abstraction and then you can probably remove e_stop_ variable

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -546,6 +612,81 @@ bool PantherSystem::CheckIfSafetyStopActive()
...
+
+void PantherSystem::SetEStop()
+{
+  RCLCPP_INFO(logger_, "Setting E-stop");
+  bool gpio_controller_error = false;
+
+  try {
+    gpio_controller_->EStopTrigger();
+  } catch (const std::runtime_error & e) {
+    RCLCPP_ERROR_STREAM(logger_, "Error when trying to set E-stop stop using GPIO: " << e.what());
```
KmakD:
```suggestion
    RCLCPP_ERROR_STREAM(logger_, "Error when trying to set E-stop using GPIO: " << e.what());
```

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -546,6 +612,81 @@ bool PantherSystem::CheckIfSafetyStopActive()
...
+    {
+      std::lock_guard<std::mutex> lck_g(motor_controller_write_mtx_);
+      motors_controller_->TurnOnSafetyStop();
+    }
+  } catch (const std::runtime_error & e) {
+    RCLCPP_ERROR_STREAM(
+      logger_, "Error when trying to set safety stop using CAN command: " << e.what());
+    if (gpio_controller_error) {
+      RCLCPP_ERROR_STREAM(logger_, "Both attempts at setting E-stop failed");
+      throw std::runtime_error("Both attempts at setting E-stop failed");
```
KmakD: Is there a point in log error if we throw the exception with the same info?

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -546,6 +612,81 @@ bool PantherSystem::CheckIfSafetyStopActive()
...
+  }
+
+  e_stop_ = true;
+}
+
+void PantherSystem::ResetEStop()
+{
+  RCLCPP_INFO(logger_, "Resetting E-stop");
+
+  // On side of the motors controller safety stop is reset by sending 0.0 commands
```
KmakD:
```suggestion
  // On the side of the motors controller safety stop is reset by sending 0.0 commands
```

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -546,6 +612,81 @@ bool PantherSystem::CheckIfSafetyStopActive()
...
+  if (!last_commands_zero_) {
+    RCLCPP_ERROR(
+      logger_,
+      "Can't reset E-stop - last velocity commands are different than zero. Make sure that your "
+      "controller sends zero commands before trying to reset E-stop.");
+    throw std::runtime_error(
+      "Can't reset E-stop - last velocity commands are different than zero. Make sure that your "
+      "controller sends zero commands before trying to reset E-stop.");
```
KmakD: Same as above

KmakD (changes requested):

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
pkowalsk1: It's similar to an integer. The copy process is faster, and there is no dereference stage when using this parameter in the method.

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
pkowalsk1: I believe that passing this by value and making a copy of it is a good practice ([link](https://stackoverflow.com/questions/8385457/should-i-pass-a-shared-ptr-by-reference)).

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
pkowalsk1: We've attempted this previously, and it's quite challenging because the std::function parameter may or may not include the 'enable' parameter. It's difficult to pass callbacks with varying parameters and ensure that this parameter is correctly passed during callback invocation when using templates.

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
pkowalsk1: same as above

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
pkowalsk1: While those topics are latched, it's necessary to publish them once during the activation transition. Subsequent publications occur during the GPIO event callback named `PublishGPIOState`. I propose to rename `PublishGPIOState` to `PublishIOState`.

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
pkowalsk1: Refer to the preceding comment.

panther_hardware_interfaces/src/panther_system.cpp
pkowalsk1: However, `e_stop_` is also manipulated in PantherSystem.

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
pkowalsk1: After discussion with @KmakD and @pawelirh, we have two options:

1. Overriding the `IsPinActive` method in `GPIOController` so that in PTH 1.2 it returns `gpio_driver_->IsPinActive(pin)`, and in 1.06 it has a switch case that returns the appropriate pin state based on the enum.
1. Implementing a Get method in GPIOControllerInterface that would be called in IsPinActive and behave as follows:
- for version 1.20: `return gpio_driver_->IsPinActive(pin);`
- for version 1.06:
    - if the pin is MOTOR_ON, it would call the IsPinActive method of the driver (as in 1.20),
    - if not, it would return a default pin value defined in some map (switch case still needed).

@macstepien, what is your opinion?

pkowalsk1: I have proposed an initial solution in this commit: https://github.com/husarion/panther_ros/pull/222/commits/666983f2b733ad13acdc5e32ed9a67a2eeec40e1

This solution has not yet been tested on the robot.

panther_hardware_interfaces/src/gpio_controller.cpp
```diff
@@ -0,0 +1,260 @@
...
+  }
+}
+
+void GPIOControllerPTH12X::EStopReset()
+{
+  const auto e_stop_pin = panther_gpiod::GPIOPin::E_STOP_RESET;
+  bool e_stop_state = !gpio_driver_->IsPinActive(e_stop_pin);
+
+  if (!e_stop_state) {
+    std::cout << "[GPIOController] E-STOP is not active, reset is not needed" << std::endl;
```
macstepien:
```suggestion
    std::cout << "[GPIOController] E-stop is not active, reset is not needed" << std::endl;
```

panther_hardware_interfaces/src/gpio_controller.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+  gpio_driver_->ChangePinDirection(e_stop_pin, gpiod::line::direction::INPUT);
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+  e_stop_state = !gpio_driver_->IsPinActive(e_stop_pin);
+
+  if (e_stop_state) {
+    watchdog_->TurnOff();
+    throw std::runtime_error(
+      "E-STOP reset failed, check for pressed E-STOP buttons or other triggers");
```
macstepien:
```suggestion
      "E-stop reset failed, check for pressed E-stop buttons or other triggers");
```

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -28,34 +28,69 @@
...
+{
+  try {
+    callback_(request->data);
+    response->success = true;
+  } catch (const std::exception & err) {
+    response->success = false;
+    response->message = err.what();
+
+    RCLCPP_WARN_STREAM(
+      rclcpp::get_logger("PantherSystem"), "SetBool service response: " << response->message);
```
macstepien:
```suggestion
      rclcpp::get_logger("PantherSystem"), "SetBool service failed: " << response->message);
```

panther_hardware_interfaces/CODE_STRUCTURE.md
```diff
@@ -38,9 +38,14 @@ Feedback converters are combined in the `RoboteqData` class to provide the full
...
 
-## GPIODriver
+## GPIOController
 
-WIP - it will handle reading/writing pins of the RPi GPIO.
+The GPIOController provides wrappers for the GPIO driver from the `panther_gpiod` package, handling reading and writing pins of the RPi GPIO. It includes the following utilities:
+
+* `GPIOControllerInterface`: Interface for all wrappers that handle GPIO control tasks.
+* `GPIOControllerPTH12X`: Class with specific logic for the Panther robot with version 1.20 and above.
+* `GPIOControllerPTH10X`: Class with specific logic for the Panther robot with version less than 1.20.
```
macstepien:
```suggestion
* `GPIOControllerPTH10X`: Class with specific logic for the Panther robot with version below 1.20.
```

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
macstepien: I agree, but in this case we should also change it in other places (battery, utils, roboteq_driver)

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
macstepien: It overrides method from base class, so I think that param should stay

panther_hardware_interfaces/src/gpio_controller.cpp
```diff
@@ -0,0 +1,260 @@
...
+{
+  return gpio_driver_->SetPinValue(panther_gpiod::GPIOPin::CHRG_DISABLE, !enable);
+}
+
+std::unordered_map<panther_gpiod::GPIOPin, bool>
+GPIOControllerPTH12X::QueryControlInterfaceIOStates() const
+{
+  std::unordered_map<panther_gpiod::GPIOPin, bool> io_state;
+
+  std::vector<panther_gpiod::GPIOPin> pins_to_querry = {
```
macstepien:
```suggestion
  std::vector<panther_gpiod::GPIOPin> pins_to_query = {
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
```diff
@@ -89,12 +132,15 @@ class PantherSystemRosInterface
    */
   void UpdateMsgErrors(const CANErrors & can_errors);
 
+  void PublishEStopStateMsg(const bool e_stop);
+  void PublishEStopStateIfChanged(const bool e_stop);
   void PublishDriverState();
+  void InitializeAndPublishIOStateMsg(
+    const std::unordered_map<panther_gpiod::GPIOPin, bool> & io_state);
+  void PublishIOState(const panther_gpiod::GPIOInfo & gpio_info);
+  void UpdateIOStateMsg(const panther_gpiod::GPIOPin pin, const bool pin_value);
```
macstepien: `UpdateIOStateMsg` should be private

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -107,9 +124,40 @@ CallbackReturn PantherSystem::on_activate(const rclcpp_lifecycle::State &)
...
+    "~/charger_enable",
+    std::bind(&GPIOControllerInterface::ChargerEnable, gpio_controller_, std::placeholders::_1));
+
+  panther_system_ros_interface_->AddTriggerService(
+    "~/e_stop_trigger", std::bind(&PantherSystem::SetEStop, this));
+  panther_system_ros_interface_->AddTriggerService(
+    "~/e_stop_reset", std::bind(&PantherSystem::ResetEStop, this));
+
+  const auto io_state = gpio_controller_->QueryControlInterfaceIOStates();
+  panther_system_ros_interface_->InitializeAndPublishIOStateMsg(io_state, panther_version_);
```
macstepien:
```suggestion
  panther_system_ros_interface_->InitializeAndPublishIOStateMsg(io_state);
```

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
macstepien: I think that it is needed, as we initially have to publish E-stop state regardless of its value.

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -546,6 +619,66 @@ bool PantherSystem::CheckIfSafetyStopActive()
...
+    throw std::runtime_error(
+      "Can't reset E-stop - last velocity commands are different than zero. Make sure that your "
+      "controller sends zero commands before trying to reset E-stop.");
+  }
+
+  try {
+    gpio_controller_->EStopReset();
+  } catch (const std::runtime_error & e) {
+    RCLCPP_ERROR_STREAM(logger_, "Error when trying to reset E-stop using GPIO: " << e.what());
+    throw e;
```
KmakD:
```suggestion
    throw std::runtime_error("Error when trying to reset E-stop using GPIO: " + std::string(e.what()));
```

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,318 @@
...
+   */
+  bool DigitalPowerEnable(const bool enable) override;
+
+  /**
+   * @brief Controls the charging process based on the 'enable' parameter.
+   *
+   * @param enable Set to 'true' to enable the charging process, 'false' to disable.
+   * @return 'true' if the charger control pin value is successfully set, 'false' otherwise.
+   */
+  bool ChargerEnable(const bool enable) override;
```
KmakD: I believe purpose of this pin is different. According to our documentation its: "Enable external charger", "if a non-standard charger is available, this pin allows enabling and disabling it."

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -0,0 +1,318 @@
...
+
+  /**
+   * @brief Placeholder method indicating lack of support for controlling charging process in this
+   * robot version.
+   *
+   * @param enable Ignored parameter in this version.
+   * @exception std::runtime_error Always throws a runtime error due to lack of support for charging
+   * process control.
+   */
+  bool ChargerEnable(const bool /* enable */) override;
```
KmakD:
```suggestion
  /**
   * @brief Placeholder method indicating lack of support for enabling external charger in this
   * robot version.
   *
   * @param enable Ignored parameter in this version.
   * @exception std::runtime_error Always throws a runtime error due to lack of support for charging
   * process control.
   */
  bool ChargerEnable(const bool /* enable */) override;
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
```diff
@@ -103,8 +162,14 @@ class PantherSystemRosInterface
...
-  rclcpp::Service<TriggerSrv>::SharedPtr clear_errors_srv_;
-  std::function<void()> clear_errors_;
+  rclcpp::Publisher<IOStateMsg>::SharedPtr io_state_publisher_;
+  std::unique_ptr<realtime_tools::RealtimePublisher<IOStateMsg>> realtime_io_state_publisher_;
+
+  rclcpp::Publisher<BoolMsg>::SharedPtr e_stop_state_publisher_;
+  std::unique_ptr<realtime_tools::RealtimePublisher<BoolMsg>> realtime_e_stop_state_publisher_;
+
+  std::vector<std::shared_ptr<TriggerServiceWrapper>> trigger_wrappers_;
+  std::vector<std::shared_ptr<SetBoolServiceWrapper>> set_bool_wrappers_;
```
KmakD, line 172: are this used?

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
KmakD: Still think this can be done to template. Like:
```cpp
  template <typename SrvWrapperT, typename CallbackT>
  void AddService(const std::string service_name, const CallbackT & callback);
```
and then:
```cpp
template <typename SrvWrapperT, typename CallbackT>
void PantherSystemRosInterface::AddService(
  const std::string service_name, CallbackT & callback)
{
  auto wrapper = std::make_shared<SrvWrapperT>(callback);

  wrapper->CreateService(service_name);
}
```
And each wrapper would have a method for creating service. This is assuming `set_bool_wrappers_` and `trigger_wrappers_` are not needed
And you would simply call:
```cpp
panther_system_ros_interface_->AddService<SetBoolServiceWrapper>(
    "~/motor_power_enable",
    std::bind(&GPIOControllerInterface::MotorPowerEnable, gpio_controller_, std::placeholders::_1));
```

KmakD (changes requested):

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_ros_interface.hpp
pkowalsk1: Wrappers should be stored properly to ensure that servers are not deleted and services remain enabled.

panther_hardware_interfaces/src/gpio_controller.cpp
```diff
@@ -0,0 +1,257 @@
...
+    bool is_active = gpio_driver_->IsPinActive(pin);
+    io_state.emplace(pin, is_active);
+  });
+
+  return io_state;
+}
+
+void GPIOControllerPTH10X::Start()
+{
+  gpio_driver_ = std::make_unique<panther_gpiod::GPIODriver>(gpio_config_info_storage_);
```
macstepien:
```suggestion
  gpio_driver_ = std::make_shared<panther_gpiod::GPIODriver>(gpio_config_info_storage_);
```

KmakD (approved):

pkowalsk1 (approved):

PR #223, opened 2024-02-06
ROS 2 lights converter

KmakD:
This is a basic converter and most likely its functionality may change and be extended. Right now it is hard to predict what will be required by controller and I want to avoid rewriting parts of code.
The converter has only one simple method, this may change when implementing controller that's why I've put it inside separate files.


panther_lights/include/panther_lights/led_panel.hpp
```diff
@@ -0,0 +1,42 @@
+// Copyright 2023 Husarion sp. z o.o.
```
pkowalsk1: Should we update it to 2024?

panther_lights/include/panther_lights/led_segment.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <panther_lights/animation/animation.hpp>
+
+namespace panther_lights
+{
+
+class LEDSegment
+{
+public:
+  /**
+   * @brief Virtual LED segment of the robot
```
pkowalsk1: I prefer to move this description to the class brief. In my opinion, it would be better to add short descriptions to all or most impoertant classes in this package as it becomes more complex.

panther_lights/src/led_segment.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include <cmath>
+#include <cstdint>
+#include <stdexcept>
+
+#include <yaml-cpp/yaml.h>
+
+#include <panther_lights/animation/animation.hpp>
+
+#include <iostream>
```
pkowalsk1: Is it unused?

panther_lights/include/panther_lights/led_panel.hpp
```diff
@@ -0,0 +1,42 @@
...
+{
+
+class LEDPanel
+{
+public:
+  LEDPanel(const std::size_t num_led);
+
+  ~LEDPanel() = default;
+
+  void UpdateFrame(const std::size_t iterator_first, const std::vector<std::uint8_t> & values);
```
pkowalsk1, line 42: It's just a suggestion, but maybe we could add a short description and information about exceptions here?

panther_lights/include/panther_lights/led_segment.hpp
```diff
@@ -0,0 +1,87 @@
...
+  {
+    return animation_->GetFrame(invert_led_order_);
+  }
+
+private:
+  const float controller_frequency_;
+  bool invert_led_order_ = false;
+  std::size_t channel_;
+  std::size_t first_led_iterator_;
+  std::size_t last_led_iterator_;
```
pkowalsk1, line 81: Are you sure it is an iterator? Maybe something similar to `first_led_position_`?

panther_lights/src/led_segment.cpp
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  try {
+    channel_ = segment_description["channel"].as<std::size_t>();
+  } catch (const YAML::BadConversion & e) {
+    throw std::invalid_argument("Invalid channel expression: " + std::string(e.what()));
+  }
+
+  auto led_range = segment_description["led_range"].as<std::string>();
+  std::size_t split_char = led_range.find('-');
```
pkowalsk1:
```suggestion
  const auto led_range = segment_description["led_range"].as<std::string>();
  const std::size_t split_char = led_range.find('-');
```

panther_lights/test/test_led_panel.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+class TestLEDPanel : public testing::Test
+{
+public:
+  TestLEDPanel() { led_panel_ = std::make_unique<panther_lights::LEDPanel>(num_led_); }
+  ~TestLEDPanel() {}
+
+protected:
+  void UpdateAndTestFrame(
+    const std::size_t iterator_first, const std::vector<std::uint8_t> & test_frame);
```
pkowalsk1, line 32: Same here. Is it an iterator?

panther_lights/src/led_panel.cpp
```diff
@@ -0,0 +1,46 @@
...
+namespace panther_lights
+{
+
+LEDPanel::LEDPanel(const std::size_t num_led) : num_led_(num_led)
+{
+  frame_ = std::vector<std::uint8_t>(num_led_ * 4, 0);
+}
+
+void LEDPanel::UpdateFrame(
+  const std::size_t iterator_first, const std::vector<std::uint8_t> & values)
```
pkowalsk1, line 31: Same here. Is it an iterator?

panther_lights/test/test_led_panel.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+TEST_F(TestLEDPanel, UpdateFrame)
+{
+  // full frame
+  auto test_full_frame = std::vector<std::uint8_t>(num_led_ * 4, 64);
+  UpdateAndTestFrame(0, test_full_frame);
+
+  // test updating shorter frame at different positions
+  std::size_t test_frame_size = 16;
```
pkowalsk1:
```suggestion
  const std::size_t test_frame_size = 16;
```

panther_lights/test/test_led_panel.cpp
```diff
@@ -0,0 +1,124 @@
...
+  auto test_frame = std::vector<std::uint8_t>(test_frame_size, 77);
+  UpdateAndTestFrame(0, test_frame);
+
+  // shorter test frame in the middle
+  UpdateAndTestFrame(12, test_frame);
+
+  // shorter test frame at the end of the frame
+  UpdateAndTestFrame(frame_size_ - (test_frame_size), test_frame);
```
pkowalsk1:
```suggestion
  UpdateAndTestFrame(frame_size_ - test_frame_size, test_frame);
```
Can we skip one pair of parentheses here?

panther_lights/test/test_led_panel.cpp
```diff
@@ -0,0 +1,124 @@
...
+  EXPECT_THROW(led_panel_->UpdateFrame(0, frame), std::runtime_error);
+}
+
+TEST_F(TestLEDPanel, UpdateFrameValuesOutOfRange)
+{
+  auto frame = std::vector<std::uint8_t>(num_led_ * 4, 0);
+  EXPECT_THROW(led_panel_->UpdateFrame(7, frame), std::runtime_error);
+
+  // put frame the way only last value is out of range
+  auto frame_2_size = 10;
```
pkowalsk1:
```suggestion
  const auto frame_2_size = 10;
```

panther_lights/test/test_led_segment.cpp
```diff
@@ -0,0 +1,199 @@
...
+  std::shared_ptr<panther_lights::LEDSegment> led_segment_;
+
+  const float controller_freq = 50.0;
+  const std::size_t segment_led_num_ = 10;
+};
+
+TestLEDSegment::TestLEDSegment()
+{
+  auto segment_desc =
+    YAML::Load("{led_range: 0-" + std::to_string(segment_led_num_ - 1) + ", channel: 1}");
```
pkowalsk1:
```suggestion
  const auto segment_desc =
    YAML::Load("{led_range: 0-" + std::to_string(segment_led_num_ - 1) + ", channel: 1}");
```

panther_lights/test/test_led_segment.cpp
```diff
@@ -0,0 +1,199 @@
...
+  led_segment_ = std::make_shared<panther_lights::LEDSegment>(segment_desc, controller_freq);
+}
+
+// this method may be moved to utils
```
pkowalsk1: It's a good idea. Should we do it now?

panther_lights/test/test_led_segment.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  segment_desc["led_range"] = "0-p10";
+  ExpectThrowWithDescription<std::invalid_argument>(
+    [segment_desc]() { panther_lights::LEDSegment(segment_desc, 10.0); },
+    "Error converting string to integer");
+}
+
+TEST(TestLEDSegmentInitialization, ValidDescription)
+{
+  auto segment_desc = YAML::Load("{led_range: 0-10, channel: 1}");
```
pkowalsk1:
```suggestion
  const auto segment_desc = YAML::Load("{led_range: 0-10, channel: 1}");
```

panther_lights/test/test_led_segment.cpp
```diff
@@ -0,0 +1,199 @@
...
+        EXPECT_TRUE(std::string(e.what()).find(error) != std::string::npos);
+        throw;
+      }
+    },
+    ExceptionType);
+}
+
+TEST(TestLEDSegmentInitialization, DescriptionMissingRequiredKey)
+{
+  auto segment_desc = YAML::Load("");
```
pkowalsk1, line 54: Could we move it to function like CreateSegmentDescription() in test_segment_converter.cpp?

panther_lights/test/test_segment_converter.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+TEST_F(TestSegmentConverter, ConvertInvalidLedRange)
+{
+  segments_.push_back(std::make_shared<panther_lights::LEDSegment>(
+    CreateSegmentDescription(panel_1_num_led_, panel_1_num_led_ + 10, 1), 50.0));
+  auto anim_desc = YAML::Load(
+    "{type: panther_lights::ImageAnimation, "
+    "image: $(find panther_lights)/animations/triangle01_red.png, "
+    "duration: 2}");
```
pkowalsk1: Same with this. Could we move it to function?

pkowalsk1 (changes requested):
I really like the structure of this package. I pointed out a few minor things that, in my opinion, would look better. What are we doing with the header files? "" or <>? xd

panther_lights/include/panther_lights/led_segment.hpp
KmakD: Methods from algorithm library use "iterator" naming and I was using this logic

panther_lights/test/test_led_panel.cpp
KmakD: as above

panther_lights/src/led_panel.cpp
KmakD: as above

panther_lights/test/test_led_segment.cpp
KmakD: Looked at it and it may be a bit tricky - test_utils have dependencie on rclcpp. There is no point in adding rclcpp dependencies in here. The best solution, I guess, is to devide test_utils into 2 files(for ros and non ros utils). This however makes need for changes in other repositories using current test_utils and I dont nkow if it is a good idea to make them in this PR

pkowalsk1: ok, so it's up to you

KmakD: I'll make another PR with this

pkowalsk1 (approved):

panther_utils/include/panther_utils/test/test_utils.hpp
```diff
@@ -53,6 +44,28 @@ bool CheckNaNVector(const std::vector<T> & vector)
...
+      try {
+        func();
+      } catch (const ExceptionType & e) {
+        EXPECT_TRUE(std::string(e.what()).find(error_msg) != std::string::npos);
+        throw;
+      }
+    },
+    ExceptionType);
+}
+
```
pawelirh: Not sure if wouldn't be better to use test macros directly in test files. 

In this example, we would have a function returning bool and we would test `EXPECT_TRUE(IsMessageThrown(fcn, msg))`

panther_utils/include/panther_utils/test/ros_test_utils.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include <rclcpp/rclcpp.hpp>
+
+namespace panther_utils::test_utils
+{
+
+/**
+ * @brief Wait for ROS message to arrive
+ *
+ * @param node ROS node that will be spun
```
pawelirh:
```suggestion
 * @param node ROS node that will be spin
```

panther_utils/include/panther_utils/test/ros_test_utils.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * @param msg Reference to the message, it should point to a variable,
+ * that will be overwritten by subscriber callback
+ * @param timeout timeout to wait for message to arrive
+ *
+ * @return True if message was received, false if timeout was reached
+ */
+template <typename NodeT, typename MsgT>
+bool WaitForMsg(
+  const std::shared_ptr<NodeT> & node, std::shared_ptr<MsgT> & msg,
+  const std::chrono::nanoseconds & timeout)
```
pawelirh: I would suggest milliseconds here as well, probably no need for that high time resolution.

panther_utils/test/test_test_utils.cpp
```diff
@@ -58,13 +37,26 @@ TEST(TestTestUtils, CheckNanVector)
...
+  panther_utils::test_utils::ExpectThrowWithDescription<std::invalid_argument>(
+    []() { throw std::invalid_argument("Example exception"); }, "Example exception");
+
+  panther_utils::test_utils::ExpectThrowWithDescription<std::runtime_error>(
+    []() { throw std::runtime_error("Example exception"); }, "Example");
 
-  auto run_tests = RUN_ALL_TESTS();
+  panther_utils::test_utils::ExpectThrowWithDescription<std::runtime_error>(
+    []() { throw std::runtime_error("Example exception"); }, "exception");
+}
```
pawelirh: Because of nested macros, it's quite hard to test negative results which would be nice here as well.

panther_lights/include/panther_lights/animation/animation.hpp
```diff
@@ -111,10 +112,11 @@ class Animation
         finished_ = true;
       }
 
-      return frame;
+      std::copy(frame.begin(), frame.end(), frame_.begin());
```
pawelirh: Why not `frame_ = frame`? The most important thing that we care about here is clarity - no doubt that the assignment is more convenient for readers. Additionally, we don't have a risk related to the size collision.

panther_lights/include/panther_lights/led_segment.hpp
pawelirh: I'm fine with iterator 🤓

panther_lights/src/controller_node.cpp
pawelirh, line 1: I assume, that this file is heavily under development -> no check

panther_lights/src/segment_converter.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+
+void SegmentConverter::Convert(
+  const std::vector<std::shared_ptr<LEDSegment>> & segments,
+  const std::unordered_map<std::size_t, std::shared_ptr<LEDPanel>> & panels)
+{
+  for (auto & segment : segments) {
+    try {
+      panels.at(segment->GetChannel())
+        ->UpdateFrame(segment->GetFirstLEDPosition(), segment->GetAnimationFrame());
```
pawelirh: Would be nice to split this statement for better readability.

panther_lights/src/led_segment.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+  // make sure that animation is destroyed before pluginlib loader
+  animation_.reset();
+  animation_loader_.reset();
+}
+
+void LEDSegment::SetAnimation(const YAML::Node & animation_description)
+{
+  if (!animation_description["type"]) {
+    throw std::runtime_error("Missing 'type' in animaiton description");
```
pawelirh:
```suggestion
    throw std::runtime_error("Missing 'type' in animation description");
```

panther_lights/include/panther_lights/led_panel.hpp
```diff
@@ -0,0 +1,50 @@
...
+#ifndef PANTHER_LIGHTS_LED_PANEL_HPP_
+#define PANTHER_LIGHTS_LED_PANEL_HPP_
+
+#include <cstdint>
+#include <vector>
+
+namespace panther_lights
+{
+
+class LEDPanel
```
pawelirh, line 27: Can we add a brief class description? It's easier then to recognize the component's responsibilities.

panther_lights/test/test_led_panel.cpp
```diff
@@ -0,0 +1,124 @@
...
+  // shorter test frame in the middle
+  UpdateAndTestFrame(12, test_frame);
+
+  // shorter test frame at the end of the frame
+  UpdateAndTestFrame(frame_size_ - test_frame_size, test_frame);
+
+  // frame with one element inserted at the back of the frame
+  auto frame_with_one_element = std::vector<std::uint8_t>(1, 88);
+  UpdateAndTestFrame(frame_size_ - 1, frame_with_one_element);
+}
```
pawelirh: Wouldn't be easier to split those variants for stand-alone test cases? We would no longer require the comments for them.

pawelirh (changes requested):

panther_utils/include/panther_utils/test/test_utils.hpp
KmakD: Make sense, this is used by this PR: https://github.com/husarion/panther_ros/pull/220
I'll make a merge to ros2-lights-controller and later to ros2-lights-converter, and then fix this to avoid conflicts later I guess

panther_utils/include/panther_utils/test/ros_test_utils.hpp
KmakD: I think this should be past tense, so spun is correct

panther_utils/include/panther_utils/test/ros_test_utils.hpp
KmakD: Agree, if we sleep in milliseconds, nanoseconds timeout does not make sense

panther_lights/include/panther_lights/animation/animation.hpp
KmakD: There was something not working earlier when I tried this and that why I used copy. I tested it now and it works both in tests and when displaying animations so I guess we can change this.

panther_lights/src/controller_node.cpp
KmakD: yep

panther_utils/include/panther_utils/test/ros_test_utils.hpp
pawelirh: Right!

panther_utils/test/test_test_utils.cpp
pawelirh: Please add negative tests as well, e.g. as follows:
1. Throw different exception
2. Throw with the same exception but with a mismatch in error messages
3. Do not throw from a provided function

pawelirh (approved):

PR #224, opened 2024-02-09
Ros2 diagnostics

pawelirh:
### Description

This PR incorporates the diagnostics in the `panther_battery` and 'panther_lights' packages.

### Modifications

- Added needed dependencies
- Added diagnostic updater for selected packages
- Updated dependencies in unit tests

### DIFF
--- a/panther_battery/include/panther_battery/dual_battery_publisher.hpp
+++ b/panther_battery/include/panther_battery/dual_battery_publisher.hpp
@@ -29,8 +29,9 @@ class DualBatteryPublisher : public BatteryPublisher
 {
 public:
   DualBatteryPublisher(
-    const rclcpp::Node::SharedPtr & node, const std::shared_ptr<Battery> & battery_1,
-    const std::shared_ptr<Battery> & battery_2);
+    const rclcpp::Node::SharedPtr & node,
+    std::shared_ptr<diagnostic_updater::Updater> diagnostic_updater,
+    const std::shared_ptr<Battery> & battery_1, const std::shared_ptr<Battery> & battery_2);
 
   ~DualBatteryPublisher() {}
 
@@ -39,6 +40,7 @@ protected:
   void Reset() override;
   void PublishBatteryState() override;
   void LogErrors() override;
+  void DiagnoseBattery(diagnostic_updater::DiagnosticStatusWrapper & status);
 
   BatteryStateMsg MergeBatteryMsgs(
     const BatteryStateMsg & battery_msg_1, const BatteryStateMsg & battery_msg_2);
--- a/panther_battery/include/panther_battery/dual_battery_publisher.hpp
+++ b/panther_battery/include/panther_battery/dual_battery_publisher.hpp
@@ -39,6 +39,7 @@ protected:
   void Update() override;
   void Reset() override;
   void PublishBatteryState() override;
+  void LogErrors() override;
   void DiagnoseBattery(diagnostic_updater::DiagnosticStatusWrapper & status) override;
 
   BatteryStateMsg MergeBatteryMsgs(
--- a/panther_battery/include/panther_battery/dual_battery_publisher.hpp
+++ b/panther_battery/include/panther_battery/dual_battery_publisher.hpp
@@ -30,7 +30,7 @@ class DualBatteryPublisher : public BatteryPublisher
 public:
   DualBatteryPublisher(
     const rclcpp::Node::SharedPtr & node,
-    std::shared_ptr<diagnostic_updater::Updater> diagnostic_updater,
+    const std::shared_ptr<diagnostic_updater::Updater> & diagnostic_updater,
     const std::shared_ptr<Battery> & battery_1, const std::shared_ptr<Battery> & battery_2);
 
   ~DualBatteryPublisher() {}
--- a/panther_battery/include/panther_battery/dual_battery_publisher.hpp
+++ b/panther_battery/include/panther_battery/dual_battery_publisher.hpp
@@ -39,8 +39,7 @@ protected:
   void Update() override;
   void Reset() override;
   void PublishBatteryState() override;
-  void LogErrors() override;
-  void DiagnoseBattery(diagnostic_updater::DiagnosticStatusWrapper & status);
+  void DiagnoseBattery(diagnostic_updater::DiagnosticStatusWrapper & status) override;
 
   BatteryStateMsg MergeBatteryMsgs(
     const BatteryStateMsg & battery_msg_1, const BatteryStateMsg & battery_msg_2);
--- a/panther_battery/src/battery_node.cpp
+++ b/panther_battery/src/battery_node.cpp
@@ -21,6 +21,7 @@
 #include <stdexcept>
 #include <string>
 
+#include <diagnostic_updater/diagnostic_updater.hpp>
 #include <rclcpp/rclcpp.hpp>
 
 #include <panther_battery/adc_battery.hpp>
@@ -60,10 +61,6 @@ void BatteryNode::Initialize()
     } catch (std::runtime_error & err) {
       RCLCPP_WARN(this->get_logger(), "Failed to initialize ADC Battery: %s", err.what());
       RCLCPP_INFO(this->get_logger(), "Using Roboteq drivers to publish battery data.");
-
-      diagnostic_updater_->broadcast(
-        diagnostic_msgs::msg::DiagnosticStatus::WARN,
-        "Failed to initialize ADC Battery: " + std::string(err.what()));
     }
   }
   InitializeWithRoboteqBattery();
--- a/panther_battery/src/battery_node.cpp
+++ b/panther_battery/src/battery_node.cpp
@@ -35,7 +35,8 @@ namespace panther_battery
 {
 
 BatteryNode::BatteryNode(const std::string & node_name, const rclcpp::NodeOptions & options)
-: Node(node_name, options)
+: Node(node_name, options),
+  diagnostics_updater_(std::make_shared<diagnostic_updater::Updater>(this))
 {
   this->declare_parameter<float>("panther_version", 1.2);
   this->declare_parameter<int>("ma_window_len/voltage", 10);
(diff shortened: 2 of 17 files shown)


panther_battery/include/panther_battery/dual_battery_publisher.hpp
```diff
@@ -39,6 +40,7 @@ class DualBatteryPublisher : public BatteryPublisher
   void Reset() override;
   void PublishBatteryState() override;
   void LogErrors() override;
+  void DiagnoseBattery(diagnostic_updater::DiagnosticStatusWrapper & status);
```
KmakD:
```suggestion
  void DiagnoseBattery(diagnostic_updater::DiagnosticStatusWrapper & status) override;
```

panther_battery/src/battery_node.cpp
KmakD, line 1: Maybe it will be a good idea to bradcast if battery was initialized uisng adc or motor controllers data

panther_battery/src/battery_node.cpp
```diff
@@ -58,6 +60,10 @@ void BatteryNode::Initialize()
     } catch (std::runtime_error & err) {
       RCLCPP_WARN(this->get_logger(), "Failed to initialize ADC Battery: %s", err.what());
       RCLCPP_INFO(this->get_logger(), "Using Roboteq drivers to publish battery data.");
+
+      diagnostic_updater_->broadcast(
+        diagnostic_msgs::msg::DiagnosticStatus::WARN,
+        "Failed to initialize ADC Battery: " + std::string(err.what()));
```
KmakD: I failed to get that message, logs above were present

panther_battery/src/battery_publisher.cpp
```diff
@@ -43,6 +48,10 @@ void BatteryPublisher::Publish()
   } catch (const std::runtime_error & e) {
     RCLCPP_ERROR_THROTTLE(
       node_->get_logger(), *node_->get_clock(), 1000, "Error reading battery data: %s. ", e.what());
+
+    diagnostic_updater_->broadcast(
+      diagnostic_msgs::msg::DiagnosticStatus::ERROR,
+      "Error reading battery data: " + std::string(e.what()));
```
KmakD, line 55: hardware_id is empty for this message

panther_battery/src/dual_battery_publisher.cpp
```diff
@@ -153,4 +154,30 @@ void DualBatteryPublisher::MergeBatteryPowerSupplyHealth(
   }
 }
 
+void DualBatteryPublisher::DiagnoseBattery(diagnostic_updater::DiagnosticStatusWrapper & status)
+{
+  std::vector<diagnostic_msgs::msg::KeyValue> key_values;
+  unsigned char error_level{diagnostic_updater::DiagnosticStatusWrapper::OK};
+  std::string message{"Battery has no error messages"};
+
+  if (battery_1_->HasErrorMsg() || battery_2_->HasErrorMsg()) {
```
KmakD, line 163: If there is error on only one battery wouldn't it look weird that we have key saying error and empty value?

panther_lights/src/driver_node.cpp
```diff
@@ -39,14 +39,20 @@ using std::placeholders::_1;
...
+  diagnostic_updater_(this)
 {
   rclcpp::on_shutdown(std::bind(&DriverNode::OnShutdown, this));
 
   this->declare_parameter<double>("global_brightness", 1.0);
   this->declare_parameter<double>("frame_timeout", 0.1);
   this->declare_parameter<int>("num_led", 46);
 
+  diagnostic_updater_.setHardwareID("Lights");
+  diagnostic_updater_.add("Lights status", this, &DriverNode::DiagnoseLigths);
```
KmakD:
```suggestion
  diagnostic_updater_.setHardwareID("LightsDriver");
  diagnostic_updater_.add("Lights driver status", this, &DriverNode::DiagnoseLigths);
```

panther_lights/src/driver_node.cpp
```diff
@@ -119,10 +125,15 @@ void DriverNode::FrameCB(
...
       RCLCPP_WARN_THROTTLE(
         this->get_logger(), *this->get_clock(), 5000, "%s on front panel!", meessage.c_str());
+
     } else if (panel_name == "rear") {
       RCLCPP_WARN_THROTTLE(
         this->get_logger(), *this->get_clock(), 5000, "%s on rear panel!", meessage.c_str());
     }
+
+    diagnostic_updater_.broadcast(
+      diagnostic_msgs::msg::DiagnosticStatus::WARN, "Front panel: " + meessage);
```
KmakD:
```suggestion
    diagnostic_updater_.broadcast(
      diagnostic_msgs::msg::DiagnosticStatus::WARN, panel_name + " panel: " + meessage);
```

panther_battery/src/dual_battery_publisher.cpp
```diff
@@ -153,4 +154,30 @@ void DualBatteryPublisher::MergeBatteryPowerSupplyHealth(
...
+    battery_1_kv.key = "Error message: battery 1";
+    battery_1_kv.value = battery_1_->GetErrorMsg();
+
+    diagnostic_msgs::msg::KeyValue battery_2_kv;
+    battery_2_kv.key = "Error message: battery 2";
+    battery_2_kv.value = battery_2_->GetErrorMsg();
+
+    key_values.push_back(battery_1_kv);
+    key_values.push_back(battery_2_kv);
+  }
```
KmakD: There is LogError method that checks similar conditions. I wonder if we could move contents of LogError() in here and do: `diagnostic_updater_->force_update()` to achieve the same result. This would however depend on the way how force_update works - dose it reset updater publisher timer. The idea is if publish diagnostic in 10Hz and call force_update 10Hz. Will the diagnostic be publishing 20 messages every second? 
Soooo we don't wont to publish to much diagnostic messages but if force_update resets publisher timer  then it may be usfull to move LogErrors content in here.

KmakD (changes requested):
Add include diagnostic_updater to cpp files

panther_battery/include/panther_battery/dual_battery_publisher.hpp
pawelirh: Done ✅

panther_battery/src/battery_publisher.cpp
pawelirh: The reason is the `broadcast` method. The hardware ID is available only for diagnostic tasks. 

Possible options are:
- fix this using aggregation (possibly in the future)
- use a raw publisher
- try to implement diagnostic task (cumbersome, since we would probably need dig into `Battery` implementation)

panther_lights/src/driver_node.cpp
pawelirh: Keep in mind that this is the ID of a **hardware**. What is your motivation here? 🤔

panther_lights/src/driver_node.cpp
pawelirh: Done ✅

panther_battery/src/battery_publisher.cpp
KmakD: This explanation satisfies me, no point in fixing this at this point but aggregation seems the best option

panther_lights/src/driver_node.cpp
KmakD: The lights package will have another node - lights controller, 
It make sense set Hardware ID to a real hardware component. That would be "Bumper Lights" according to glossary

panther_battery/src/dual_battery_publisher.cpp
pawelirh: We can move the stuff from `LogError` to the diagnostic routine. Together with the `throttle` logging macro, it'll not spam us with too frequent messages in the terminal.

Done ✅

panther_lights/src/driver_node.cpp
pawelirh: Changed to `Bumper Lights` then ✅

panther_battery/src/dual_battery_publisher.cpp
pawelirh: Modified ✅

panther_battery/src/battery_node.cpp
pawelirh: I dig into this problem to find a solution. In general, we can't see the broadcasted message because the logic underneath sends the provided message on **all defined tasks**. And since at this point, we don't have any task defined (the `battery_publisher` adds one) we see no message.

The workaround would be to define a new additional task at this point to enable wanted broadcasting, but once I started thinking about it, it finally seemed pointless. Diagnostic tasks after all should be used for periodic diagnostics. In case of an initialization issue relevant to the node and not to other diagnostic tasks, it seems more natural to have an `RCLCPP_INFO/`WARN` for this message. 

My proposal here is to get rid of this broadcasted message in favor of the standard ROS logging. 

Let me know if u're ok with it.

panther_lights/src/driver_node.cpp
```diff
@@ -116,13 +123,11 @@ void DriverNode::FrameCB(
...
-        this->get_logger(), *this->get_clock(), 5000, "%s on front panel!", meessage.c_str());
-    } else if (panel_name == "rear") {
-      RCLCPP_WARN_THROTTLE(
-        this->get_logger(), *this->get_clock(), 5000, "%s on rear panel!", meessage.c_str());
-    }
+    auto warn_msg = meessage + " on " + panel_name + " panel!";
+
+    diagnostic_updater_.broadcast(diagnostic_msgs::msg::DiagnosticStatus::WARN, warn_msg);
+    RCLCPP_WARN_THROTTLE(this->get_logger(), *this->get_clock(), 5000, warn_msg.c_str());
+
```
KmakD: There should be two RCLCPP_WARN_THROTTLE
The reason there were two throttle messages is, if there is error for battery 1 it will trigger log message, but if there is also error on battery 2 the throttle won't be called until 5000ms pass. So the result is we only get log that there is problem on one battery, not both.

panther_battery/src/dual_battery_publisher.cpp
KmakD: As we discussed, we go pack to using LogError

panther_lights/src/driver_node.cpp
KmakD: HardwareID is ok, and how about the name so we can later differentiate between controller and driver? Like `Lights driver status` and later in controller we would have `Lights controller status`

KmakD (changes requested):

panther_lights/src/driver_node.cpp
pawelirh: Will be fine, edited ✅

panther_battery/src/battery_node.cpp
KmakD: As we discussed, we stay with current implementation

KmakD (approved):

PR #225, opened 2024-02-12
Gazebo - fix collisions

rafal-gorecki:


panther_description/urdf/wheel.urdf.xacro
```diff
@@ -49,10 +49,11 @@
...
 
       <collision>
-        <origin xyz="0.0 0.0 0.0" rpy="0.0 0.0 0.0" />
+        <origin xyz="0.0 0.0 0.0" rpy="-${pi/2} 0.0 0.0" />
         <geometry>
-          <mesh
-            filename="package://${wheel_config['mesh_package']}/${wheel_config['folder_path']}/wheel_collision.stl" />
+          <cylinder radius="${wheel_config['wheel_radius']}" length="${wheel_config['wheel_width']}"/>
+          <!-- Simulation do not work well with mesh as collision layer. Kept just in case. -->
+          <!-- <mesh filename="package://${wheel_config['mesh_package']}/${wheel_config['folder_path']}/wheel_collision.stl" />  -->
```
pawelirh: IMO we can get rid of this comment. Using a simplified representation of a specific part for collision definition is quite natural.

panther_bringup/launch/bringup.launch.py
```diff
@@ -255,7 +255,7 @@ def generate_launch_description():
     )
 
     other_action_timer = TimerAction(
-        period=20.0,
+        period=7.0,
```
pawelirh: Can u motivate the new changes? Are they tested on hardware?

panther_lights/CMakeLists.txt
```diff
@@ -16,12 +16,20 @@ include_directories(include)
...
+         ${sensor_msgs_INCLUDE_DIRS})
 ament_target_dependencies(driver_node image_transport panther_gpiod
                           panther_msgs rclcpp sensor_msgs)
 
 add_executable(dummy_scheduler_node src/dummy_scheduler_node_main.cpp
                                     src/dummy_scheduler_node.cpp)
+target_include_directories(
+  dummy_scheduler_node
+  PUBLIC ${image_transport_INCLUDE_DIRS} ${rclcpp_INCLUDE_DIRS}
+         ${sensor_msgs_INCLUDE_DIRS})
```
pawelirh: Again, why do we modify the lights driver library linkage in the simulation task?

panther_bringup/launch/bringup.launch.py
rafal-gorecki: I change it because it taking, so long to make quick test. I didn't want to commit this thanks.

panther_lights/CMakeLists.txt
rafal-gorecki: Similar thing like above. It not build in all cases and I make few changes that's, work form me. Thanks for this.

pawelirh (approved):

PR #234, opened 2024-02-27
Ros2 lights tests

rafal-gorecki:
Add:
- APA102 tests
- Lights Driver Node tests
- renaming `front_panel` -> `chanel_1` & `rear_panel` -> `chanel_2`


panther_lights/CMakeLists.txt
```diff
@@ -164,6 +165,29 @@ if(BUILD_TESTING)
...
                             panther_msgs panther_utils rclcpp sensor_msgs)
   target_link_libraries(${PROJECT_NAME}_test_controller_node yaml-cpp)
+
+  ament_add_gtest(${PROJECT_NAME}_test_apa102 test/test_apa102.cpp
+                  src/apa102.cpp)
+  target_include_directories(
+    ${PROJECT_NAME}_test_apa102
+    PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+           $<INSTALL_INTERFACE:include>)
+  ament_target_dependencies(${PROJECT_NAME}_test_apa102)
```
KmakD:
```suggestion
```
Not needed here

panther_utils/include/panther_utils/test/ros_test_utils.hpp
```diff
@@ -28,7 +28,7 @@ namespace panther_utils::test_utils
 /**
  * @brief Wait for ROS message to arrive
  *
- * @param node ROS node that will be spun
+ * @param node ROS node that will be spin
```
KmakD: I think "spun" was correct

rafal-gorecki:
1. There is one error with the `Publish Success` test that I do not understand. For some reason, when this test is performed before other tests, the error shown in the screenshot below appears.

image

I don't see any reason why this would happen because with each test all objects in the `DriverNodeFixture` class are destroyed.

2. I also added this change:

```cpp
  if (panels_initialised_) {
    SetPowerPin(false);
  }
```

I add it, because it also caused an error similar to the one above. I don't know yet whether I should replace `true` -> `false` in this line

```cpp
  std::vector<panther_gpiod::GPIOInfo> gpio_info_storage = {panther_gpiod::GPIOInfo{
    panther_gpiod::GPIOPin::LED_SBC_SEL, gpiod::line::direction::OUTPUT, true}};
```

3. It shouldn't be, but it would be good if you ran this test once (you can use `panther-81b2-rpi:~/rafal_ws$ colcon test --packages-select panther_lights`) to check if the panther doesn't light up for a split second like a Christmas tree during the tests.

panther_utils/include/panther_utils/test/ros_test_utils.hpp
```diff
@@ -38,7 +38,7 @@ namespace panther_utils::test_utils
 template <typename NodeT, typename MsgT>
 bool WaitForMsg(
   const std::shared_ptr<NodeT> & node, std::shared_ptr<MsgT> & msg,
-  const std::chrono::milliseconds & timeout)
+  const std::chrono::nanoseconds & timeout)
```
KmakD: Should be milliseconds - later this method uses sleep in milliseconds so timeout in nanoseconds does not make sense.

panther_utils/include/panther_utils/test/ros_test_utils.hpp
```diff
@@ -53,6 +53,30 @@ bool WaitForMsg(
...
+ *
+ * @param node ROS node that will be spin
+ * @param future Reference to the service future, it should point to a variable,
+ * that will be overwritten after service response
+ * @param timeout timeout to wait for service response
+ *
+ * @return True if message was received, false if timeout was reached
+ */
+template <typename NodeT, typename FutureT>
+bool WaitForFuture(const NodeT & node, FutureT & future, const std::chrono::nanoseconds & timeout)
```
KmakD:
```suggestion
bool WaitForFuture(const NodeT & node, FutureT & future, const std::chrono::milliseconds & timeout)
```
Again milliseconds should be good enough

panther_utils/include/panther_utils/test/ros_test_utils.hpp
```diff
@@ -53,6 +53,30 @@ bool WaitForMsg(
   return false;
 }
 
+/**
+ * @brief Wait for ROS service future to arrive
+ *
+ * @param node ROS node that will be spin
```
KmakD:
```suggestion
 * @param node ROS node that will be spun
```

panther_utils/include/panther_utils/test/ros_test_utils.hpp
```diff
@@ -53,6 +53,30 @@ bool WaitForMsg(
...
+ * @return True if message was received, false if timeout was reached
+ */
+template <typename NodeT, typename FutureT>
+bool WaitForFuture(const NodeT & node, FutureT & future, const std::chrono::nanoseconds & timeout)
+{
+  rclcpp::Time start_time = node->now();
+
+  while (rclcpp::ok() && node->now() - start_time <= rclcpp::Duration(timeout)) {
+    rclcpp::spin_some(node->get_node_base_interface());
+    if (future.wait_for(std::chrono::microseconds(10)) == std::future_status::ready) {
```
KmakD:
```suggestion
    if (future.wait_for(std::chrono::milliseconds(10)) == std::future_status::ready) {
```

panther_lights/include/panther_lights/test/mock_apa102.hpp
```diff
@@ -0,0 +1,38 @@
...
+#ifndef PANTHER_LIGHTS__MOCK_APA102_HPP_
+#define PANTHER_LIGHTS__MOCK_APA102_HPP_
+
+#include <panther_lights/apa102.hpp>
+
+namespace panther_lights::mock_apa102
+{
+
+// Mock the SPI communication functions to isolate testing
+class MockAPA102 : public panther_lights::apa102::APA102
```
KmakD: This looks more like a wrapper tbh. Maybe move it to test directly similar to other test files in this package. For reference: https://github.com/husarion/panther_ros/blob/7d216e0d0f84d58307553acb9f808e5860465064/panther_lights/test/test_charging_animation.cpp#L26

panther_lights/include/panther_lights/test/mock_driver_node.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#ifndef PANTHER_LIGHTS__MOCK_DRIVER_NODE_HPP_
+#define PANTHER_LIGHTS__MOCK_DRIVER_NODE_HPP_
+
+#include <panther_lights/driver_node.hpp>
+
+namespace panther_lights::mock_driver_node
+{
+
+class MockDriverNode : public panther_lights::DriverNode::DriverNode
```
KmakD: Same as above

panther_lights/include/panther_lights/test/mock_driver_node.hpp
```diff
@@ -0,0 +1,37 @@
...
+{
+public:
+  MockDriverNode(const std::string & device) : DriverNode(device) {}
```
KmakD:
```suggestion
class MockDriverNode : public panther_lights::DriverNode
{
public:
  MockDriverNode(const std::string & node_name, const rclcpp::NodeOptions & options = rclcpp::NodeOptions()) : DriverNode(node_name, options) {}
```

panther_lights/test/test_apa102.cpp
KmakD, line 1: Have you tried to test SPISendBuffer? Reading from SPI would look sth like:
```cpp
  struct spi_ioc_transfer tr;
  memset(&tr, 0, sizeof(tr));
  tr.tx_buf = 0;
  tr.rx_buf = reinterpret_cast<std::uint64_t>(buffer.data());
  tr.len = static_cast<std::uint32_t>(buffer.size());
  tr.speed_hz = speed_;
  tr.delay_usecs = 0;
  tr.bits_per_word = kBits;

const int ret = ioctl(fd_, SPI_IOC_MESSAGE(1), &tr);
```

panther_lights/test/test_driver_node.cpp
```diff
@@ -0,0 +1,157 @@
...
+    channel_1_pub_ = std::make_shared<image_transport::Publisher>(
+      it_->advertise("lights/driver/channel_1_frame", 5));
+    channel_2_pub_ = std::make_shared<image_transport::Publisher>(
+      it_->advertise("lights/driver/channel_2_frame", 5));
+    set_brightness_client_ =
+      node_->create_client<SetLEDBrightnessSrv>("lights/driver/set/brightness");
+    node_->Initialize();
+  }
+
+  void TearDown() override { rclcpp::shutdown(); }
```
KmakD: Is there a reason to use SetUp and TearDown instead of constructor and deconstructor?

panther_lights/test/test_driver_node.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+protected:
+  ImageMsg::SharedPtr CreateImageMsg()
+  {
+    auto msg = std::make_shared<ImageMsg>();
+
+    // Filling msg with dummy data
+    msg->header.stamp = node_->now();
+    msg->header.frame_id = "camera_frame";
```
KmakD:
```suggestion
    msg->header.frame_id = "image_frame";
```

panther_lights/test/test_apa102.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <panther_lights/test/mock_apa102.hpp>
+
+using MockAPA102 = panther_lights::mock_apa102::MockAPA102;
+
+class APA102Test : public ::testing::Test
+{
+protected:
+  void SetUp() override { apa102_ = std::make_unique<MockAPA102>("/dev/spidev0.0"); }
+
+  void TearDown() override { apa102_.reset(); }
```
KmakD: Is there a reason to use SetUp and TearDown instead of constructor and deconstructor?

panther_lights/test/test_apa102.cpp
```diff
@@ -0,0 +1,97 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+#include <panther_lights/test/mock_apa102.hpp>
+
+using MockAPA102 = panther_lights::mock_apa102::MockAPA102;
+
+class APA102Test : public ::testing::Test
```
KmakD:
```suggestion
class TestAPA102 : public testing::Test
```
Just to be compatible with other tests

panther_lights/test/test_driver_node.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+#include <panther_lights/test/mock_driver_node.hpp>
+#include <panther_msgs/srv/set_led_brightness.hpp>
+#include <panther_utils/test/ros_test_utils.hpp>
+
+using ImageMsg = sensor_msgs::msg::Image;
+using SetLEDBrightnessSrv = panther_msgs::srv::SetLEDBrightness;
+using MockDriverNode = panther_lights::mock_driver_node::MockDriverNode;
+
+class DriverNodeFixture : public ::testing::Test
```
KmakD:
```suggestion
class TestDriverNode : public testing::Test
```
same

panther_lights/test/test_driver_node.cpp
```diff
@@ -0,0 +1,157 @@
...
+  EXPECT_FALSE(node_->isInitialised());
+
+  // Older Msg Fail
+  msg = CreateImageMsg();
+  node_->setChanel1TS(msg->header.stamp);
+  msg->header.stamp.nanosec--;
+  channel_1_pub_->publish(msg);
+  rclcpp::spin_some(node_->get_node_base_interface());
+  EXPECT_FALSE(node_->isInitialised());
+}
```
KmakD: This should be separate tests

panther_lights/test/test_driver_node.cpp
```diff
@@ -0,0 +1,157 @@
...
+  EXPECT_TRUE(node_->isInitialised());
+}
+
+int main(int argc, char ** argv)
+{
+  // rclcpp::init(argc, argv);
+  testing::InitGoogleTest(&argc, argv);
+  int result = RUN_ALL_TESTS();
+  // rclcpp::shutdown();
+  return result;
```
KmakD:
```suggestion
  testing::InitGoogleTest(&argc, argv);
  return RUN_ALL_TESTS();
```

panther_lights/include/panther_lights/driver_node.hpp
```diff
@@ -42,6 +42,16 @@ class DriverNode : public rclcpp::Node
 
   void Initialize();
 
+protected:
+  void SetPowerPin(const bool value);
```
KmakD: Is this method used in test?

KmakD (changes requested):

panther_lights/test/test_apa102.cpp
rafal-gorecki: In apa102 driver datasheets I don't see any information if the driver sand any ack of reviving the data.

KmakD (approved):

PR #236, opened 2024-03-01
Ros2 control imu hardware

delihus:
Moved IMU driver from `spatial_driver` and `imu_madgwick_filter` to sensor hardware interface in` ros2_control`.
IMU data is published by `imu_sensor_broadcaster` with covariances described in `spatial_driver` and `imu_madgwick_filter`.

Tested:
- loading hardware_interface,
- initial values of state interfaces,
- read function.

To test @pkowalsk1 :
- plug and unplug the device,
- realtime capabilities.


panther_controller/config/WH01_controller.yaml
```diff
@@ -14,8 +14,13 @@
...
       frame_id: imu_link
-      static_covariance_angular_velocity: [1.1e-4, 0.0, 0.0, 0.0, 1.1e-4, 0.0, 0.0, 0.0, 1.1e-4]  # Values measured experimentally
-      static_covariance_linear_acceleration: [6.2e-6, 0.0, 0.0, 0.0, 6.2e-6, 0.0, 0.0, 0.0, 6.2e-6] # Values measured experimentally
+
+      # orientation_stddev: 1.0e-2 for madgwick filter
+      static_covariance_orientation: [1.0e-4, 0.0, 0.0, 0.0, 1.0e-4, 0.0, 0.0, 0.0, 1.0e-4]
+      # angular_velocity_stdev: 0.095 deg/s gyroscope white noise sigma, as per manual
+      static_covariance_angular_velocity: [2.76e-6, 0.0, 0.0, 0.0, 2.76e-6, 0.0, 0.0, 0.0, 2.76e-6]
+      # linear_acceleration_stdev: 280 ug accelerometer white noise sigma, as per manual
+      static_covariance_linear_acceleration: [7.56e-6, 0.0, 0.0, 0.0, 7.56e-6, 0.0, 0.0, 0.0, 7.56e-6]
```
pkowalsk1: Is `static_covariance_angular_velocity` and `static_covariance_linear_acceleration` related to WH01? I hope they are connected to the Panther robot's wheel odometry.

panther_controller/config/WH02_controller.yaml
```diff
@@ -14,8 +14,13 @@
...
       frame_id: imu_link
-      static_covariance_angular_velocity: [1.1e-4, 0.0, 0.0, 0.0, 1.1e-4, 0.0, 0.0, 0.0, 1.1e-4]  # Values measured experimentally
-      static_covariance_linear_acceleration: [6.2e-6, 0.0, 0.0, 0.0, 6.2e-6, 0.0, 0.0, 0.0, 6.2e-6] # Values measured experimentally
+
+      # orientation_stddev: 1.0e-2 for madgwick filter
+      static_covariance_orientation: [1.0e-4, 0.0, 0.0, 0.0, 1.0e-4, 0.0, 0.0, 0.0, 1.0e-4]
+      # angular_velocity_stdev: 0.095 deg/s gyroscope white noise sigma, as per manual
+      static_covariance_angular_velocity: [2.76e-6, 0.0, 0.0, 0.0, 2.76e-6, 0.0, 0.0, 0.0, 2.76e-6]
+      # linear_acceleration_stdev: 280 ug accelerometer white noise sigma, as per manual
+      static_covariance_linear_acceleration: [7.56e-6, 0.0, 0.0, 0.0, 7.56e-6, 0.0, 0.0, 0.0, 7.56e-6]
```
pkowalsk1: Same here

panther_controller/config/WH04_controller.yaml
```diff
@@ -14,8 +14,13 @@
...
       frame_id: imu_link
-      static_covariance_angular_velocity: [1.1e-4, 0.0, 0.0, 0.0, 1.1e-4, 0.0, 0.0, 0.0, 1.1e-4]  # Values measured experimentally
-      static_covariance_linear_acceleration: [6.2e-6, 0.0, 0.0, 0.0, 6.2e-6, 0.0, 0.0, 0.0, 6.2e-6] # Values measured experimentally
+
+      # orientation_stddev: 1.0e-2 for madgwick filter
+      static_covariance_orientation: [1.0e-4, 0.0, 0.0, 0.0, 1.0e-4, 0.0, 0.0, 0.0, 1.0e-4]
+      # angular_velocity_stdev: 0.095 deg/s gyroscope white noise sigma, as per manual
+      static_covariance_angular_velocity: [2.76e-6, 0.0, 0.0, 0.0, 2.76e-6, 0.0, 0.0, 0.0, 2.76e-6]
+      # linear_acceleration_stdev: 280 ug accelerometer white noise sigma, as per manual
+      static_covariance_linear_acceleration: [7.56e-6, 0.0, 0.0, 0.0, 7.56e-6, 0.0, 0.0, 0.0, 7.56e-6]
```
pkowalsk1: Same here

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,131 @@
...
+  CallbackReturn on_activate(const rclcpp_lifecycle::State & previous_state) override;
+  CallbackReturn on_deactivate(const rclcpp_lifecycle::State & previous_state) override;
+  CallbackReturn on_shutdown(const rclcpp_lifecycle::State & previous_state) override;
+  CallbackReturn on_error(const rclcpp_lifecycle::State & previous_state) override;
+
+  std::vector<StateInterface> export_state_interfaces() override;
+
+  return_type read(const rclcpp::Time & /* time */, const rclcpp::Duration & /* period */) override;
+
+protected:
```
pkowalsk1: We adhere to the convention of placing methods at the beginning of the class definition and objects at the end. 

Is it necessary to make all members protected? In my opinion, some of these could be designated as private.

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,131 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef PANTHER_HARDWARE_INTERFACES_PANTHER_IMU_HPP_
+#define PANTHER_HARDWARE_INTERFACES_PANTHER_IMU_HPP_
+
+#include <array>
+#include <memory>
+#include <string>
+#include <vector>
```
pkowalsk1:
```suggestion
#include <array>
#include <memory>
#include <mutex>
#include <string>
#include <vector>
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+  static constexpr size_t kImuInterfacesSize = 10;
+  static constexpr double KImuMagneticFieldUnknownValue = 1e300;
+  static constexpr float G = 9.80665;
+  inline static std::string kImuSensorName = "imu";
+  inline static std::array<std::string, kImuInterfacesSize> kImuInterfacesNames = {
+    "orientation.x",         "orientation.y",         "orientation.z",      "orientation.w",
+    "angular_velocity.x",    "angular_velocity.y",    "angular_velocity.z", "linear_acceleration.x",
+    "linear_acceleration.y", "linear_acceleration.z",
+  };
```
pkowalsk1:
```suggestion
  inline static const std::string kImuSensorName = "imu";
  inline static const std::array<std::string, kImuInterfacesSize> kImuInterfacesNames = {
    "orientation.x",         "orientation.y",         "orientation.z",      "orientation.w",
    "angular_velocity.x",    "angular_velocity.y",    "angular_velocity.z", "linear_acceleration.x",
    "linear_acceleration.y", "linear_acceleration.z",
  };
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+  phidgets_spatial::Params params_;
+  std::unique_ptr<phidgets::Spatial> spatial_;
+  std::mutex spatial_mutex_;
+
+  bool imu_connected_;
+  ImuFilter filter_;
+  WorldFrame::WorldFrame world_frame_;
+  bool first_data_callback_{true};
+  double last_spatial_data_callback_time_s_;
```
pkowalsk1:
```suggestion
  ImuFilter filter_;
  WorldFrame::WorldFrame world_frame_;
  bool imu_connected_;
  bool first_data_callback_ = true;
  double last_spatial_data_callback_time_s_;
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+  void CheckSensor() const;
+  void CheckStatesSize() const;
+  void SetInitialValues();
+  void CheckInterfaces();
```
pkowalsk1:
```suggestion
  void CheckInterfaces() const;
```
Is it const?

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,131 @@
...
+  void ReadObligatoryParams();
+  void ReadCompassParams();
+  void ReadMadgwickFilterParams();
+  void CheckMadgwickFilterWorldFrameParam();
+  bool IsParamDefined(const std::string & param_name);
```
pkowalsk1:
```suggestion
  bool IsParamDefined(const std::string & param_name) const;
```
Is it const?

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+  void ConfigureCompassParams();
+  void ConfigureHeating();
+  void ConfigureMadgwickFilter();
+
+  void SpatialDataCallback(
+    const double acceleration[3], const double angular_rate[3], const double magnetic_field[3],
+    double timestamp);
```
pkowalsk1:
```suggestion
  void SpatialDataCallback(
    const double acceleration[3], const double angular_rate[3], const double magnetic_field[3],
    const double timestamp);
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,131 @@
...
+  void SpatialAttachCallback();
+  void SpatialDetachCallback();
+
+  void UpdateMadgwickAlgorithm(
+    const geometry_msgs::msg::Vector3 & ang_vel, const geometry_msgs::msg::Vector3 & lin_acc,
+    const geometry_msgs::msg::Vector3 & mag_compensated, double dt);
```
pkowalsk1:
```suggestion
  void UpdateMadgwickAlgorithm(
    const geometry_msgs::msg::Vector3 & ang_vel, const geometry_msgs::msg::Vector3 & lin_acc,
    const geometry_msgs::msg::Vector3 & mag_compensated, const double dt);
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp
```diff
@@ -0,0 +1,77 @@
...
+ * @brief Class that takes care of additional ROS interface of panther imu, such as providing service for calibration
+ */
+class PantherImuRosInterface
+{
+public:
+  /**
+   * @brief Creates node and executor (in a separate thread), publishers, subscribers and services
+   *
+   * @param calibrate - functions that should be called, when calibration is triggered
+   * service is called
```
pkowalsk1:
```suggestion
    * @param calibrate - function to be called when the calibration service is triggered
```
Is it the proper meaning of this sentence?

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp
```diff
@@ -0,0 +1,77 @@
...
+   */
+  phidgets_spatial::Params &GetParams();
+
+private:
+  void CalibrateCb(
+    TriggerSrv::Request::ConstSharedPtr /* request */, TriggerSrv::Response::SharedPtr response);
+
+  rclcpp::Node::SharedPtr node_;
+  rclcpp::executors::SingleThreadedExecutor::UniquePtr executor_;
+  std::unique_ptr<std::thread> executor_thread_;
```
pkowalsk1: Is it a need to make it a pointer?

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+  rclcpp::Service<TriggerSrv>::SharedPtr calibrate_srv_;
+  std::function<void()> calibrate_;
+
+  std::shared_ptr<phidgets_spatial::ParamListener> param_listener_;
```
pkowalsk1: Same here.  Is there a necessity to make this a pointer, and if that is the case, should it not be a std::unique_ptr?

panther_hardware_interfaces/include/panther_hardware_interfaces/phidgets_spatial_driver.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <atomic>
+#include <mutex>
+
+#include <phidgets_api/spatial.hpp>
+
+
+namespace panther_hardware_interfaces
+{
+
+class PhidgetsSpatialDriver{
```
pkowalsk1: I'm unclear about the purpose of this file. Could you please explain its intended use?

panther_hardware_interfaces/include/panther_hardware_interfaces/phidgets_spatial_driver.hpp
```diff
@@ -0,0 +1,37 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef PANTHER_HARDWARE_INTERFACES_PHIDGETS_SPATIAL_DRIVER_HPP_
+#define PANTHER_HARDWARE_INTERFACES_PHIDGETS_SPATIAL_DRIVER_HPP_
+
+#include <atomic>
+#include <mutex>
```
pkowalsk1:
```suggestion
#include <atomic>
#include <mutex>
```
```suggestion
#include <mutex>
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,131 @@
...
+  std::mutex spatial_mutex_;
+
+  bool imu_connected_;
+  ImuFilter filter_;
+  WorldFrame::WorldFrame world_frame_;
+  bool first_data_callback_{true};
+  double last_spatial_data_callback_time_s_;
+
+  void CheckSensor() const;
+  void CheckStatesSize() const;
```
pkowalsk1: These methods, among others, throw exceptions. Perhaps adding brief descriptions would be a good idea to clarify their behavior?

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+#include <functional>
+#include <limits>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include <rclcpp/logging.hpp>
+
+#include <hardware_interface/types/hardware_interface_type_values.hpp>
```
pkowalsk1: In cases such as this and others, I prefer to use double quotes `""` instead of angle brackets `<>`.
https://app.clickup.com/t/8693r8b21

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+  try {
+    ReadObligatoryParams();
+  } catch (const std::invalid_argument & e) {
+    RCLCPP_FATAL_STREAM(logger_, "Exception during reading obligatory parameters: " << e.what());
+    return CallbackReturn::ERROR;
+  } catch (const std::runtime_error & e) {
+    RCLCPP_FATAL_STREAM(
+      logger_, "Exception during interpreting obligatory parameters: " << e.what());
+    return CallbackReturn::ERROR;
+  }
```
pkowalsk1: I suggest implementing only a single catch block. In my opinion, providing detailed information to the user like this isn't worthwhile.

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+    return CallbackReturn::ERROR;
+  } catch (const std::invalid_argument & e) {
+    RCLCPP_FATAL_STREAM(
+      logger_, "Exception during reading Madgwick Filter parameters: " << e.what());
+    return CallbackReturn::ERROR;
+  } catch (const std::runtime_error & e) {
+    RCLCPP_FATAL_STREAM(
+      logger_, "Exception during interpreting Madgwick Filter parameters: " << e.what());
+    return CallbackReturn::ERROR;
+  }
```
pkowalsk1: Similar to the previous case, in my opinion, one or at most two catch statements should be sufficient.

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+}
+
+CallbackReturn PantherImuSensor::on_activate(const rclcpp_lifecycle::State &)
+{
+  RCLCPP_INFO(logger_, "Activating Panther Imu");
+  rclcpp::NodeOptions ros_interface_options;
+  panther_imu_ros_interface_ = std::make_unique<PantherImuRosInterface>(
+    std::bind(&PantherImuSensor::Calibrate, this),
+    "panther_imu_"
+    "node");
```
pkowalsk1: Could you confirm if `PantherImuRosInterface` and `PantherImuSensor` are being properly deleted at the end of the program? I'm asking because we faced a similar issue with `PantherSystem`. 😃

Why this string is formatted this way?

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+  RCLCPP_INFO_STREAM(
+    logger_, "State interfaces count: " << info_.sensors[0].state_interfaces.size());
+
+  for (auto i = 0u; i < info_.sensors[0].state_interfaces.size(); i++) {
+    RCLCPP_INFO_STREAM(
+      logger_, "State interface name: " << info_.sensors[0].state_interfaces[i].name);
+
+    state_interfaces.emplace_back(StateInterface(
+      info_.sensors[0].name, info_.sensors[0].state_interfaces[i].name, &imu_sensor_state_[i]));
+  }
```
pkowalsk1:
```suggestion
  for (size_t i = 0; i < info_.sensors[0].state_interfaces.size(); i++) {
    state_interfaces.emplace_back(
      info_.sensors[0].name, info_.sensors[0].state_interfaces[i].name, &imu_sensor_state_[i]);
  }
```
Could we also use kImuInterfacesSize instead of `kImuInterfacesSize`

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+{
+  auto names_iter = kImuInterfacesNames.begin();
+  auto states_iter = info_.sensors[0].state_interfaces.begin();
+  auto compare_name_with_interface_info =
+    [](const std::string & name, const hardware_interface::InterfaceInfo & interface_info) {
+      return name == interface_info.name;
+    };
+
+  if (not std::equal(
+        names_iter, kImuInterfacesNames.end(), states_iter, compare_name_with_interface_info)) {
```
pkowalsk1: Why `names_iter` is local variable and `kImuInterfacesNames.end()` is not?

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+  params_.cc_offset2 = hardware_interface::stod(cc_offset2);
+  params_.cc_gain0 = hardware_interface::stod(cc_gain0);
+  params_.cc_gain1 = hardware_interface::stod(cc_gain1);
+  params_.cc_gain2 = hardware_interface::stod(cc_gain2);
+  params_.cc_t0 = hardware_interface::stod(cc_t0);
+  params_.cc_t1 = hardware_interface::stod(cc_t1);
+  params_.cc_t2 = hardware_interface::stod(cc_t2);
+  params_.cc_t3 = hardware_interface::stod(cc_t3);
+  params_.cc_t4 = hardware_interface::stod(cc_t4);
+  params_.cc_t5 = hardware_interface::stod(cc_t5);
```
pkowalsk1: Perhaps this approach wouldn't be too bad:
```c++
params_.cc_mag_field = hardware_interface::stod(
    info_.hardware_parameters.at("cc_mag_field"));
```

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+}
+
+void PantherImuSensor::ReadMadgwickFilterParams()
+{
+  const auto gain = info_.hardware_parameters.at("gain");
```
pkowalsk1: Same here

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+  params_.use_mag = hardware_interface::parse_bool(use_mag);
+  params_.stateless = hardware_interface::parse_bool(stateless);
+  params_.remove_gravity_vector = hardware_interface::parse_bool(remove_gravity_vector);
+}
+
+void PantherImuSensor::CheckMadgwickFilterWorldFrameParam()
+{
+  const auto world_frame = info_.hardware_parameters.at("world_frame");
+
+  auto it = kImuWorldFramesMap.find(world_frame);
```
pkowalsk1: Using a map for this class member seems like overkill in my opinion.

I would even suggest a simpler approach such as:
```c++
if (world_frame == "ned") world_frame_  = WorldFrame::NED;
if (world_frame == "nwu") world_frame_  = WorldFrame::NWU;
if (world_frame == "enu") world_frame_  = WorldFrame::ENU;
```

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+  }
+
+  spatial_->zero();
+  // The API call returns directly, so we "enforce" the recommended 2 sec
+  // here. See: https://github.com/ros-drivers/phidgets_drivers/issues/40
+
+  // FIXME: Ideally we'd use an rclcpp method that honors use_sim_time here,
+  // but that doesn't actually exist.  Once
+  // https://github.com/ros2/rclcpp/issues/465 is solved, we can revisit this.
+  std::this_thread::sleep_for(std::chrono::seconds(2));
```
pkowalsk1: Perhaps we could utilize concurrency utilities such as `std::promise` or `std::condition_variable` to eliminate the need for `sleep_for`?

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+bool PantherImuSensor::IsParamDefined(const std::string & param_name)
+{
+  return info_.hardware_parameters.find(param_name) != info_.hardware_parameters.end();
+}
+
+void PantherImuSensor::ConfigureCompassParams()
+{
+  if (
+    IsParamDefined("cc_mag_field") and IsParamDefined("cc_offset0") and
```
pkowalsk1: Could we either overload `IsParamDefined` or introduce `AreParamsDefined` to check multiple parameters simultaneously, for instance, by passing an `std::unordered_set` or a similar container?

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+void PantherImuSensor::ConfigureMadgwickFilter()
+{
+  ReadMadgwickFilterParams();
+
+  filter_.setWorldFrame(world_frame_);
+  filter_.setAlgorithmGain(params_.gain);
+  filter_.setDriftBiasGain(params_.zeta);
+}
+
+void PantherImuSensor::SpatialDataCallback(
```
pkowalsk1: Could we refactor this method to be shorter by introducing additional smaller methods?

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+#include <panther_hardware_interfaces/panther_imu.hpp>
+
+#include <array>
+#include <chrono>
+#include <functional>
+#include <limits>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <vector>
```
pkowalsk1:
```suggestion
#include <array>
#include <chrono>
#include <cmath>
#include <functional>
#include <limits>
#include <memory>
#include <stdexcept>
#include <string>
#include <vector>
```
```suggestion
#include <array>
#include <chrono>
#include <functional>
#include <limits>
#include <memory>
#include <stdexcept>
#include <string>
#include <vector>
```

panther_hardware_interfaces/src/phidgets_spatial_parameters.yaml
```diff
@@ -0,0 +1,144 @@
+# https://github.com/ros-drivers/phidgets_drivers/blob/humble/phidgets_spatial/README.md
```
pkowalsk1: Is it a common practice to store .yaml files in the src directory? Would it be better to create a config or params directory for them instead?

pkowalsk1 (changes requested):
The overall architecture of these changes looks good, but there are some minor adjustments that need to be made. Additionally, a review of tests and tests with real robots remains, which I will conduct in the meantime.

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+  world_frame_ = it->second;
+}
+
+void PantherImuSensor::SetInitialValues()
+{
+  imu_sensor_state_.resize(
+    info_.sensors[0].state_interfaces.size(), std::numeric_limits<double>::quiet_NaN());
+}
+
+void PantherImuSensor::Calibrate()
```
pkowalsk1, line 286: Should we block the read functionality when the calibration callback is executed, or is there something else we should do?

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+}
+
+void PantherImuSensor::SpatialDataCallback(
+  const double acceleration[3], const double angular_rate[3], const double magnetic_field[3],
+  double timestamp)
+{
+  auto timestamp_s = timestamp * 1e-3;
+  RCLCPP_DEBUG(logger_, "SpatialDataCallback...");
+
+  std::lock_guard<std::mutex> lock(spatial_mutex_);
```
pkowalsk1: In my opinion, lock_guard and mutex are not necessary.

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+  mag_fld.x = magnetic_field[0] * 1e-4;
+  mag_fld.y = magnetic_field[1] * 1e-4;
+  mag_fld.z = magnetic_field[2] * 1e-4;
+
+  for (auto i = 0u; i < 3; ++i) {
+    if (magnetic_field[i] == KImuMagneticFieldUnknownValue) {
+      mag_fld.x = std::numeric_limits<double>::quiet_NaN();
+    }
+  }
```
pkowalsk1: What is the purpose of setting only the x value to NaN? Why do we perform this operation, and then immediately check if x is finite in the following lines?

pkowalsk1 (changes requested):

panther_controller/config/WH01_controller.yaml
delihus: These covariances describe only the `orientation`, `angular_velocity` and `linear_acceleration` of the IMU. It is not connected with the robot's wheel odometry even the type of a wheel. It only depends on a IMU device. These values are included in the `sensor_msgs/msg/Imu`.

panther_controller/config/WH02_controller.yaml
delihus: I answered in the first comment.

panther_controller/config/WH04_controller.yaml
delihus: I answered in the first comment.

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
delihus: Okay!

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
delihus: It should be.

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
delihus: It should be.

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp
delihus: Yup this sounds weird

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp
delihus: The thread should be started after the constructor of the executor and after adding the node to the executor.

```cpp
PantherImuRosInterface::PantherImuRosInterface(...){
  node_ = std::make_shared<rclcpp::Node>(node_name, node_options);
  executor_ = std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
  executor_->add_node(node_);

  executor_thread_ = std::make_unique<std::thread>([this]() { executor_->spin(); });
  ...
}
```
That's why I used the pointer.

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp
delihus: Like above. The `param_listener_` needs a created node.
```cpp
  param_listener_ = std::make_shared<phidgets_spatial::ParamListener>(node_->get_node_parameters_interface());
  ```

File: panther_hardware_interfaces/include/panther_hardware_interfaces/phidgets_spatial_driver.hpp (hunk 15 above)
Comment (delihus) [reply panther_hardware_interfaces/include/panther_hardware_interfaces/phidgets_spatial_driver.hpp]: Ohh... I see now. It's a mess.

File: panther_hardware_interfaces/include/panther_hardware_interfaces/phidgets_spatial_driver.hpp (hunk 16 above)
Comment (delihus) [reply panther_hardware_interfaces/include/panther_hardware_interfaces/phidgets_spatial_driver.hpp]: deleted

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 18 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp]: Let's leave this topic until we've proposed some rules regarding file inclusion.

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 21 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp]: Formatted fixed. I have to think how to check that.

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 22 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp]: I don't get it. Did you mean to use `kImuInterfacesSize` instead of `info_.sensors[0].state_interfaces.size()`?

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 27 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp]: I have to check if the calibration zeros the gyration if so we can wait until the gyration is zero.

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 28 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp]: Great idea!

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 29 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp]: I do think so. I will try!

File: panther_hardware_interfaces/src/phidgets_spatial_parameters.yaml (hunk 31 above)
Comment (delihus) [reply panther_hardware_interfaces/src/phidgets_spatial_parameters.yaml]: The ros2_control controllers store the yaml files inside `src/` folder for building the structure of params.  It is like the source file for `generate_parameters_library` but I can move it to the `config` folder.

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 32 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp:286]: We decided to calibrate the sensor only before reading the sensor. The calibration service is removed so there is no need to block this.

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 33 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp]: Yup there is no other lock_guard

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 34 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp]: Ohhhh.. It is a mistake! All values should be nan. Thank you for the catch.

File: panther_controller/config/WH01_controller.yaml (hunk 1 above)
Comment (pkowalsk1) [reply panther_controller/config/WH01_controller.yaml]: Nevermind, I was looking in the wrong way :p

File: panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp (hunk 13 above)
Comment (pkowalsk1) [reply panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp]: Nonetheless, I believe it's unnecessary. You can refer to this example for context: https://github.com/husarion/panther_ros/blob/c54e3f95549a5be55ac1ff1f9dd587ac6746d1a2/panther_hardware_interfaces/src/canopen_controller.cpp#L46

There are several threads in this project that are managed with smart pointers; however, in my opinion, we should consider changing this approach.

File: panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp (hunk 14 above)
Comment (pkowalsk1) [reply panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp]: In my opinion, this is a similar case to the one mentioned above, but I won't insist on it. :p

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 22 above)
Comment (pkowalsk1) [reply panther_hardware_interfaces/src/panther_imu.cpp]: Exactly, sorry xd

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 27 above)
Comment (pkowalsk1) [reply panther_hardware_interfaces/src/panther_imu.cpp]: As we discussed earlier, we can remove this service and the entire imu_ros_interface, right? :p

File: panther_hardware_interfaces/src/phidgets_spatial_parameters.yaml (hunk 31 above)
Comment (pkowalsk1) [reply panther_hardware_interfaces/src/phidgets_spatial_parameters.yaml]: Okay, so I have no strong opinion on this matter.

File: panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp (hunk 13 above)
Comment (pkowalsk1) [reply panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp]: https://stackoverflow.com/questions/44832054/storing-an-stdthread-in-c11-smart-pointer :p

File: panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp (hunk 5 above)
Comment (delihus) [reply panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp]: I deleted the mutex so it is no need to include this,

File: panther_hardware_interfaces/src/panther_imu.cpp (hunk 22 above)
Comment (delihus) [reply panther_hardware_interfaces/src/panther_imu.cpp]: It can be. I check if the size is correct in 'CheckStatesSize' and then also in  `CheckInterfaces` function.

File: panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp (hunk 13 above)
Comment (delihus) [reply panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp]: Okay, thank you for that! Next time I will apply the thread without smart pointer but for now we removed the thread.

File: panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp (hunk 14 above)
Comment (delihus) [reply panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu_ros_interface.hpp]: Okay, but also outdated ;)

File: panther_hardware_interfaces/src/panther_imu.cpp [hunk 35]
```diff
@@ -0,0 +1,507 @@
...
+{
+  if (
+    not std::isfinite(mag_compensated.x) or not std::isfinite(mag_compensated.y) or
+    not std::isfinite(mag_compensated.z)) {
+    // The nan values are provided when imu is not callibrated.
+    if (not imu_calibrated_) {
+      return;
+    } else {
+      throw std::runtime_error("Magnetometer has nan values.");
+    }
```
pkowalsk1: Maybe I'm a boomer but can we use `!`, `&&` and `||` instead of `and`, `or` and `not`? :p

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,507 @@
...
+  RCLCPP_INFO_STREAM(
+    logger_, "Connecting to Phidgets Spatial serial " << params_.serial << ", hub port "
+                                                      << params_.hub_port << "...");
+
+  try {
+    spatial_ = std::make_unique<phidgets::Spatial>(
+      params_.serial, params_.hub_port, false,
+      std::bind(
+        &PantherImuSensor::SpatialDataCallback, this, std::placeholders::_1, std::placeholders::_2,
+        std::placeholders::_3, std::placeholders::_4),
```
pkowalsk1: Could we add `using namespace std::placeholders;` in the header?

panther_hardware_interfaces/src/panther_imu.cpp
pkowalsk1:
```suggestion
  if (not imu_calibrated_) {
    return;
  }
  
  if (
    not std::isfinite(mag_compensated.x) or not std::isfinite(mag_compensated.y) or
    not std::isfinite(mag_compensated.z)) {
    throw std::runtime_error("Magnetometer has nan values.");
  }
```
Is it the same logic?

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,507 @@
...
+  }
+
+  geometry_msgs::msg::Quaternion init_q;
+  if (!StatelessOrientation::computeOrientation(world_frame_, lin_acc, mag_compensated, init_q)) {
+    throw std::runtime_error(
+      "The IMU seems to be in free fall, cannot determine gravity direction.");
+  }
+  filter_.setOrientation(init_q.w, init_q.x, init_q.y, init_q.z);
+
+  first_data_callback_ = false;
```
pkowalsk1: Isn't it sufficient to use just the imu_calibrated_ flag and remove the first_data_callback_ flag, since they both change state at the same points in the code?

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,149 @@
...
+  static constexpr double KImuMagneticFieldUnknownValue = 1e300;
+  static constexpr float G = 9.80665;
+  inline static const std::string kImuSensorName = "imu";
+  inline static const std::array<std::string, kImuInterfacesSize> kImuInterfacesNames = {
+    "orientation.x",         "orientation.y",         "orientation.z",      "orientation.w",
+    "angular_velocity.x",    "angular_velocity.y",    "angular_velocity.z", "linear_acceleration.x",
+    "linear_acceleration.y", "linear_acceleration.z",
+  };
+
+  std::map<std::string, WorldFrame::WorldFrame> kImuWorldFramesMap = {
```
pkowalsk1: I believe this is unused.

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,149 @@
...
+    const geometry_msgs::msg::Vector3 & mag_compensated, const double dt);
+  void UpdateStatesValues(
+    const geometry_msgs::msg::Vector3 & ang_vel, const geometry_msgs::msg::Vector3 & lin_acc);
+  void RemoveGravityVectorFromAccelerationState();
+
+  void Calibrate();
+
+  std::vector<double> imu_sensor_state_;
+  rclcpp::Logger logger_{rclcpp::get_logger("PantherImuSensor")};
+  rclcpp::Clock steady_clock_{RCL_STEADY_TIME};
```
pkowalsk1: Is this unused?

panther_hardware_interfaces/test/include/test_imu_constants.hpp
```diff
@@ -0,0 +1,65 @@
...
+  {"mag_bias_y", "0.0"},
+  {"mag_bias_z", "0.0"},
+  {"use_mag", "true"},
+  {"stateless", "false"},
+  {"remove_gravity_vector", "true"},
+  {"world_frame", "enu"}};
+
+const std::string kPluginName = "<plugin>panther_hardware_interfaces/PantherImuSensor</plugin>";
+
+const std::string kCalibrateService = "/panther_imu_node/calibrate";
```
pkowalsk1: It is removed

panther_hardware_interfaces/test/test_panther_imu.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <hardware_interface/resource_manager.hpp>
+#include <hardware_interface/types/lifecycle_state_names.hpp>
+
+#include <lifecycle_msgs/msg/state.hpp>
+
+#include <panther_utils/test/test_utils.hpp>
+
+#include <panther_imu_test_utils.hpp>
+
+TEST(TestPantherImu, check_interfaces)
```
pkowalsk1: This might be a significant change for these tests, but I suggest that these utilities and constants be organized within a class that inherits from testing::Test, similar to the approach in other packages like panther_battery or panther_lights.

Additionally, there are many functionalities that remain untested, such as exception throwing, calibration, various process failures, and more. It would be beneficial to expand the tests to cover these scenarios.

pkowalsk1 (changes requested):

panther_hardware_interfaces/src/panther_imu.cpp
delihus: No, because when values are `nan` at the beggining it means when the imu is calibrating. If the `if(imu_calibration_)` instruction is before checking nans the imu will not calibrate (the values of `imu_calibrated_` will not change).

The imu is calibrated when there is no nans and the `imu_calibrated_` is `false`.

panther_hardware_interfaces/src/panther_imu.cpp
delihus: After last changes this is unused.

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,498 @@
...
+{
+  imu_sensor_state_.resize(
+    info_.sensors[0].state_interfaces.size(), std::numeric_limits<double>::quiet_NaN());
+}
+
+void PantherImuSensor::Calibrate()
+{
+  spatial_->zero();
+  // The API call returns directly, so we "enforce" the recommended 2 sec
+  // here. See: https://github.com/ros-drivers/phidgets_drivers/issues/40
```
pkowalsk1:
```suggestion
```

panther_hardware_interfaces/test/include/panther_imu_test_utils.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include <test_imu_constants.hpp>
+
+namespace panther_hardware_interfaces_test
+{
+
+/**
+ * @brief Utility class for testing Panther Imu
+ */
+class PantherImuTestUtils
```
pkowalsk1: I believe this class is unnecessary. In my opinion, these methods should be defined in the TestPantherImuSensor class as they are specific to its functionality. You can override SetUp and TearDown from ::testing::Test instead of defining Start and Stop methods. The same applies to constants; they could even be, for example, a static struct in the .cpp file.

panther_hardware_interfaces/test/test_panther_imu.cpp
```diff
@@ -0,0 +1,327 @@
...
+  info.sensors.push_back({});
+
+  std::list<std::string> example_interfaces_names = {"state1", "state2", "state3", "state4",
+                                                     "state5", "state6", "state7", "state8",
+                                                     "state9", "state10"};
+  for (const auto & interface_name : example_interfaces_names) {
+    hardware_interface::InterfaceInfo interface_info;
+    interface_info.name = interface_name;
+    info.sensors.front().state_interfaces.push_back(interface_info);
+  }
```
pkowalsk1: The same lines of code are present in the CheckStatesSize method. Could we extract the for loop and info initialization into a separate function?

panther_hardware_interfaces/test/test_panther_imu.cpp
```diff
@@ -0,0 +1,327 @@
...
+TEST_F(TestPantherImuSensor, HandleFirstDataCallback)
+{
+  geometry_msgs::msg::Vector3 mag_compensated;
+  geometry_msgs::msg::Vector3 lin_acc;
+
+  mag_compensated.x = std::numeric_limits<float>::quiet_NaN();
+  mag_compensated.y = std::numeric_limits<float>::quiet_NaN();
+  mag_compensated.z = std::numeric_limits<float>::quiet_NaN();
+
+  // When imu is not calibrated  nans are skipped
```
pkowalsk1:
```suggestion
  // When imu is not calibrated NaNs are skipped
```

panther_hardware_interfaces/test/test_panther_imu.cpp
```diff
@@ -0,0 +1,327 @@
...
+  EXPECT_NO_THROW({ imu_sensor_->HandleFirstDataCallback(mag_compensated, lin_acc, 0.0); });
+
+  mag_compensated.x = 1.0;
+  mag_compensated.y = 1.0;
+  mag_compensated.z = 1.0;
+
+  // IMU can't find gravity vector due to empyy acceleration vector
```
pkowalsk1:
```suggestion
  // IMU can't find gravity vector due to empty acceleration vector
```

panther_hardware_interfaces/test/test_panther_imu.cpp
```diff
@@ -0,0 +1,327 @@
...
+  EXPECT_THROW(
+    { imu_sensor_->HandleFirstDataCallback(mag_compensated, lin_acc, 0.0); }, std::runtime_error);
+
+  // IMU should calibrate
+  lin_acc.z = 9.80665;
+  EXPECT_NO_THROW({ imu_sensor_->HandleFirstDataCallback(mag_compensated, lin_acc, 0.0); });
+}
+
+TEST(TestPantherImu, CheckInterfacesLoadedByResourceManager)
```
pkowalsk1: This comment is in reference to the earlier suggestion regarding the relocation of PantherImuTestUtils to TestPantherImu. My preference is to transform these cases into a test suite and make use of the methods available in TestPantherImuSensor to achieve a more cohesive approach.

panther_hardware_interfaces/test/test_panther_imu.cpp
pkowalsk1: I am concerned that the first part of this comment has not been addressed.

pkowalsk1 (changes requested):
In my opinion, the issues with these tests have still not been fully addressed. However, I must commend the organization of `panther_imu.cpp` and others; it is very well structured, clean, and easy to understand. I truly appreciate this piece of code. ;)

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
delihus: Update. It is necessary to use protected because `PantherImuSensorWrapper` uses these functions.

panther_hardware_interfaces/test/include/panther_imu_test_utils.hpp
delihus: The functionalities are moved to `TestPantherImuSensor`

panther_hardware_interfaces/test/test_panther_imu.cpp
delihus: Created:
- `hardware_interface::HardwareInfo CreateExampleInterfaces(const hardware_interface::HardwareInfo &info)`
- `hardware_interface::HardwareInfo CreateCorrectInterfaces(const hardware_interface::HardwareInfo &info)`

pkowalsk1 (approved):

panther_description/urdf/panther_macro.urdf.xacro
```diff
@@ -145,6 +146,56 @@
...
+              Gyroscope Drift Max is
+              gyroMeasDrift  = 0.1 deg/s = 0.00175 rad/s
+              read from specification https://www.phidgets.com/?prodid=1205#Tab_Specifications
+
+              Depend on the paper of Madgwick Filter https://x-io.co.uk/downloads/madgwick_internal_report.pdf
+              gain is represented as
+              gain = sqrt(3/4)* gyroMeasError = 0.00303
+              zeta is represented as
+              zeta = sqrt(3/4)* gyroMeasDrift = 0.00151
+              -->
```
pkowalsk1: I would rather move this information to a README file or something similar.

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,177 @@
...
+  CallbackReturn on_shutdown(const rclcpp_lifecycle::State & previous_state) override;
+  CallbackReturn on_error(const rclcpp_lifecycle::State & previous_state) override;
+
+  std::vector<StateInterface> export_state_interfaces() override;
+
+  return_type read(const rclcpp::Time & /* time */, const rclcpp::Duration & /* period */) override;
+
+  static constexpr size_t kImuInterfacesSize = 10;
+  static constexpr double KImuMagneticFieldUnknownValue = 1e300;
+  static constexpr float G = 9.80665;
```
pkowalsk1, line 76: Is there any reason to make it public?

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+protected:
+  /**
+   * @brief Checks if the sensor name defined in the urdf matches the expected name.
+   * @throw std::runtime_error If the sensor name does not match the expected name.
+   */
+  void CheckSensor() const;
```
pkowalsk1: So, maybe CheckSensorName is a better name?

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,542 @@
...
+  imu_sensor_state_.resize(
+    info_.sensors[0].state_interfaces.size(), std::numeric_limits<double>::quiet_NaN());
+}
+
+void PantherImuSensor::Calibrate()
+{
+  spatial_->zero();
+
+  RCLCPP_WARN(logger_, "IMU is calibrating. Do not move the robot for 2 seconds!");
+  while (!imu_calibrated_) {
```
pkowalsk1: Should we use IsIMUCalibrated here? Also, I would like to inquire again if it's feasible to wait until the IMU is calibrated using synchronization mechanisms such as std::condition_variable or std::promise. I believe it is possible :p

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,542 @@
...
+  const double acceleration[3], const double angular_rate[3], const double magnetic_field[3],
+  const double timestamp)
+{
+  const auto timestamp_s = timestamp * 1e-3;
+
+  const auto mag_compensated = ParseMagnitude(magnetic_field);
+  const auto ang_vel = ParseGyration(angular_rate);
+  const auto lin_acc = ParseAcceleration(acceleration);
+
+  // Skip the data callback when IMU if not calibrated
```
pkowalsk1:
```suggestion
  // Skip the data callback when IMU is not calibrated
```

panther_hardware_interfaces/src/panther_imu.cpp
```diff
@@ -0,0 +1,542 @@
...
+  }
+}
+
+void PantherImuSensor::UpdateMadgwickAlgorithmIMU(
+  const geometry_msgs::msg::Vector3 & ang_vel, const geometry_msgs::msg::Vector3 & lin_acc,
+  const double dt)
+{
+  filter_->madgwickAHRSupdateIMU(
+    ang_vel.x, ang_vel.y, ang_vel.z, lin_acc.x, lin_acc.y, lin_acc.z, dt);
+}
```
pkowalsk1: I don't understand the division into two methods. Are we certain that their use is appropriate?

pkowalsk1 (changes requested):

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_imu.hpp
delihus: They are used in tests

panther_hardware_interfaces/src/panther_imu.cpp
delihus: There should not be `if (params_.use_mag) {` .

pkowalsk1 (approved):

PR #238, opened 2024-03-05
ROS2 - Fix Power Motor Service

pkowalsk1:


panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system.hpp
```diff
@@ -146,6 +147,8 @@ class PantherSystem : public hardware_interface::SystemInterface
 
   rclcpp::Time next_driver_state_update_time_{0, 0, RCL_ROS_TIME};
   rclcpp::Duration driver_states_update_period_{0, 0};
+
+  std::mutex e_stop_manipulation_mtx_;
```
pawelirh: Would be cleaner to place it together with a controlled attribute (I believe `e_stop_` here)

pawelirh (approved):

PR #246, opened 2024-03-14
Ros2 panther manager

KmakD:
Most of the work is done, can't be merged as it is waiting for [plugin PR](https://github.com/husarion/panther_ros/pull/232) first. Plugin PR should include merging newest ros2-devel branch.
Things to finish after plugins are added: 
- remap fan_enable, aux_enable etc. (should be /hardware/name) services controller launch
- update bringup launch 
- remove dummy_scheduler
- add shutdown_hosts.yaml 
- change ExpectThrowWithDesription to EXPECT_TRUE(IsMessageThrown) in tests


panther_manager/config/Panther106BT.btproj
pawelirh, line 1: The `config` directory seems to be a bit messy. What do you think of the following idea?
- The `config` stores the node's configuration file
- A new directory (on the same level as `config` not in the config) is made to store BT-related files, let's say `behaviour_trees`

panther_manager/include/panther_manager/manager_bt_node.hpp
```diff
@@ -0,0 +1,119 @@
...
+#include <panther_msgs/msg/io_state.hpp>
+#include <panther_msgs/msg/led_animation.hpp>
+#include <panther_msgs/msg/system_status.hpp>
+
+#include <panther_utils/moving_average.hpp>
+
+namespace panther_manager
+{
+
+using BatteryStateMsg = sensor_msgs::msg::BatteryState;
```
pawelirh, line 40: Please sort definitions alphabetically

panther_manager/src/manager_bt_node.cpp
```diff
@@ -0,0 +1,441 @@
...
+  const float battery_state_anim_period =
+    this->get_parameter("lights.battery_state_anim_period").as_double();
+  const float low_battery_anim_period =
+    this->get_parameter("lights.low_battery_anim_period").as_double();
+  const float low_battery_threshold_percent =
+    this->get_parameter("lights.low_battery_threshold_percent").as_double();
+
+  const std::map<std::string, std::any> lights_initial_bb = {
+    {"charging_anim_percent", ""},
+    {"current_anim_id", -1},
```
pawelirh, line 210: I believe all of the elements in this map should be uppercased.

panther_manager/src/manager_bt_node.cpp
```diff
@@ -0,0 +1,441 @@
...
+      config.blackboard->set<std::string>(name, std::any_cast<std::string>(value));
+    } else {
+      throw std::invalid_argument(
+        "Invalid type for blackboard entry. Valid types are: bool, int, unsigned, float, double, "
+        "const char*, std::string");
+    }
+  }
+
+  return config;
+}
```
pawelirh: Can we move this function to the external header file - sth like `bt_utils`?

panther_manager/src/manager_bt_node.cpp
```diff
@@ -0,0 +1,441 @@
...
+  // -------------------------------
+
+  if (launch_lights_tree) {
+    lights_tree_timer_ = this->create_wall_timer(
+      std::chrono::milliseconds(100), std::bind(&ManagerBTNode::LightsTreeTimerCB, this));
+  }
+  if (launch_safety_tree) {
+    safety_tree_timer_ = this->create_wall_timer(
+      std::chrono::milliseconds(100), std::bind(&ManagerBTNode::SafetyTreeTimerCB, this));
+  }
```
pawelirh: I would see the time interval as a parameter

panther_manager/test/test_manager_bt_node.cpp
```diff
@@ -0,0 +1,273 @@
...
+};
+
+class TestManagerBTNode : public testing::Test
+{
+public:
+  TestManagerBTNode();
+  ~TestManagerBTNode();
+
+protected:
+  void CreateBTPorjectFile(const std::string & tree_xml);
```
pawelirh: Typo

panther_manager/test/test_manager_bt_node.cpp
```diff
@@ -0,0 +1,273 @@
...
+};
+
+void TestManagerBTNode::CreateBTPorjectFile(const std::string & tree_xml)
+{
+  std::ofstream out(bt_project_path_);
+  if (out.is_open()) {
+    out << tree_xml;
+    out.close();
+  }
+}
```
pawelirh: Please move it after constructor

panther_manager/test/test_manager_bt_node.cpp
```diff
@@ -0,0 +1,273 @@
...
+  EXPECT_NO_THROW(manager_bt_node_->RegisterBehaviorTree());
+}
+
+TEST_F(TestManagerBTNode, CreateBTConfigInvalidItem)
+{
+  const std::map<std::string, std::any> bb_values = {{"value", 1l}};
+
+  panther_utils::test_utils::ExpectThrowWithDescription<std::invalid_argument>(
+    [&]() { manager_bt_node_->CreateBTConfig(bb_values); }, "Invalid type for blackboard entry.");
+}
```
pawelirh: This method is not compatible with the current state of `ros2-devel`, is sth that blocks us before pulling `ros2-devel` branch so further development can be made without any conflicts?

pawelirh (changes requested):
I have some doubts regarding the architecture (coupling between different behavior trees). I'm not sure if at this exact moment, we want to pay attention to such a discussion. However, let me mention a few aspects.

- The states (received messages) are updated on the blackboard frequently.
- It's hard to scale. A new BT, or modification of an existing one demands changes in the core node and pulls after unit test modifications.
- Single responsibility for the node is not kept, since the node in this example is responsible for creating and configuring the BT and blackboard.
- Why do we push to have several independent trees, since they are in the end using the same resources and executing interchangeably (SingleThreadedExecutor)?

panther_manager/src/manager_bt_node.cpp
KmakD: Only const values of the blackboard are upper case, this two can change in runtime

panther_manager/test/test_manager_bt_node.cpp
KmakD: There was a comment explaining that this will be changed in the future, the future has come so I changed it :P

KmakD:
>     * Why do we push to have several independent trees, since they are in the end using the same resources and executing interchangeably (SingleThreadedExecutor)?

My mistake, this should be MultiThreadedExecutor. It's fixed. The idea behind two separate BTs is that safety is a crucial part of system  and should be handled independently, we don't want deadlock or some sleep behavior in lights affect safety features.

panther_manager/main.cpp
pawelirh, line 1: What is this file for?

panther_manager/src/main.cpp
```diff
@@ -25,8 +25,11 @@ int main(int argc, char ** argv)
   auto manager_bt_node = std::make_shared<panther_manager::ManagerBTNode>("manager_bt_node");
   manager_bt_node->Initialize();
 
+  rclcpp::executors::MultiThreadedExecutor executor;
```
pawelirh: MultiThreadedExecutor without specified proper callback groups is in practice SingleThreadedExecutor. I assume that changes based on our latest discussion have not yet been implemented.

panther_manager/src/manager_bt_node.cpp
pawelirh, line 1: WIll the README file be added at a later stage of the development?

panther_manager/main.cpp
KmakD: some merging oopsie

panther_manager/src/manager_bt_node.cpp
KmakD: No, forgot to update that. It is mostly copied from ROS 1

pawelirh: This readme seems to be a bit overloaded, how about:
- leaving ros node-related things in the main readme file
- creating additional markdown doc for BT things

panther_manager/README.md
```diff
@@ -1,7 +1,301 @@
...
+- `safety.timer_frequency` [*float*, default: **10.0**]: frequency **[Hz]** at which safety tree will be ticked.
+- `shutdown_hosts_path` [*string*, default: **None**]: path to a YAML file containing a list of hosts to request shutdown. To correctly format the YAML file, include a **hosts** field consisting of a list with the following fields:
+  - `command` [*string*, default: **sudo shutdown now**]: command executed on shutdown of given device.
+  - `ip` [*string*, default: **None**]: IP of a host to shutdown over SSH.
+  - `ping_for_success` [*bool*, default: **true**]: ping host until it is not available or timeout is reached.
+  - `port` [*string*, default: **22**]: SSH communication port.
+  - `timeout` [*string*, default: **5.0**]: time in **[s]** to wait for the host to shutdown. The Built-in Computer will turn off after all computers are shutdown or reached timeout. Keep in mind that hardware will cut power off after a given time after pressing the power button. Refer to the hardware manual for more information.
+  - `username` [*string*, default: **None**]: username used to log in to over SSH.
+
+[//]: # (ROS_API_NODE_PARAMETERS_END)
```
pawelirh, line 86: Timer frequency parameters are missing in the configuration file. Can we add them?

pawelirh (approved):
LGTM 👌🏼

PR #247, opened 2024-03-15
ROS 2 - Add GPIOController Tests

pkowalsk1:


panther_hardware_interfaces/test/test_gpio_controller.cpp
```diff
@@ -0,0 +1,161 @@
...
+  void WatchdogEnable() { this->watchdog_->TurnOn(); }
+  void WatchdogDisable() { this->watchdog_->TurnOff(); }
+  bool IsWatchdogEnabled() { return this->watchdog_->IsWatchdogEnabled(); }
+};
+
+class TestGPIOController : public ::testing::Test
+{
+public:
+  void SetUp() override;
+  void TearDown() override { gpio_controller_wrapper_.reset(); }
```
pawelirh: Is there any specific reason for using the `SetUp` and `TearDown` methods? The general approach suggests using constructor/destructor except. Despite the GPIO tests, we use the const/dest approach in the Panther project.

https://fuchsia.googlesource.com/third_party/googletest/+/HEAD/googletest/docs/faq.md#ctorvssetup

panther_hardware_interfaces/test/test_gpio_controller.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+protected:
+  float GetRobotVersion();
+
+  std::unique_ptr<GPIOControllerWrapper> gpio_controller_wrapper_;
+  const std::vector<panther_gpiod::GPIOInfo> gpio_config_info_storage_{
```
pawelirh: I'd make this vector a global variable to prevent recreating it on each test case.

panther_hardware_interfaces/test/test_gpio_controller.cpp
```diff
@@ -0,0 +1,161 @@
...
+  auto edge_cnt = std::make_shared<int>(0);
+
+  gpio_controller_wrapper_->RegisterGPIOEventCallback([this, edge_cnt](const auto & state) mutable {
+    if (state.pin == panther_gpiod::GPIOPin::WATCHDOG) {
+      (*edge_cnt)++;
+    }
+  });
+
+  gpio_controller_wrapper_->WatchdogEnable();
+  EXPECT_TRUE(gpio_controller_wrapper_->IsWatchdogEnabled());
```
pawelirh: Assertion should be a better option here - no sense in further testing if the watchdog is not successfully enabled.

panther_hardware_interfaces/test/test_gpio_controller.cpp
```diff
@@ -0,0 +1,161 @@
...
+  gpio_controller_wrapper_->ChargerEnable(false);
+  EXPECT_TRUE(gpio_controller_wrapper_->IsPinActive(panther_gpiod::GPIOPin::CHRG_DISABLE));
+}
+
+TEST_F(TestGPIOController, TestQueryControlInterfaceIOStates)
+{
+  std::unordered_map<panther_gpiod::GPIOPin, bool> io_states =
+    gpio_controller_wrapper_->QueryControlInterfaceIOStates();
+
+  EXPECT_EQ(io_states.size(), 7);
```
pawelirh: If having failure here makes no sens in further testing, we could use assertion as well.

pawelirh (changes requested):

pawelirh (approved):
LGTM 👌🏼

PR #249, opened 2024-03-16
ROS 2 - Fix estop threads

pkowalsk1:
## Changes

- During EStop triggering, CAN communication is only used if GPIO pin manipulation fails,
- The single-thread executor has been replaced with a multi-threaded one, enabling simultaneous handling of the `e_stop_trigger` and `e_stop_reset` services. This has helped to implement behavior where the trigger has a higher "priority" than the reset, ensuring that each trigger is immediate (unlike before, where it had to wait max. 200ms for the reset to complete). These changes should guarantee thread safety. The rest of the services are still managed in the node's default callback group, so they are not executed concurrently.

### DIFF
--- a/panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
+++ b/panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
@@ -34,10 +34,10 @@
 namespace panther_hardware_interfaces
 {
 
-class e_stop_reset_interrupted : public std::exception
+class EStopResetInterrupted : public std::exception
 {
 public:
-  e_stop_reset_interrupted(const std::string & message) : msg_(message) {}
+  EStopResetInterrupted(const std::string & message) : msg_(message) {}
   const char * what() const noexcept override { return msg_.c_str(); }
 
 private:
@@ -223,6 +223,17 @@ protected:
   std::unique_ptr<Watchdog> watchdog_;
 
 private:
+  /**
+   * @brief Waits for a specific duration or until an interruption is signaled.
+   *
+   * This method is designed to block execution for the specified timeout duration. It also monitors
+   * for an interruption signal which, if received, will cause the method to return early. The
+   * interruption is controlled by the `should_abort_e_stop_reset_` flag.
+   *
+   * @param timeout Duration to wait for in milliseconds.
+   * @return `true` if the wait completed without interruption, `false` if an interruption was
+   * signaled.
+   */
   bool WaitFor(std::chrono::milliseconds timeout);
 
   /**
--- a/panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
+++ b/panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
@@ -37,7 +37,7 @@ namespace panther_hardware_interfaces
 class e_stop_reset_interrupted : public std::exception
 {
 public:
-  e_stop_reset_interrupted(const std::string & message = "") : msg_(message) {}
+  e_stop_reset_interrupted(const std::string & message) : msg_(message) {}
   const char * what() const noexcept override { return msg_.c_str(); }
 
 private:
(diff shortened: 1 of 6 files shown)


panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
```diff
@@ -34,6 +34,16 @@
 namespace panther_hardware_interfaces
 {
 
+class e_stop_reset_interrupted : public std::exception
```
KmakD: Why not CamelCase?

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -257,4 +257,30 @@ bool PantherSystemRosInterface::UpdateIOStateMsg(
...
+  if (group_id == 0) {
+    if (callback_group_type == rclcpp::CallbackGroupType::Reentrant) {
+      throw std::runtime_error(
+        "Node callback group with id 0 (default group) cannot be of "
+        "rclcpp::CallbackGroupType::Reentrant type.");
+    }
+    return nullptr;  // default node callback group
+  }
+
+  auto search = callback_groups_.find(group_id);
```
KmakD:
```suggestion
  const auto search = callback_groups_.find(group_id);
```

panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -153,9 +153,11 @@ CallbackReturn PantherSystem::on_activate(const rclcpp_lifecycle::State &)
     std::bind(&PantherSystem::MotorsPowerEnable, this, std::placeholders::_1));
 
   panther_system_ros_interface_->AddService<TriggerSrv, std::function<void()>>(
-    "~/e_stop_trigger", std::bind(&PantherSystem::SetEStop, this));
+    "~/e_stop_trigger", std::bind(&PantherSystem::SetEStop, this), 1,
+    rclcpp::CallbackGroupType::MutuallyExclusive);
   panther_system_ros_interface_->AddService<TriggerSrv, std::function<void()>>(
-    "~/e_stop_reset", std::bind(&PantherSystem::ResetEStop, this));
+    "~/e_stop_reset", std::bind(&PantherSystem::ResetEStop, this), 2,
+    rclcpp::CallbackGroupType::MutuallyExclusive);
```
KmakD, line 162: This may throw. Is it expected for system to die if this fails?

panther_hardware_interfaces/src/panther_system_ros_interface.cpp
```diff
@@ -257,4 +257,30 @@ bool PantherSystemRosInterface::UpdateIOStateMsg(
...
+  if (search != callback_groups_.end()) {
+    if (search->second->type() == callback_group_type) {
+      return search->second;
+    } else {
+      throw std::runtime_error("Requested node callback group has incorrect type.");
+    }
```
KmakD:
```suggestion
    if (search->second->type() != callback_group_type) {
      throw std::runtime_error("Requested node callback group has incorrect type.");
    }
    return search->second;
```

panther_hardware_interfaces/src/gpio_controller.cpp
```diff
@@ -189,6 +196,24 @@ GPIOControllerPTH12X::QueryControlInterfaceIOStates() const
...
 }
 
+void GPIOControllerPTH12X::InterruptEStopReset()
+{
+  std::lock_guard<std::mutex> lck(e_stop_cv_mtx_);
+  should_abort_e_stop_reset_ = true;
+  e_stop_cv_.notify_one();
+}
+
+bool GPIOControllerPTH12X::WaitFor(std::chrono::milliseconds timeout)
```
KmakD, line 206: this could use a brief explaining that this method is not only waiting but also checks for interruption

KmakD (changes requested):

panther_hardware_interfaces/src/panther_system.cpp
pkowalsk1: Yes, it is. Theoretically, it could cause 'safety' issues.

panther_hardware_interfaces/include/panther_hardware_interfaces/gpio_controller.hpp
JanBrzyk: CamelCase approved

KmakD (approved):

PR #250, opened 2024-03-16
ROS 2 - Fix Motor States QoS

pkowalsk1:


panther_battery/src/battery_node.cpp
```diff
@@ -131,7 +131,7 @@ void BatteryNode::InitializeWithRoboteqBattery()
   };
 
   driver_state_sub_ = this->create_subscription<DriverStateMsg>(
-    "driver/motor_controllers_state", rclcpp::SensorDataQoS(),
+    "driver/motor_controllers_state", 5,
```
pawelirh, line 134: Why do we go back to the default system QoS settings?

pawelirh (approved):
LGTM

PR #258, opened 2024-03-28
ROS 2- Fix Error Clearing Mechanism for Roboteq Controllers

pkowalsk1:


panther_hardware_interfaces/src/panther_system.cpp
```diff
@@ -563,6 +563,12 @@ void PantherSystem::UpdateFlagErrors()
         << "\tFront: " << motors_controller_->GetFrontData().GetFlagErrorLog()
         << "\tRear: " << motors_controller_->GetRearData().GetFlagErrorLog());
     roboteq_error_filter_->UpdateError(ErrorsFilterIds::ROBOTEQ_DRIVER, true);
+
+    // Attempt to clear Roboteq motor controller error flags by sending a zero velocity command via
+    // CAN
+    if (last_commands_zero_) {
+      SendCommands();
+    }
```
pawelirh: I wonder if it wouldn't be more readable creating a function with descriptive name wrapping the condition and triggering command sending so we can get rid of the comment. What do you think?

pkowalsk1: I have implemented a wrapper for PDO operations in `PantherSystem` and used it for the `AttemptErrorFlagResetWithZeroSpeed` and `SendSpeedCommands` function call. What are your thoughts?

pawelirh (approved):
LGTM 👌🏼

pawelirh (approved):

PR #266, opened 2024-04-11
Add unit test workflow

pawelirh:
## Changes

- Implemented testing workflow
- Fully tested on the `panteher_utils` package ✅ 
- Runner needs to be further improved to handle successfully all unit tests

❗ SQUASH COMMITS! ❗

### DIFF
--- a/.github/workflows/run-unit-tests.yaml
+++ b/.github/workflows/run-unit-tests.yaml
@@ -42,7 +42,7 @@ jobs:
               working-directory: ros2_ws
               run: |
                   if [ -f install/setup.bash ]; then source install/setup.bash; fi
-                  colcon build --symlink-install --parallel-workers $(nproc) --packages-up-to ${{ env.ROS_PKG_NAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/run-unit-tests.yaml
delihus, line 1: It will be nice to provide the information that the tests are ran on hardware. Maybe modify the file name a little:
`run-unit-tests-on-hardware.yaml`

delihus (changes requested):
Please describe the workflow in the readme file. Maybe add some nice workflow status badge?
https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

pawelirh:
Regarding both comments

I'd wait with the readme update and badge until the tests are fully functional and pass to avoid always having a red badge.

As for the information that tests are run on hardware I'd see that info somewhere near/on the badge in readme. I'm not sure about the proposed filename since we don't need to distinguish between "hardware" and "non-hardware"(GitHub-hosted). What do you think, is it a good option in your opinion?

delihus (approved):

PR #268, opened 2024-04-15
Add E-Stop manager

pkowalsk1:


panther_hardware_interfaces/src/motors_controller.cpp
```diff
@@ -209,6 +214,13 @@ void MotorsController::TurnOnSafetyStop()
   }
 }
 
+void MotorsController::AttemptErrorFlagResetWithZeroSpeed()
+{
+  SendSpeedCommands(0.0, 0.0, 0.0, 0.0);
+}
+
+bool MotorsController::AreVelocityCommandsNearZero() { return last_command_zero_; }
```
pawelirh: Both functions might be inline defined in header file

panther_hardware_interfaces/include/panther_hardware_interfaces/motors_controller.hpp
```diff
@@ -117,13 +117,20 @@ class MotorsController
    */
   void AttemptErrorFlagResetWithZeroSpeed();
 
+  /**
+   * @brief Check if last velocity commands are near zero.
+   * @return Returns true if the last velocity commands are near zero, otherwise returns false.
+   */
+  bool AreVelocityCommandsNearZero();
```
pawelirh: I wonder if the name is not too much detailed. 
Maybe sth on a higher level of abstraction, e.g. `AreMotorsIdle()` is better option? What do you think?

If we decide to change a name here, function `AttemptErrorFlagResetWithZeroSpeed()` should be renamed as well.

panther_hardware_interfaces/src/motors_controller.cpp
```diff
@@ -135,6 +135,11 @@ void MotorsController::UpdateDriversState()
 void MotorsController::SendSpeedCommands(
   const float speed_fl, const float speed_fr, const float speed_rl, const float speed_rr)
 {
+  const float zero_threshold = std::numeric_limits<float>::epsilon();
+  last_command_zero_ = std::abs(speed_fl) <= zero_threshold &&
+                       std::abs(speed_fr) <= zero_threshold &&
+                       std::abs(speed_rl) <= zero_threshold && std::abs(speed_rr) <= zero_threshold;
+
```
pawelirh: How about extracrting this part to a separate method? In a current shape, the `SendSpeedCommands()` seems to breake Single Responsibility rule.

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system_e_stop_manager.hpp
pawelirh, line 1: 1. Why do we name this file `panther_system_e_stop_manager` instead of just `e_stop_manager`?
2. Please update `CODE_STRUCTURE.md` document.

panther_hardware_interfaces/src/panther_system_e_stop_manager.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  if (ConfirmEStopResetSuccessful()) {
+    e_stop_triggered_ = false;
+  }
+}
+
+void EStopManager::SetStrategy(std::unique_ptr<EStopStrategy> && strategy)
+{
+  strategy_ = std::move(strategy);
+  strategy_->SetManagerResources(resources_);
```
pawelirh: I wonder if such a dependency injection is neccessary here.

pros:
- flexibility in strategy change
- no need to pass dependencies separately to dependencies constructor

cons:
- additional struct created and increased the code complexity
- the context is coupled with the strategies details (what if there is a new strategy that requires different attributes)

I'm not saying this approach is wrong, it's more like a discussion point.
Did you find a similar implementation as an example?

panther_hardware_interfaces/src/panther_system_e_stop_manager.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+#include "panther_hardware_interfaces/panther_system_e_stop_manager.hpp"
+
+namespace panther_hardware_interfaces
+{
+
+void EStopStrategy::SetManagerResources(std::shared_ptr<EStopManagerResources> resources)
+{
+  manager_resources_ = resources;
+}
```
pawelirh: May be moved as inline to header

panther_hardware_interfaces/src/panther_system_e_stop_manager.cpp
```diff
@@ -0,0 +1,167 @@
...
+{
+  strategy_ = std::move(strategy);
+  strategy_->SetManagerResources(resources_);
+}
+
+void EStopManager::TriggerEStop() { strategy_->TriggerEStop(); };
+
+void EStopManager::ResetEStop() { strategy_->ResetEStop(); };
+
+bool EStopManager::ReadEStopState() { return strategy_->ReadEStopState(); };
```
pawelirh: I'd make them inline as well

panther_hardware_interfaces/CMakeLists.txt
pawelirh, line 1: Is implementation of unit tests for `e_stop_manager` in the scope of current task?

Isn't it neccessary to update `motors_controller` and `panther_system` tests or tests' dependencies?

pawelirh (changes requested):

pawelirh (approved):

PR #269, opened 2024-04-16
Manager launch fix

rafal-gorecki:
Fix `not None obj` and `not None obj` => always True
image


delihus:
Same issue here during simulation bringup.
This PR fixes this!
Screenshot from 2024-04-19 10-18-58

delihus (approved):

PR #271, opened 2024-04-22
Update README.md

YuriVanWarmerdam:
Proposed Minor change to install instructions, switching repo to ssh, since http cloning has been deprecated on most git clients.


rafal-gorecki (approved):

rafal-gorecki:
Thank @YuriVanWarmerdam for the suggestion. This solution will not work if someone does not have an SSH key created in the system. However, in the future it will probably be necessary, for now, to make cloning easier for users, I would leave it as it is.

PR #277, opened 2024-04-24
Ros2 add components

delihus:
After adding components
https://github.com/husarion/ros_components_description/pull/43

### DIFF
--- a/panther_bringup/config/components.yaml
+++ b/panther_bringup/config/components.yaml
@@ -1,4 +1,7 @@
-# The uncomment content in this file is the minimal requirement for components configuration file.
+# By default panther is loaded without any components.
+
+components: []
+
 # If you want to add for example LDR01, LDR06 and CAM01 to your Panther look at this example below:
 
 # components:
@@ -6,22 +9,17 @@
 #     parent_link: cover_link
 #     xyz: 0.0 0.1 0.0
 #     rpy: 0.0 0.0 0.0
-#     tf_prefix: main_lidar
-#     namespace: /main_lidar
-#
+#     device_namespace: main_lidar
+
 #   - type: LDR06
 #     parent_link: cover_link
 #     xyz: 0.0 -0.1 0.0
 #     rpy: 0.0 0.0 0.0
-#     tf_prefix: second_lidar
-#     namespace: /second_lidar
-#
+#     device_namespace: second_lidar
+
 #   - type: CAM01
 #     name: camera
 #     parent_link: cover_link
 #     xyz: 0.1 0.0 0.0
 #     rpy: 0.0 0.0 0.0
-#     tf_prefix: front_cam
-#     namespace: /front_cam
-
-components: []
+#     device_namespace: front_cam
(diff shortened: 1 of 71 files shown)


panther_bringup/config/components.yaml
```diff
@@ -0,0 +1,27 @@
+# The uncomment content in this file is the minimal requirement for components configuration file.
```
rafal-gorecki: Move it to panther_description ;)

panther_description/urdf/panther.urdf.xacro
```diff
@@ -36,4 +37,17 @@
     battery_config_file="$(arg battery_config_file)"
     namespace="$(arg namespace)" />
 
+  <xacro:arg name="components_config_path" default="$(find panther_bringup)/config/components.yaml" />
+  <xacro:property name="components_config_path_property" value="$(arg components_config_path)" />
+  <xacro:unless value="${components_config_path_property == 'None'}">
```
rafal-gorecki:
```suggestion
  <xacro:property name="components_config_path_property" value="$(arg components_config_path)" />

  <xacro:unless value="${components_config_path_property == 'None'}">
```

rafal-gorecki (changes requested):
Nice

rafal-gorecki:
There are also a few shortcomings that I suggest you fix immediately and add a few sprint points to your task.

1. First of all, I don't see the point of tf_prefix and namespace, because they are duplicated everywhere. In components, I would only leave device_namespace.

2. I would remove all topic arguments from the sensor because changing this parameter causes gz_bridge to not find the topic. Therefore, the change must be made in two places. Moreover, the need to remap a topic will be extremely rare and the user can always do it in launch.py

panther_bringup/config/components.yaml
delihus: This is good point. I was thinking about this. I put this the bringup because you choose your components at runtime let's say but maybe the default components `components: []` should be in description.

What do you prefer?

panther_bringup/config/components.yaml
```diff
@@ -0,0 +1,27 @@
...
+#
+#   - type: CAM01
+#     name: camera
+#     parent_link: cover_link
+#     xyz: 0.1 0.0 0.0
+#     rpy: 0.0 0.0 0.0
+#     tf_prefix: front_cam
+#     namespace: /front_cam
+
+components: []
```
rafal-gorecki: suggestion

```yaml
# By default panther is loaded without any components.

components: []

# If you want to add for example LDR01, LDR06 and CAM01 to your Panther look at this example below:

# components:
#   - type: LDR01
#     parent_link: cover_link
#     xyz: 0.0 0.1 0.0
#     rpy: 0.0 0.0 0.0
#     device_namespace: main_lidar

#   - type: LDR06
#     parent_link: cover_link
#     xyz: 0.0 -0.1 0.0
#     rpy: 0.0 0.0 0.0
#     device_namespace: second_lidar

#   - type: CAM01
#     name: camera
#     parent_link: cover_link
#     xyz: 0.1 0.0 0.0
#     rpy: 0.0 0.0 0.0
#     device_namespace: front_cam
```

rafal-gorecki (approved):

rafal-gorecki (changes requested):
Add short info to Readme about new arg or how to launch it

rafal-gorecki (approved):

PR #307, opened 2024-05-16
Launch refactor

rafal-gorecki:
### Description

Refactor launch arguments: leave only necessary and alphabetic order.

## Summary by CodeRabbit

- **New Features**
  - Introduced new launch configurations for Panther robot setup, including various parameters for simulation and localization.
  - Added functionality for launching URDF descriptions and spawning multiple robots in a simulation environment.
  
- **Bug Fixes**
  - Updated ROS node configurations by removing deprecated parameters and adjusting descriptions for existing ones.

- **Documentation**
  - Updated README files to reflect changes in parameters and configurations.
  - Enhanced package.xml files with additional dependencies and updated versioning.

- **Refactor**
  - Refactored launch files to improve readability and maintainability, including function renaming and logic refactoring.
  - Updated list syntax for wheel names in controller configurations for consistency.

- **Chores**
  - Adjusted pre-commit hook settings to allow larger file sizes.
  - Modified installation directory structure to include additional directories for better organization.


panther_controller/launch/controller.launch.py
```diff
@@ -34,31 +34,15 @@
...
+    components_config_path = LaunchConfiguration("components_config_path")
     controller_config_path = LaunchConfiguration("controller_config_path")
-    declare_controller_config_path_arg = DeclareLaunchArgument(
-        "controller_config_path",
-        description="Path to controller configuration file.",
-    )
+    namespace = LaunchConfiguration("namespace")
+    publish_robot_state = LaunchConfiguration("publish_robot_state")
+    wheel_config_path = LaunchConfiguration("wheel_config_path")
+    use_sim = LaunchConfiguration("use_sim")
```
pawelirh: APPLIES TO ALL

Is it worth instantiating the `LaunchConfiguration` object if used only once in the whole file? What do you think about specifying `LaunchConfiguration("<arg_name>")` in-place, and instantiating it only if used at least twice?

panther_controller/launch/controller.launch.py
```diff
@@ -34,31 +34,15 @@
...
-        description="Whether simulation is used",
-    )
-
-    wheel_config_path = LaunchConfiguration("wheel_config_path")
-    declare_wheel_config_path_arg = DeclareLaunchArgument(
-        "wheel_config_path",
-        description="Path to wheel configuration file.",
-    )
-
+    wheel_type = LaunchConfiguration("wheel_type")
```
pawelirh: APPLIES TO ALL

Sorting this file vertically by the object's context we should better place the `LaunchConfiguration` instantiation close the the corresponding argument declaration.

panther_description/launch/load_urdf.launch.py
pawelirh, line 1: 1. Where do we use this launch?
2. How about distinguishing a separate launch for just robot description so we don't need to define the same objects in `controller.launch.py`?

panther_gazebo/launch/spawn.launch.py
pawelirh, line 1: What is the motivation behind including here a bring-up launch? This makes the `spawn` launch not only spawning and is not intuitive to the user.

panther_lights/launch/lights.launch.py
```diff
@@ -17,29 +17,37 @@
...
         "led_config_file",
-        description="Path to a YAML file with a description of led configuration",
+        default_value=PathJoinSubstitution(
+            [FindPackageShare("panther_lights"), "config", "led_config.yaml"]
+        ),
+        description="Path to a YAML file with a description of led configuration.",
     )
-
-    namespace = LaunchConfiguration("namespace")
     declare_namespace_arg = DeclareLaunchArgument(
```
pawelirh:
```suggestion
    
    declare_namespace_arg = DeclareLaunchArgument(
```

panther_bringup/launch/bringup.launch.py
```diff
@@ -239,52 +128,39 @@ def generate_launch_description():
...
     )
 
     robot_localization_node = Node(
         package="robot_localization",
         executable="ekf_node",
         name="ekf_node",
-        output="screen",
         parameters=[ekf_config_path, {"tf_prefix": namespace}],
         namespace=namespace,
         remappings=[
```
pawelirh: How about creating a separate package `panther_robot_localization` for ekf things? We would store there config and launch file so from here ekf_launch will be included. Bring up as long as possible shouldn't define Nodes itself (I know there is still system_status node but it'll be changed in one of Dawid's upcoming PR).

pawelirh (changes requested):

pawelirh:
One thing more came to my mind. What do you think about listing launch files in a readme file? This would define what exact actions are performed by launch and would be even more convenient for packages with more than one launch file. 

**Response:** Not in this task. First fix docs parser.

panther_bringup/launch/bringup.launch.py
rafal-gorecki: IMO good idea. Also we get cleaner bringup, thanks to move declaration args.

pawelirh:
@coderabbitai review

panther_gazebo/launch/spawn.launch.py
```diff
@@ -32,37 +31,120 @@
 )
 from launch_ros.actions import Node, SetParameter
 from launch_ros.substitutions import FindPackageShare
-from nav2_common.launch import ParseMultiRobotPose
+from nav2_common.launch import ParseMultiRobotPose, ReplaceString
```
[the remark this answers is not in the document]
pawelirh: Hey rabbit, isn't the `ParseMultiRobotPose` used in line 115?

panther_gazebo/launch/spawn.launch.py
```diff
@@ -32,37 +31,120 @@
...
+    )
+
+    declare_robots_arg = DeclareLaunchArgument(
+        "robots",
+        default_value=[],
+        description=(
+            "The list of the robots spawned in the simulation e. g. robots:='robot1={x: 1.0, y:"
+            " -2.0}; robot2={x: 1.0, y: -4.0}'."
+        ),
+    )
```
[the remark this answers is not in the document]
pawelirh: Suggested change is not compliant with our formatting rules :(

panther_description/launch/load_urdf.launch.py
rafal-gorecki: This was the idea when creating this launch, but an error appears. Description on CU

panther_gazebo/launch/spawn.launch.py
rafal-gorecki: Nice point, I also wonder about it sometimes, but I'm used to it and don't notice it. Done

panther_bringup/launch/bringup.launch.py
pawelirh: I meant a truly separate package - sth as we got for ros2_control  -> `panther_controller`.

panther/panther_hardware.repos
```diff
@@ -11,6 +11,10 @@ repositories:
     type: git
     url: https://github.com/husarion/panther_msgs.git
     version: ros2
+  robot_localization: # Caused by this error https://github.com/cra-ros-pkg/robot_localization/pull/876, it can be removed after the next synchronization of apt packages.
+    type: git
+    url: https://github.com/cra-ros-pkg/robot_localization.git
+    version: humble-devel
```
pawelirh: We should refer to a specific commit hash

panther_bringup/launch/ekf.launch.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def generate_launch_description():
+    use_ekf = LaunchConfiguration("use_ekf")  # use_ekf must be before ekf_config_path
+    declare_use_ekf_arg = DeclareLaunchArgument(
+        "use_ekf",
+        default_value="True",
+        description="Enable or disable EKF.",
+        choices=["True", "False"],
+    )
```
pawelirh: If we decide to use this launch, I believe we aim to "use ekf". If so, what is the sense of this arg? If we want to somehow decide if ekf should be launched or not, the conditional should be at the stage of including ekf_launch

pawelirh (changes requested):

panther_gazebo/launch/spawn.launch.py
pawelirh: This is not exactly what I meant. The idea was to keep a single responsibility for a spawn launch file. As the name says, it spawns the robot. The user including such a launch file doesn't expect to perform additional actions like launching ekf, controller, etc. I understand that you want to somehow handle multiple robots, how about the following:
- spawn.launch.py - just spawns the robot
- simulate_robot.launch.py - includes spawn, ekf, controller and components launch descriptions
- simualte_multiple_robotes.launch.py - iterate over robots list including simulate_robot launch description 
- simualtion.launch.py - launches the whole simulation (master launch, basically as it is now)

What is your opinion here?

panther_description/launch/load_urdf.launch.py
pawelirh: The `joint_state_publisher` is still in the `load_urdf.launch.py`. Did you launch it with hardware and ros2_control stack?

panther/panther_hardware.repos
rafal-gorecki: This undo changes, because Jakub already remove robot_localization depend. This change is not considered in this PR.

panther_description/launch/load_urdf.launch.py
rafal-gorecki: That's why is `add_wheel_joints` arg added. One we fix passing `robot_description` topic to the `controller.launch.py` we can just include this launch with `add_wheel_joints` set as `False`. Without `add_wheel_joints` this launch runned separetly don't create all transformations.

panther_gazebo/launch/spawn.launch.py
rafal-gorecki: It is possible, but I am not shure why we need that, because to create static pather (e.g as a decoration) it is probably better to add mesh to the map. To create `spawn.launch.py` it is necesary to have `robot_description` so we will have to use `load_urdf.launch.py`. `controller.launch.py` already contain robot_state_publisher so we will need to use `publish_robot_state` arg to turn off it. Additionaly some features like battery are loaded to sim by default like `battery` topic, so maybe we should also add `gz_bridge` to `spawn.launch.py`.

We discuss this case on tomorow meeting.

panther_bringup/launch/ekf.launch.py
rafal-gorecki: I had in mind that each time this launch is included, the use_sim argument should also be automatically added, but the explicit declaration looks more readable.

panther_bringup/launch/bringup.launch.py
rafal-gorecki: Maybe shorter e.g. `panther_ekf` `panther_localization`?

pawelirh: `panther_localization` sounds the best 🏆

The name doesn't narrow the specific localization technique so we're able to add e.g. second launch and configuration for GPS setup in the future. 

What do u think? The package name would be changed but the config and launch file stay named `ekf`

panther_bringup/package.xml
```diff
@@ -15,14 +15,14 @@
...
   <exec_depend condition="$HUSARION_ROS_BUILD_TYPE == hardware">panther_battery</exec_depend>
-  <exec_depend>panther_controller</exec_depend>
+  <exec_depend condition="$HUSARION_ROS_BUILD_TYPE == hardware">panther_controller</exec_depend>
   <exec_depend condition="$HUSARION_ROS_BUILD_TYPE == hardware">panther_diagnostics</exec_depend>
-  <exec_depend>panther_lights</exec_depend>
+  <exec_depend condition="$HUSARION_ROS_BUILD_TYPE == hardware">panther_lights</exec_depend>
   <exec_depend condition="$HUSARION_ROS_BUILD_TYPE == hardware">panther_manager</exec_depend>
-  <exec_depend>python3-click</exec_depend>
+  <exec_depend condition="$HUSARION_ROS_BUILD_TYPE == hardware">panther_utils</exec_depend>
   <exec_depend>robot_localization</exec_depend>
```
pawelirh: `panther_bringup` is no longer dependent on `robot_localization` -> should be switched to `panther_ekf`/`panther_localization`

pawelirh (changes requested):

panther_bringup/launch/ekf.launch.py
pawelirh: We can easily set the use_sim_time while including this launch if that's the case. As you said (and I I got u right) this will be even more readable 👍🏼

panther_bringup/launch/bringup.launch.py
```diff
@@ -239,111 +71,72 @@ def generate_launch_description():
...
 
-    other_action_timer = TimerAction(
+    delayed_action = TimerAction(
         period=10.0,
         actions=[
             battery_launch,
             lights_launch,
-            robot_localization_node,
             manager_launch,
+            ekf_launch,
```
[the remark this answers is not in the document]
pawelirh: @rafal-gorecki rabbit is right, could you add a short comment to brief the goal delaying those nodes?

panther_bringup/launch/bringup.launch.py
rafal-gorecki: With gps it's `robot_localization` pkg still use Kalman Filter, so is the most correct one name should be related with **kalman filter**, as we use extended kalman filter and there is no plan to change it right now `ekf` is the most concrete name for this pkg. In future of course we can change the name if we decide to use other filtration metod.

`localization` in my opinion, is too broad a word and may as well catching the AMCL algorithm there and I not sure if this is what we want.

panther_ekf/CMakeLists.txt
pawelirh, line 1: The name of the package directory stayed unchanged

pawelirh (changes requested):

panther_gazebo/launch/simulate_multiple_robots.launch.py
```diff
@@ -0,0 +1,120 @@
...
+        simulate_robot.append(group)
+
+    return LaunchDescription(
+        [
+            declare_add_world_transform_arg,
+            declare_robots_arg,
+            SetUseSimTime(True),
+            *simulate_robot,
+        ]
+    )
```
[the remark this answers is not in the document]
rafal-gorecki: it's the same content xd

pawelirh (approved):

PR #333, opened 2024-06-06
ROS 2 remove obsolate tests

KmakD:
### Description

Remove tests that are no longer necessary. The type mismatch check is performed during tree creation inside factory. This tests do not check any part of our code so there is no point having them. 

Also, remove unnecessary brackets inside `lights.xml`

### Modifications

-

## Summary by CodeRabbit

- **Refactor**
  - Simplified logical expression for setting animations to improve readability.

- **Tests**
  - Removed test cases for incorrect service values to streamline the test suite.

### DIFF
--- a/panther_manager/behavior_trees/lights.xml
+++ b/panther_manager/behavior_trees/lights.xml
@@ -49,7 +49,7 @@ current_anim_id = CHARGING_BATTERY_ANIM_ID"/>
                                     param=""
                                     repeating="true"
                                     service_name="lights/controller/set/animation"
-                                    _skipIf="(!e_stop_state) || (current_anim_id == E_STOP_ANIM_ID)"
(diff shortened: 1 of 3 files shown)


panther_manager/behavior_trees/lights.xml
pawelirh, line 1: If you aimed to reverse all changes in BT XML files, there are still changes to make ->
https://github.com/husarion/panther_ros/pull/319/files#diff-3705029f3f8b5eaabd76e1f9edb8bd63d3999e98c6b378882af7be3e52b85620

KmakD: Not exactly. For some reason after update if you have negation before or after "OR" statement you need to put it in brackets. So this changes are correct but brackets in "EQUAL" expression are not necessary.

PR #351, opened 2024-06-18
Add EKF GPS configuration

rafal-gorecki:
### Description

Add EKF GPS configuration

### Modifications

- logic for process GPS data to odometry output
- `odom/wheels` -> `odometry/wheels` - more consistant (`odometry/filtered/global`, `odometry/filtered/local`, `odometry/wheels`)
- fix IMU ENU orientation

## Summary by CodeRabbit

- **New Features**
  - Added configurations for GPS fusion and updated localization modes.
  - Introduced new launch file for localization with dynamic configuration options.
  - Enhanced localization system with new parameters for EKF and navsat transform nodes.

- **Bug Fixes**
  - Updated topic mappings for odometry and IMU data.

- **Documentation**
  - Improved README to reflect new localization features and configurations.

- **Refactor**
  - Renamed and restructured several configuration files for clarity and consistency.

- **Chores**
  - Updated paths and references in launch files for better organization and maintainability.


panther_bringup/README.md
```diff
@@ -63,7 +63,7 @@ Extended Kalman Filter node for more accurate odometry. For more information, re
 
 [//]: # (ROS_API_NODE_SUBSCRIBERS_START)
 
-- `~/odom/wheels` [*nav_msgs/Odometry*]: robot odometry calculated from wheels.
+- `~/odometry/wheels` [*nav_msgs/Odometry*]: robot odometry calculated from wheels.
```
pawelirh, line 66: The `~/odometry..` means `/namespace` and `/node_name` are attached as a prefix. If our goal is to specify that only a `/namespace` is added -> we should use `odometry...` 

We're mistaken in most readme files in our stack, but we can fix it iteratively once a specific readme is "touched".

panther_localization/launch/ekf.launch.py
```diff
@@ -50,27 +62,60 @@ def generate_launch_description():
...
+        parameters=[ekf_config_path, {"tf_prefix": namespace}],
+        namespace=namespace,
+        remappings=[
+            ("gps/fix", "gps/fix"),
+            ("odometry/filtered", "odometry/filtered/global"),
+            ("odometry/gps", "_odometry/gps"),
         ],
-        emulate_tty=True,
+        condition=LaunchConfigurationEquals("ekf_configuration", "global"),
     )
```
pawelirh: Do we need a 3x node definition? Underneath these are 2 different executables. Wouldn't be better to define one `ekf_node` with two configs or one config for two configurations?

panther_localization/README.md
pawelirh, line 1: This file format seems to be a bit different from others. Can we try to stick somehow similar to other readme files in the `panther_ros` repo?

pawelirh, line 1: Same issue as mentioned in the launch file-related comment. Having defined two different nodes makes this file cleaner as well.

pawelirh (changes requested):

panther_bringup/README.md
rafal-gorecki: Agree and should be changed in all our readme. I'm not sure if this PR is appropriate place for that changes.

panther_localization/launch/ekf.launch.py
rafal-gorecki: This is how it is usually done:
image

There are some alternatives:

**1. Single `ekf_node` with arg `ekf_mode:=global/local`.**
In such case the `/odometry/filtered` will contain:
- relative position to start point  for `local` arg (x:=5 means move 5m stright).
- adds + imu orientation + gps to calculate ENU position for `global` (x:=5 means move 5m to the east).

**2. Two nodes `ekf_local`, `ekf_global` (as it is) with arg `ekf_mode:=global/local`.**
In such case:
- setting arg to `local` creates `/odometry/filtered/local` which contain relative position to the init pose
- setting arg to `global` add new `ekf_global` node which uses `/odometry/filtered/local` + imu orientation + gps to calculate ENU position

**3. Two nodes `ekf_relative`, `ekf_enu` and two type of arg `gps_data:=true/false` `ekf_mode:=relative/enu/both`.** 
This will most likely result in you having 4 ekf configurations and switching them accordingly:
- `ekf_relative_with_gps.yaml`
- `ekf_relative.yaml`
- `ekf_enu_with_gps.yaml`
- `ekf_enu.yaml` 
- `both` will crate `odom_enu` -> `odom_relative` -> `base_link` frames

Probably 3. is the best, but I need more time to create all configurations and test it.
Let me know what do you think.

panther_localization/README.md
rafal-gorecki: Our docs need much bigger refactor. The refactor of our Readme files is one of the next comming task. So probably there will be many changes. I'm not sure we should conform to a standard that isn't working. 

My main point is that: 
- for convenient linking, we should not have headers with the same names
- source code should not adapt to the parser, but the parser should adapt to the readme format
- there is one place missing where the entire API is briefly described, not just a fragment in each package. For the target customer, a short API download is more important.

For these reasons, I don't know if there is time to devote time to something that will have to be changed anyway.

panther_bringup/README.md
pawelirh: I'm not saying you should change all occurrences of this issue. Just proposed to make it right in readme files that you already edited so we can fix it successively.

panther_localization/README.md
pawelirh: If it's already in our pipeline, that's fine 👌🏼

panther_localization/launch/ekf.launch.py
pawelirh: Ok, now I got you more precise, and the proposed launch makes more sense 👍🏼 
I wonder if approach no.3 isn't an overhead?

One thing that came to my mind you can consider is leaving the config file as it is now and defining two separate launch files:
1. `ekf.launch.py` - local EKF
2. `gps_localization.launch.py`/`global_ekf.launch.py` (or any appropriate name) - local + global + navsat

Seems to me more intuitive from a user perspective but demands more work maintaining the files. The decision is yours here.

pawelirh (approved):

PR #364, opened 2024-07-08
ROS 2 update release workflow

KmakD:
### Description

[Summary of the changes]

### Modifications

-

## Summary by CodeRabbit

- **Chores**
  - Added an automated workflow to build and publish Panther flash OS image for Ubuntu 22.04 under specific conditions.


.github/workflows/release-candidate.yaml
pawelirh, line 1: Not sure if it's needed to generate the flash image on the release candidate stage 🤔
Do you consider we're gonna use it?

KmakD: I thought that we may also want to flash new image using USB in test phase, in case robots rear cover is not disassembled.

pawelirh (approved):

PR #365, opened 2024-07-15
ROS 2 lights initialization

KmakD:
### Description

- image messages are now passed using intra process communication
- nodes were changed to components
- initialization was moved to constructor, new structure representing LED control status was introduced
- some files were renamed

### Modifications

-

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved LED control handling to enhance stability and reliability.
  
- **Refactor**
  - Streamlined the internal workings of the `DriverNode` class to improve performance and maintainability.

- **New Features**
  - Added functionality to better diagnose light status and control LEDs more effectively.


panther_lights/launch/lights.launch.py
```diff
@@ -58,26 +59,31 @@ def generate_launch_description():
...
+                plugin="panther_lights::DriverNode",
+                name="lights_driver",
+                namespace=namespace,
+                remappings=[("/diagnostics", "diagnostics")],
+                condition=UnlessCondition(use_sim),
+            ),
+            ComposableNode(
+                package="panther_lights",
+                plugin="panther_lights::ControllerNode",
+                name="lights_controller",
```
pawelirh, line 82: I know this is not an objective of this task but I've spotted the debris after node renaming in file `led_animations_queue.cpp` line 121. Could you change it?

panther_lights/launch/lights.launch.py
```diff
@@ -58,26 +59,31 @@ def generate_launch_description():
...
+        executable="component_container",
+        composable_node_descriptions=[
+            ComposableNode(
+                package="panther_lights",
+                plugin="panther_lights::DriverNode",
+                name="lights_driver",
+                namespace=namespace,
+                remappings=[("/diagnostics", "diagnostics")],
+                condition=UnlessCondition(use_sim),
+            ),
```
pawelirh: We can specify intra-process comm here with

```python
        extra_arguments=[
            {"use_intra_process_comms": True},
        ],
```

panther_lights/include/panther_lights/lights_controller_node.hpp
```diff
@@ -42,7 +42,7 @@ class ControllerNode : public rclcpp::Node
 {
 public:
   ControllerNode(
-    const std::string & node_name, const rclcpp::NodeOptions & options = rclcpp::NodeOptions());
+    const rclcpp::NodeOptions & options = rclcpp::NodeOptions().use_intra_process_comms(true));
```
pawelirh: It'd be better not to hardcode the intra-process communication here. We can use arguments in the launch file to specify this configuration.

panther_lights/include/panther_lights/lights_driver_node.hpp
```diff
@@ -35,27 +35,28 @@ using ImageMsg = sensor_msgs::msg::Image;
...
-    const std::string & node_name, const rclcpp::NodeOptions & options = rclcpp::NodeOptions());
-
-  /**
-   * @brief Initializes the ImageTransport subscribers for to subscribe frames. IMPORTANT: this must
-   * be invoked after the node class is instantiated to establish whole functionality.
-   *
-   * @param it ImageTransport object used to create subscribers for Image topics.
-   */
-  void InitializeSubscribers(const std::shared_ptr<image_transport::ImageTransport> & it);
+    const rclcpp::NodeOptions & options = rclcpp::NodeOptions().use_intra_process_comms(true));
```
pawelirh: Same as above regarding the intra-process configuration.

panther_lights/src/lights_driver_node.cpp
```diff
@@ -59,45 +61,41 @@ DriverNode::DriverNode(const std::string & node_name, const rclcpp::NodeOptions
...
+
+  RCLCPP_INFO(this->get_logger(), "Node constructed successfully.");
 }
 
 void DriverNode::OnShutdown()
 {
-  ClearLEDs();
-
-  if (led_control_granted_) {
-    ToggleLEDControl(false);
```
pawelirh: If the lights driver shut down but the rest of the system operates as before, the LED stays unnecessarily occupied.

panther_lights/src/lights_driver_node.cpp
```diff
@@ -109,6 +107,12 @@ void DriverNode::ClearLEDs()
 
 void DriverNode::ToggleLEDControl(const bool enable)
 {
+  RCLCPP_DEBUG(
+    this->get_logger(), "Calling service to toggle LED control to '%s'.",
+    enable ? "true" : "false");
+
+  led_control_status_ = LEDControlStatus::PENDING;
```
pawelirh: If there is no response from the server, the client node is locked in a `PENDING` state.

panther_lights/include/panther_lights/lights_driver_node.hpp
```diff
@@ -35,27 +35,28 @@ using ImageMsg = sensor_msgs::msg::Image;
 using SetBoolSrv = std_srvs::srv::SetBool;
 using SetLEDBrightnessSrv = panther_msgs::srv::SetLEDBrightness;
 
+enum LEDControlStatus {
+  IDLE = 0,
+  PENDING,
+  GRANTED,
+  NOT_GRANTED,
+  ERROR,
+};
```
pawelirh: Not sure if it's necessary to maintain such a complex struct since we test `led_control_status_` only against `GRANTED` and `PENDING`

pawelirh (changes requested):

panther_lights/src/lights_driver_node.cpp
KmakD: When whole system is shut down, the service is not functional and this causes process to freeze here and wait for SIGKILL signal. I guess it's still better to put it here for scenario you mentioned.

panther_lights/src/lights_driver_node.cpp
KmakD: You suggest adding a timeout for server response?

panther_lights/include/panther_lights/lights_driver_node.hpp
KmakD: There are actually 3 states that are tested: GRANTED, PENDING and (IDLE, NOT_GRANTED, ERROR - all together). I've added more states to be more descriptive and it could be useful for debugging or diagnostics later

panther_lights/src/lights_driver_node.cpp
pawelirh: Ad. `ToggleLEDControl(false)` - There is `wait for service`, so if a service is already unavailable, after a timeout the exception is thrown. We could add a try-catch block to ensure that each cleanup method is properly executed (possibly in the future someone adds something after `ToggleLEDControl` and we may encounter incomplete execution of a cleanup process).

Ad. `ClearLEDs()` - this cleans the state of an internal class attribute and doesn't need to depend on the fact if control is granted or not. 

Putting things in the code it may look somehow like: 

```cpp
void DriverNode::OnShutdown()
{
  ClearLEDs();

  if (led_control_granted_) {
    try {
      ToggleLEDControl(false);
      RCLCPP_DEBUG(this->get_logger(), "Couldn't revoke LED control during shutdown: %s", e.what());
    }
  }
}
```

panther_lights/src/lights_driver_node.cpp
pawelirh: We would need to choose either checking timeout (most likely on a timer?) or sending the request until the control is granted without checking the PENDING state. The first option is way more elegant but demands more changes.

panther_lights/src/lights_driver_node.cpp
```diff
@@ -235,7 +236,10 @@ void DriverNode::DiagnoseLights(diagnostic_updater::DiagnosticStatusWrapper & st
...
   std::string message{"Control over LEDs granted."};
 
-  if (led_control_status_ != LEDControlStatus::GRANTED) {
+  if (
+    led_control_status_ != LEDControlStatus::GRANTED &&
+    led_control_status_ != LEDControlStatus::PENDING &&
+    led_control_status_ != LEDControlStatus::IDLE) {
     error_level = diagnostic_updater::DiagnosticStatusWrapper::ERROR;
     message = "Control over LEDs not granted, driver is not functional!";
   }
```
pawelirh: If `PENDING` or `IDLE` the control is not yet granted so we shouldn't `std::string message{"Control over LEDs granted."};` .

panther_lights/include/panther_lights/lights_driver_node.hpp
pawelirh: If we want to make use of these states in the current shape of the code, the only option is to add a `control_status` field on the diagnostics channel. (assuming that the newest logic on determining `OK` vs `ERROR` on diagnostics is changes)

panther_lights/src/lights_driver_node.cpp
KmakD: Ad. ToggleLEDControl(false) - actually quite weird thing happens here. The method does not throw, the program just freezes here I'm not sure what causes this. Sooo, try catch won't help :confused: 
Ad. ClearLEDs() - it sends buffer to SPI, but it makes sense to clear spi buffer on shutdown anyway

panther_lights/src/lights_driver_node.cpp
```diff
@@ -41,7 +41,9 @@ using std::placeholders::_2;
 
 DriverNode::DriverNode(const rclcpp::NodeOptions & options)
 : Node("lights_driver", options),
-  led_control_status_(LEDControlStatus::IDLE),
+  led_control_status_(LEDControlStatus::NOT_GRANTED),
+  initialization_attempt_(0),
+  led_control_call_time_(this->now()),
```
pawelirh: There is no led control request until the service is called, what is the goal of setting call time here? 🤔

panther_lights/src/lights_driver_node.cpp
```diff
@@ -70,6 +72,10 @@ DriverNode::DriverNode(const rclcpp::NodeOptions & options)
   set_brightness_server_ = this->create_service<SetLEDBrightnessSrv>(
     "lights/set_brightness", std::bind(&DriverNode::SetBrightnessCB, this, _1, _2));
 
+  // running at 10 Hz
+  initialization_timer_ = this->create_wall_timer(
+    std::chrono::milliseconds(100), std::bind(&DriverNode::InitializationTimerCB, this));
```
pawelirh, line 77: Isn't it too frequent for a timer that is not turned off and works during the node's whole lifecycle?

panther_lights/src/lights_driver_node.cpp
```diff
@@ -112,22 +143,24 @@ void DriverNode::ToggleLEDControl(const bool enable)
...
+    return;
   }
 
   enable_led_control_client_->async_send_request(
     request, std::bind(&DriverNode::ToggleLEDControlCB, this, std::placeholders::_1));
 
   RCLCPP_DEBUG(
     this->get_logger(), "Sent request toggling LED control to '%s'.", enable ? "true" : "false");
+  led_control_status_ = LEDControlStatus::PENDING;
+  led_control_call_time_ = this->now();
```
pawelirh: Please change the order so `RCLCPP_DEBUG` is at the end.

panther_lights/src/lights_driver_node.cpp
```diff
@@ -140,38 +173,28 @@ void DriverNode::ToggleLEDControlCB(rclcpp::Client<SetBoolSrv>::SharedFutureWith
...
-        this->get_logger(), *this->get_clock(), 5000,
-        "Waiting for LED control to be granted. Ignoring frame for " << panel_name << "!");
-      return;
-    }
-
-    RCLCPP_WARN_STREAM_THROTTLE(
-      this->get_logger(), *this->get_clock(), 5000, "LED control not granted. Retrying.");
-    ToggleLEDControl(true);
-
+  if (led_control_status_ == LEDControlStatus::PENDING) {
```
pawelirh: What if the state is still `NOT_GRANTED`?

panther_lights/src/lights_driver_node.cpp
```diff
@@ -140,38 +173,28 @@ void DriverNode::ToggleLEDControlCB(rclcpp::Client<SetBoolSrv>::SharedFutureWith
   const auto response = result.second;
 
   if (!response->success) {
-    RCLCPP_ERROR_THROTTLE(
-      this->get_logger(), *this->get_clock(), 5000, "Failed to toggle LED control.");
-    led_control_status_ = LEDControlStatus::ERROR;
+    RCLCPP_ERROR(this->get_logger(), "Failed to toggle LED control.");
+    led_control_status_ = LEDControlStatus::NOT_GRANTED;
```
pawelirh: The state here is dependent if we want to turn on or off the control.

pawelirh (changes requested):

pawelirh (approved):

PR #391, opened 2024-08-19
LEDStrip plugin to Gazebo

rafal-gorecki:
### Description

Create led strip plugin to gazebo simulation.

### Modifications

- Simpolified LedStrip configuration by using gazebo system plugin and increase robustness.
- Add LEDStrip documentation.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `LEDStrip` class for enhanced control of LED strips within the Gazebo simulation.
	- Added support for configurable light parameters, including a communication topic for dynamic control.

- **Bug Fixes**
	- Updated topic naming conventions to improve clarity and organization in communication between ROS and Gazebo.

- **Documentation**
	- Removed outdated references to configuration files and added new sections detailing the `LEDStrip` plugin.

- **Chores**
	- Restructured dependencies in the package configuration for better runtime management.


panther_gazebo/README.md
```diff
@@ -48,3 +47,29 @@ Plugin based on `ign_system` is responsible for handling sensor interfaces (only
...
+
+- `{topic}` [*gz::msgs::Image*]: Subscribes to an image message for visualization. The topic is specified via a parameter.
+
+#### Service Servers
+
+- `/marker` [*gz::msgs::Marker*]: Service to request markers for visualizing the received image.
+
+#### Parameters
+
+The following parameters are required when including this interface in the URDF (you can refer to the [gazebo.urdf.xacro](../panther_description/urdf/gazebo.urdf.xacro) file for details).
```
delihus, line 68: Should this link be absolute?

panther_gazebo/config/gz_bridge.yaml
```diff
@@ -17,12 +17,14 @@
   gz_type_name: ignition.msgs.Twist
   direction: GZ_TO_ROS
 
-- topic_name: lights/channel_1_frame
+- ros_topic_name: lights/channel_1_frame
```
delihus, line 20: Is a namespace for ROS topic passed  thanks to `namespace` argument in launch?

panther_gazebo/include/panther_gazebo/gui/e_stop.hpp
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef PANTHER_GAZEBO_PLUGINS_ESTOP_HPP_
-#define PANTHER_GAZEBO_PLUGINS_ESTOP_HPP_
+#ifndef PANTHER_GAZEBO_E_STOP_HPP_
```
delihus:
```suggestion
#ifndef PANTHER_GAZEBO_GUI_E_STOP_HPP_
```
Please add folder name to the header guard.

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,104 @@
...
+// limitations under the License.
+
+#ifndef PANTHER_GAZEBO_LED_STRIP_HPP_
+#define PANTHER_GAZEBO_LED_STRIP_HPP_
+
+#include <chrono>
+#include <mutex>
+#include <string>
+
+#include <gz/msgs/image.pb.h>
```
delihus: Please separate msgs headers under external libraries.

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,104 @@
...
+   * @param id The unique ID of the marker (if not unique, the marker will replace the existing one)
+   * @param pose The pose of the marker
+   * @param color The color of the marker
+   * @param size The size of the marker
+   */
+  void CreateMarker(
+    const uint id, const gz::math::Pose3d pose, const gz::math::Color & color,
+    const gz::math::Vector3d size);
+
+  // Parameters
```
delihus: Please add to every field `_`.

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include <gz/transport/Node.hh>
+#include <sdf/Element.hh>
+
+namespace panther_gazebo
+{
+
+/**
+ * @brief Class to manage an LED strip in a Gazebo simulation based on received image
+ */
+class LEDStrip : public gz::sim::System,
```
delihus, line 42: Please add documentation for every public method.

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "panther_gazebo/led_strip.hpp"
+
+#include <gz/msgs/color.pb.h>
```
delihus: separate msgs includes

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+#include <gz/msgs/marker.pb.h>
+#include <gz/plugin/Register.hh>
+#include <gz/sim/Model.hh>
+#include <gz/sim/components/LightCmd.hh>
+#include <gz/sim/components/Name.hh>
+
+namespace panther_gazebo
+{
+
+LEDStrip::LEDStrip() : gz::sim::System() {}
```
delihus: This constructor and the destructor are necessary?

They look like default.

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+LEDStrip::~LEDStrip() = default;
+
+void LEDStrip::Configure(
+  const gz::sim::Entity & entity, const std::shared_ptr<const sdf::Element> & sdf,
+  gz::sim::EntityComponentManager & ecm, gz::sim::EventManager &)
```
delihus: Please add a name of an unused argument in a comment .

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+{
+  const auto model = gz::gazebo::Model(entity);
+  if (!model.Valid(ecm)) {
+    throw std::runtime_error(
+      "Error: Failed to initialize because [" + model.Name(ecm) +
+      "] (Entity=" + std::to_string(entity) +
+      ") is not a model. Please make sure that LEDStrip is attached to a valid model.");
+    return;
```
delihus:
```suggestion
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+  }
+}
+
+void LEDStrip::ParseParameters(const std::shared_ptr<const sdf::Element> & sdf)
+{
+  if (sdf->HasElement("light_name")) {
+    light_name = sdf->Get<std::string>("light_name");
+  } else {
+    throw std::runtime_error("Error: The light_name parameter is missing.");
+  }
```
delihus: Let's reduce brackets and elses. If it is possible to throw an error in one if statement let's do this.
```suggestion
  if (!sdf->HasElement("light_name")) {
    throw std::runtime_error("Error: The light_name parameter is missing.");
  } 
  
  light_name = sdf->Get<std::string>("light_name");
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+  if (sdf->HasElement("namespace")) {
```
delihus: Shall we add a warning log when a `namespace` is not provided?

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+    ns = sdf->Get<std::string>("namespace");
+  }
+
+  if (sdf->HasElement("topic")) {
+    image_topic = sdf->Get<std::string>("topic");
+  } else {
+    throw std::runtime_error("Error: The topic parameter is missing.");
+  }
```
delihus:
```suggestion
  if (!sdf->HasElement("topic")) {
   throw std::runtime_error("Error: The topic parameter is missing.");
  }
  image_topic = sdf->Get<std::string>("topic");
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+  if (sdf->HasElement("frequency")) {
```
delihus: This ifs look generic. I suggest to add template function.

Maybe something like this:
```c++
template<typename T>
bool GetParameterIfExists(const std::string parameter_name, T &value){
  if (!sdf->HasElement(parameter_name)) {
    return false;
  }
  value = sdf->Get<T>("width");
  return true;
  }
```

It is up to you.

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+    [&](
+      const gz::sim::Entity & entity, const gz::sim::components::Name * name,
+      const gz::sim::components::Light * light_component) -> bool {
+      if (name->Data() == light_name) {
+        light_entity = entity;
+        igndbg << "Light entity found: " << light_entity << std::endl;
+
+        // Ensure the LightCmd component is created
+        if (!ecm.Component<gz::sim::components::LightCmd>(light_entity)) {
+          sdf::Light light_sdf = light_component->Data();
```
delihus: This should be in another function. E.g. `CreateLightCommandFromSDF`

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+              break;
+            default:
+              light_cmd.set_type(gz::msgs::Light::POINT);
+              break;
+          }
+
+          ecm.CreateComponent(light_entity, gz::sim::components::LightCmd(light_cmd));
+          igndbg << "Created LightCmd component for entity: " << light_entity << std::endl;
+        }
+        return true;  // Stop searching
```
delihus: I don't get this `return`. There is no `while` or `for` to stop.

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+  node.Subscribe(ns + "/" + image_topic, &LEDStrip::ImageCallback, this);
+}
+
+void LEDStrip::PreUpdate(const gz::sim::UpdateInfo & info, gz::sim::EntityComponentManager & ecm)
+{
+  auto current_time = info.simTime;
+
+  auto period = std::chrono::milliseconds(static_cast<int>(1000 / frequency));
+  if (new_image_available && current_time - last_update_time >= period) {
```
delihus:
```suggestion
  if (new_image_available && current_time - last_update_time < period) {
  return;
  }
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+      new_image_available = false;
+    }
+
+    VisualizeLights(ecm, image);
+
+    auto light_pose = ecm.Component<gz::sim::components::Pose>(light_entity)->Data();
+    VisualizeMarkers(image, light_pose);
+
+    last_update_time = current_time;
+  }
```
delihus:
```suggestion
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+}
+
+void LEDStrip::PreUpdate(const gz::sim::UpdateInfo & info, gz::sim::EntityComponentManager & ecm)
+{
+  auto current_time = info.simTime;
+
+  auto period = std::chrono::milliseconds(static_cast<int>(1000 / frequency));
+  if (new_image_available && current_time - last_update_time >= period) {
+    gz::msgs::Image image;
+    {
```
delihus, line 55: Here I see the lack of _ in the names of fields of a class. I don't see the reason of the mutex. I have to go back to .hpp and check what is a field.

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+      return true;
+    });
+
+  if (light_entity == gz::sim::kNullEntity) {
+    ignerr << "Error: Light entity not found." << std::endl;
+    return;
+  }
+}
+
+void LEDStrip::MsgValidation(const gz::msgs::Image & msg)
```
delihus: This if function with throw sounds overkill. I suggest to create function `bool IsMsgValid(const gz::msgs::Image & ms)`  and change ImageCallback

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+void LEDStrip::ImageCallback(const gz::msgs::Image & msg)
+{
+  try {
+    MsgValidation(msg);
+    std::lock_guard<std::mutex> lock(image_mutex);
+    last_image = msg;
+    new_image_available = true;
+  } catch (const std::exception & e) {
+    ignerr << "Error: " << e.what() << std::endl;
+  }
```
delihus:
```suggestion
if(!IsMsgValid(msg))
    ignerr << "Error: " << e.what() << std::endl;
  }
  
  std::lock_guard<std::mutex> lock(image_mutex);
  last_image = msg;
  new_image_available = true;
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+{
+  int sum_r = 0, sum_g = 0, sum_b = 0, sum_a = 0;
+  int pixel_count = msg.width() * msg.height();
+
+  const std::string & data = msg.data();
+  bool is_rgba = (msg.pixel_format_type() == gz::msgs::PixelFormatType::RGBA_INT8);
+  int step = is_rgba ? 4 : 3;
+
+  for (int i = 0; i < pixel_count * step; i += step) {
+    sum_r += static_cast<unsigned char>(data[i]);
```
delihus: What do you think about `std::uint8_t`?

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+    sum_b += static_cast<unsigned char>(data[i + 2]);
+    if (is_rgba) {
+      sum_a += static_cast<unsigned char>(data[i + 3]);
+    }
+  }
+
+  int mean_r = sum_r / pixel_count;
+  int mean_g = sum_g / pixel_count;
+  int mean_b = sum_b / pixel_count;
+  int mean_a = is_rgba ? sum_a / pixel_count : 255;
```
delihus: What if image is 16bit? 

I would create an if statement at the begging of the function to check if the message type is gz::msgs::PixelFormatType::RGBA_INT8  or gz::msgs::PixelFormatType::RGB_INT8. Then  instead of 255 I would  write `std::numeric_limits<uint8_t>::max()`

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+  auto light_on = light_cmd.mutable_header()->add_data();
+  light_on->set_key("isLightOn");
+  light_on->add_value()->assign("1");
+
+  auto visualize = light_cmd.mutable_header()->add_data();
+  visualize->set_key("visualizeVisual");
+  visualize->add_value()->assign("0");
+
+  // Update the light command component with the new light message
```
delihus:
```suggestion
```
Is it necessary?

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+}
+
+void LEDStrip::VisualizeMarkers(const gz::msgs::Image & image, const gz::math::Pose3d & light_pose)
+{
+  const std::string & data = image.data();
+
+  double step_width = marker_width / image.width();
+  double step_height = marker_height / image.height();
+
+  bool is_rgba = (image.pixel_format_type() == gz::msgs::PixelFormatType::RGBA_INT8);
```
delihus, line 225: The same as above with the types of a message

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+  int step = is_rgba ? 4 : 3;
+
+  const auto pixel_size = gz::math::Vector3d(0.001, step_width, step_height);
+
+  for (uint y = 0; y < image.height(); ++y) {
+    for (uint x = 0; x < image.width(); ++x) {
+      uint idx = (y * image.width() + x) * step;
```
delihus:
```suggestion
  for (std::size_t y = 0; y < image.height(); ++y) {
    for (std::size_t x = 0; x < image.width(); ++x) {
      std::size_t idx = (y * image.width() + x) * step;
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+      auto pixel_color = gz::math::Color(
+        static_cast<float>(static_cast<unsigned char>(data[idx])) / 255.0f,
```
delihus: This casting waterfall does not look very well... Maybe a new generic function?
`NormalizeData(data, range)`

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+  marker_msg.set_type(gz::msgs::Marker::BOX);
+
+  gz::msgs::Set(marker_msg.mutable_pose(), pose);
+  gz::msgs::Set(marker_msg.mutable_scale(), size);
+
+  gz::msgs::Set(marker_msg.mutable_material()->mutable_ambient(), color);
+  gz::msgs::Set(marker_msg.mutable_material()->mutable_diffuse(), color);
+
+  // Using Request to ensure markers are visible
+  node.Request("/marker", marker_msg);
```
delihus: What about when there are 2 robots? Should it be namespaced?

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+          light_cmd.set_attenuation_linear(light_sdf.LinearAttenuationFactor());
+          light_cmd.set_attenuation_quadratic(light_sdf.QuadraticAttenuationFactor());
+          light_cmd.set_intensity(light_sdf.Intensity());
+
+          gz::msgs::Set(light_cmd.mutable_diffuse(), light_sdf.Diffuse());
+          gz::msgs::Set(light_cmd.mutable_specular(), light_sdf.Specular());
+          gz::msgs::Set(light_cmd.mutable_direction(), light_sdf.Direction());
+
+          // Set the light type
+          switch (light_sdf.Type()) {
```
delihus: Maybe new converting function?

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+    last_update_time = current_time;
+  }
+}
+
+void LEDStrip::ImageCallback(const gz::msgs::Image & msg)
+{
+  try {
+    MsgValidation(msg);
+    std::lock_guard<std::mutex> lock(image_mutex);
```
delihus: What about using `reatime_box`  instead of mutex, guard and bool fields?
https://github.com/ros-controls/realtime_tools/blob/master/include/realtime_tools/realtime_box.h

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+  // Parameters
+  std::string light_name;
+  std::string image_topic;
+  std::string ns = "";
+  double frequency = 10.0;
+  double marker_width = 1.0;
+  double marker_height = 1.0;
+
+  bool new_image_available{false};
```
delihus: why not` bool new_image_available = false;`?

delihus (changes requested):
Additionally I don't know if this PR generates this warning but you @rafal-gorecki  know the simulation best.
image

image

delihus:
What about tests for this plugin?

panther_gazebo/README.md
rafal-gorecki: If you think about docs_new, parser is able to handle that.

panther_gazebo/include/panther_gazebo/gui/e_stop.hpp
rafal-gorecki: ok

panther_gazebo/include/panther_gazebo/led_strip.hpp
rafal-gorecki: changed

panther_gazebo/include/panther_gazebo/led_strip.hpp
rafal-gorecki: Ok

panther_gazebo/include/panther_gazebo/led_strip.hpp
rafal-gorecki: ok

panther_gazebo/src/led_strip.cpp
rafal-gorecki: `marker` is a gz sim topic and have two field to distinguish two robots from each other:

```cpp
marker_msg.set_ns(ns_ + light_name_);
if (ns_.empty()) { marker_msg.set_parent("panther"); } else { marker_msg.set_parent(ns_); }
 ```

Curently multiple robot want work. There is and error with creating multiple `light_driver` on single machine. I created task for this.

File: panther_gazebo/src/led_strip.cpp (hunk 23 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: There is a `MsgValidation(msg)` function to chceck only two option RGBA8 or RGB8.

File: panther_gazebo/src/led_strip.cpp (hunk 7 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: ok

File: panther_gazebo/src/led_strip.cpp (hunk 8 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: ok

File: panther_gazebo/src/led_strip.cpp (hunk 9 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: ok

File: panther_gazebo/src/led_strip.cpp (hunk 10 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: ok

File: panther_gazebo/src/led_strip.cpp (hunk 12 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: Namespace do not seem to be obligatory arg.

File: panther_gazebo/src/led_strip.cpp (hunk 19 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp:55]: `_` added

File: panther_gazebo/src/led_strip.cpp (hunk 15 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: Ok

File: panther_gazebo/src/led_strip.cpp (hunk 29 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: One converter seems enought

File: panther_gazebo/src/led_strip.cpp (hunk 16 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: there is `ecm.Each<gz::sim::components::Name, gz::sim::components::Light>` which works like iterator

File: panther_gazebo/src/led_strip.cpp (hunk 20 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: IMO, Thanks `try catch` it is easier to add additional checks and show appropriate error msg.

File: panther_gazebo/src/led_strip.cpp (hunk 22 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: ok

File: panther_gazebo/config/gz_bridge.yaml (hunk 2 above)
Comment (rafal-gorecki) [reply panther_gazebo/config/gz_bridge.yaml:20]: yes, but namespace is not added to the gazebo namesapce

File: panther_gazebo/src/led_strip.cpp (hunk 30 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: it doesn't seem to provide any benefits and it's another piece of non-standard C++ code that can make the function harder to understand. Besides, the name would suggest realtime, and here we don't really have anything related to realtime.

File: panther_gazebo/src/led_strip.cpp (hunk 21 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: in such case isMsgValid should still `throw` error

File: panther_gazebo/src/led_strip.cpp (hunk 11 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: ok

File: panther_gazebo/src/led_strip.cpp (hunk 17 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: I know it's the same thing, but it changes my mind completely.
I'd rather think that the timer **does something** when the timer is up than think that it **doesn't do anything** unless the timer is up.

File: panther_gazebo/src/led_strip.cpp (hunk 13 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: ok

File: panther_gazebo/src/led_strip.cpp (hunk 14 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: There is also need of `sdf` object. I change it a bit, but do not use seprate function.

File: panther_gazebo/src/led_strip.cpp (hunk 27 above)
Comment (rafal-gorecki) [reply panther_gazebo/src/led_strip.cpp]: `static_cast<float>` is unnecessary. Maybe that will be enough.

**Comment by rafal-gorecki** [comment]
> Additionally I don't know if this PR generates this warning but you @rafal-gorecki know the simulation best. image
> 
> image

1. The easiest way will be add 1 sec delay. I will check for nicer solution.
2. Ignition itself has a lot of warnings. You can feel it by running `ign gazebo empty.sdf -v 3`.
a) You can ignore it
b) Add -v 1 (display only errors)

File: panther_gazebo/include/panther_gazebo/led_strip.hpp [hunk 32]
```diff
@@ -0,0 +1,116 @@
...
+public:
+  /**
+   * @brief Configures the LED strip. This function fill up parameters and light_cmd_ based on URDF.
+   */
+  void Configure(
+    const gz::sim::Entity & id, const std::shared_ptr<const sdf::Element> & sdf,
+    gz::sim::EntityComponentManager & ecm, gz::sim::EventManager & eventMgr);
+  /**
+   * @brief Displays lights and markers, with specified by URDF frequency.
+   */
```
delihus:
```suggestion
  void Configure(
    const gz::sim::Entity & id, const std::shared_ptr<const sdf::Element> & sdf,
    gz::sim::EntityComponentManager & ecm, gz::sim::EventManager & eventMgr);
    
  /**
   * @brief Displays lights and markers, with specified by URDF frequency.
   */
```

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,116 @@
...
+{
+public:
+  /**
+   * @brief Configures the LED strip. This function fill up parameters and light_cmd_ based on URDF.
+   */
+  void Configure(
+    const gz::sim::Entity & id, const std::shared_ptr<const sdf::Element> & sdf,
+    gz::sim::EntityComponentManager & ecm, gz::sim::EventManager & eventMgr);
+  /**
+   * @brief Displays lights and markers, with specified by URDF frequency.
```
delihus: It will be clear to add to the comment also what is update and when is preupdate.

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * @brief Class to manage an LED strip in a Gazebo simulation based on received image.
+ */
+class LEDStrip : public gz::sim::System,
+                 public gz::sim::ISystemConfigure,
+                 public gz::sim::ISystemPreUpdate
+{
+public:
+  /**
+   * @brief Configures the LED strip. This function fill up parameters and light_cmd_ based on URDF.
```
delihus, line 48: There is no `@param` description

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,116 @@
...
+  /**
+   * @brief Displays lights and markers, with specified by URDF frequency.
+   */
+  void PreUpdate(const gz::sim::UpdateInfo & info, gz::sim::EntityComponentManager & ecm);
+
+private:
+  void ParseParameters(const std::shared_ptr<const sdf::Element> & sdf);
+
+  /**
+   * @brief Return Light command based on light configuration specified in URDF file Light
```
delihus, line 83: no `@param` and `@return `

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,116 @@
...
+   * properties
+   */
+  gz::msgs::Light SetupLightCmd(gz::sim::EntityComponentManager & ecm);
+
+  /**
+   * @brief Convert sdf::Light (configuration from URDF) to gz::msgs::Light (command msg)
```
delihus, line 93: no `@param` and `@return`

panther_gazebo/launch/simulation.launch.py
```diff
@@ -56,7 +56,7 @@ def generate_launch_description():
                 [FindPackageShare("husarion_gz_worlds"), "launch", "gz_sim.launch.py"]
             )
         ),
-        launch_arguments={"gz_gui": namespaced_gz_gui}.items(),
+        launch_arguments={"gz_gui": namespaced_gz_gui, "gz_log_level": "1"}.items(),
```
delihus, line 59: What is 1 log level? Do warnings appear?

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,279 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "panther_gazebo/led_strip.hpp"
+
+#include <stddef.h>
```
delihus:
```suggestion
#include <cstddf>
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+#include <gz/plugin/Register.hh>
+#include <gz/sim/Model.hh>
+#include <gz/sim/components/LightCmd.hh>
+#include <gz/sim/components/Name.hh>
+
+#include <gz/msgs/color.pb.h>
+#include <gz/msgs/marker.pb.h>
+namespace panther_gazebo
```
delihus:
```suggestion
#include <gz/msgs/color.pb.h>
#include <gz/msgs/marker.pb.h>

namespace panther_gazebo
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+  node_.Subscribe(ns_ + "/" + image_topic_, &LEDStrip::ImageCallback, this);
+}
+
+void LEDStrip::PreUpdate(const gz::sim::UpdateInfo & info, gz::sim::EntityComponentManager & ecm)
+{
+  auto current_time = info.simTime;
+  auto period = std::chrono::milliseconds(static_cast<int>(1000 / frequency_));
+
+  if (new_image_available_ && current_time - last_update_time_ >= period) {
```
delihus, line 53: Let's return function when statement is not met.  
```suggestion
  if (!new_image_available_ || (current_time - last_update_time_)  < period) {
  return;
  }
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,279 @@
...
+    VisualizeMarkers(image, light_pose);
+
+    last_update_time_ = current_time;
+  }
+}
+
+void LEDStrip::ParseParameters(const std::shared_ptr<const sdf::Element> & sdf)
+{
+  if (!sdf->HasElement("light_name")) {
+    ignerr << "Error: The light_name parameter is missing." << std::endl;
```
delihus: Here should be throw or return because you only log and then after all you`Get` the param.

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,279 @@
...
+    [&](
+      const gz::sim::Entity & entity, const gz::sim::components::Name * name,
+      const gz::sim::components::Light * light_component) -> bool {
+      if (name->Data() == light_name_) {
+        light_entity_ = entity;
+        igndbg << "Light entity found: " << light_entity_ << std::endl;
+        light_cmd = ConvertLight(light_component->Data());
+        return false;  // Stop searching
+      }
+      return true;  // Continue searching
```
delihus:
```suggestion
      if (name->Data() != light_name_) {
            return true;  // Continue searching
       }
      light_entity_ = entity;
      igndbg << "Light entity found: " << light_entity_ << std::endl;
      light_cmd = ConvertLight(light_component->Data());
      return false;  // Stop searching
      }

```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,279 @@
...
+    });
+
+  if (light_entity_ == gz::sim::kNullEntity) {
```
delihus, line 104: Should it throw an exception when there is no entity?

panther_gazebo/src/led_strip.cpp
delihus: Yes, I agree but `MsgValidation` has only one validation. A catch approach would be great when there are many different kind of errors but here is only one error. In that case I prefer to change this function to `bool ValidateEncoding(msg)` and remove try catch in `ImageCallback(msg)`.

panther_gazebo/src/led_strip.cpp
delihus: If so, please add a constant variable for `255` to refrain from the magic numbers.

panther_gazebo/src/led_strip.cpp
delihus: You could pass as an argument but one line if sounds okay.

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,279 @@
...
+    ignerr << "Error: Light entity not found. Return default light command msg." << std::endl;
+  }
+  return light_cmd;
+}
+
+gz::msgs::Light LEDStrip::ConvertLight(const sdf::Light & light_sdf)
```
delihus: Could we name this function  more precisely? `ConvertLightFromSdfToMsg` `CreateLightMsgFromSdf`?

panther_gazebo/src/led_strip.cpp
delihus: Okay, then it is up to you.

panther_gazebo/src/led_strip.cpp
delihus: Okay. Then let's throw.

panther_gazebo/src/led_strip.cpp
delihus: @coderabbit could you create an issue regarding to this?

panther_gazebo/src/led_strip.cpp
delihus: I think it looks cleaner but the implementation is up to you ;)
```cpp
void LEDStrip::ImageCallback(const gz::msgs::Image & msg)
{
  try {
    MsgValidation(msg);
    auto msg_ptr = std::make_shared<gz::msg::Image>(msg);
    received_image_msg_ptr_.set(std::move(msg_ptr));
  } catch (const std::exception & e) {
    ignerr << "Error: " << e.what() << std::endl;
  }
}
```

```cpp
void useImage(){
  std::shared_ptr<Twist> last_command_msg;
  received_velocity_msg_ptr_.get(last_command_msg);
  if (last_command_msg == nullptr)
  {
    //errror
    return;
  }

}
```

delihus (changes requested):

delihus:
@rafal-gorecki 
So let's add the delay

panther_gazebo/launch/simulation.launch.py
rafal-gorecki: No only errors. 0-4 (None, Errors, Warns, Info, Dbg)

panther_gazebo/src/led_strip.cpp
rafal-gorecki: I get fatal error: cstddf: No such file or directory
   17 | #include <cstddf>

panther_gazebo/src/led_strip.cpp
rafal-gorecki: But throw look like overkill. The simulation will work fine if no lights will loaded.

panther_gazebo/src/led_strip.cpp
rafal-gorecki: The simulation will be work but plugin won't so I'm not sure if it should be thrown.

rafal-gorecki: Can be

panther_gazebo/include/panther_gazebo/led_strip.hpp
rafal-gorecki: Add link to docs.

panther_gazebo/include/panther_gazebo/led_strip.hpp
rafal-gorecki: added

panther_gazebo/src/led_strip.cpp
delihus: Sorry my typo... `#include <cstddef>`

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,139 @@
...
+   * @param light_sdf Light SDF configuration
+   * @return Light command message
+   */
+  gz::msgs::Light CreateLightMsgFromSdf(const sdf::Light & light_sdf);
+  void ImageCallback(const gz::msgs::Image & msg);
+  bool IsEncodingValid(const gz::msgs::Image & msg);
+  gz::math::Color CalculateMeanColor(const gz::msgs::Image & msg);
+
+  /**
+   * @brief Manage color of robot simulated lights based on the image message
```
delihus, line 104: no `@params`

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,139 @@
...
+   */
+  void VisualizeLights(gz::sim::EntityComponentManager & ecm, const gz::msgs::Image & image);
+
+  /**
+   * @brief Manage color of robot LED strip based on the image message
+   */
+  void VisualizeMarkers(const gz::msgs::Image & image, const gz::math::Pose3d & lightPose);
```
delihus: no `@params`

panther_gazebo/include/panther_gazebo/led_strip.hpp
```diff
@@ -0,0 +1,139 @@
...
+   * @param id The unique ID of the marker (if not unique, the marker will replace the existing one)
+   * @param pose The pose of the marker
+   * @param color The color of the marker
+   * @param size The size of the marker
+   */
+  void CreateMarker(
+    const uint id, const gz::math::Pose3d pose, const gz::math::Color & color,
+    const gz::math::Vector3d size);
+
+  // Parameters
```
delihus:
```suggestion
```

panther_gazebo/src/led_strip.cpp
delihus: So if there are no lights declared we should some how turn off the plugin, right?

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,284 @@
...
+}
+
+void LEDStrip::VisualizeMarkers(const gz::msgs::Image & image, const gz::math::Pose3d & light_pose)
+{
+  const std::string & data = image.data();
+
+  double step_width = marker_width_ / image.width();
+  double step_height = marker_height_ / image.height();
+
+  float max_value = std::numeric_limits<uint8_t>::max();
```
delihus:
```suggestion
  const float max_value = std::numeric_limits<uint8_t>::max();
```

panther_gazebo/src/led_strip.cpp
```diff
@@ -0,0 +1,284 @@
...
+    sum_r += static_cast<uint8_t>(data[i]);
+    sum_g += static_cast<uint8_t>(data[i + 1]);
+    sum_b += static_cast<uint8_t>(data[i + 2]);
+    if (is_rgba) {
+      sum_a += static_cast<uint8_t>(data[i + 3]);
+    }
+  }
+
+  float max_value = std::numeric_limits<uint8_t>::max();
+  float norm_factor = max_value * pixel_count;
```
delihus:
```suggestion
  const float max_value = std::numeric_limits<uint8_t>::max();
  const float norm_factor = max_value * pixel_count;
```

delihus (changes requested):

panther_gazebo/launch/simulation.launch.py
delihus: I think we should show warnings.

panther_gazebo/src/led_strip.cpp
delihus: @coderabbitai  could you create an issue about that?

delihus (approved):
Thank you!

PR #401, opened 2024-09-02
Ros2 lynx gpio tests

pawelirh:
### Description

Mock `gpio_driver` code to be able to run tests in development environment

### Modifications

- Added a common interface for GPIODriver
- Created a factory method deciding on which version of GPIOController to create
- Moved types to a separate file to limit code lines in some files
- Restructured tests


panther_utils/include/panther_utils/common_utilities.hpp
```diff
@@ -104,6 +105,22 @@ std::fstream OpenFile(const std::string & file_path, const std::ios_base::openmo
...
+ * @param required_version The required version.
+ * @return true if the required version is satisfied, false otherwise.
+ */
+inline bool MeetsVersionRequirement(const float version, const float required_version)
+{
+  if (std::isnan(version) || std::isnan(required_version)) {
+    return false;
+  }
+
+  return version >= required_version - std::numeric_limits<float>::epsilon();
```
rafal-gorecki, line 121: Why not like this:

```suggestion
  return version >= required_version;
```

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system/gpio/gpio_controller.hpp
```diff
@@ -139,12 +142,19 @@ class GPIOControllerInterface
...
+  std::shared_ptr<GPIODriverInterface> gpio_driver_;
 };
 
 class GPIOControllerPTH12X : public GPIOControllerInterface
 {
 public:
+  /**
+   * @brief Constructor for GPIOControllerPTH12X class.
+   *
+   * @param gpio_driver Pointer to the GPIODriver object.
```
rafal-gorecki, line 154: Missing exception

panther_hardware_interfaces/include/panther_hardware_interfaces/panther_system/gpio/gpio_controller.hpp
```diff
@@ -274,6 +277,13 @@ class GPIOControllerPTH12X : public GPIOControllerInterface
 class GPIOControllerPTH10X : public GPIOControllerInterface
 {
 public:
+  /**
+   * @brief Constructor for GPIOControllerPTH10X class.
+   *
+   * @param gpio_driver Pointer to the GPIODriver object.
```
rafal-gorecki, line 284: Missing exception

panther_hardware_interfaces/test/panther_system/gpio/test_gpio_controller.cpp
```diff
@@ -57,118 +64,129 @@ class TestGPIOController : public ::testing::Test
...
-    throw std::runtime_error("Tests for this robot versions are not implemented");
-  }
+  gpio_driver_ = std::make_shared<MockGPIODriver>();
 
-  gpio_controller_wrapper_ = std::make_unique<GPIOControllerWrapper>();
+  // Mock methods called during the initialization process
+  ON_CALL(*gpio_driver_, SetPinValue(GPIOPin::VMOT_ON, true)).WillByDefault(testing::Return(true));
+  ON_CALL(*gpio_driver_, SetPinValue(GPIOPin::DRIVER_EN, true))
+    .WillByDefault(testing::Return(true));
+  ON_CALL(*gpio_driver_, IsPinAvailable(GPIOPin::WATCHDOG)).WillByDefault(testing::Return(true));
```
rafal-gorecki, line 80: Why watchdog state is not set and not checked in TestMotorsInitSuccess?

panther_hardware_interfaces/test/panther_system/gpio/test_gpio_controller.cpp
```diff
@@ -57,118 +64,129 @@ class TestGPIOController : public ::testing::Test
...
-  ASSERT_TRUE(gpio_controller_wrapper_->IsWatchdogEnabled());
+TEST(TestGPIOControllerInitialization, WatchdogPinNotAvailable)
+{
+  auto gpio_driver = std::make_shared<MockGPIODriver>();
 
-  std::this_thread::sleep_for(std::chrono::milliseconds(100));
-  gpio_controller_wrapper_->WatchdogDisable();
+  EXPECT_CALL(*gpio_driver, SetPinValue(testing::_, true))
+    .Times(2)
+    .WillRepeatedly(testing::Return(true));
```
rafal-gorecki, line 111: Why function SetPinValue is checked? I guess, it's only chcecked wrapper implemantation and does not check the our code.

panther_hardware_interfaces/test/panther_system/gpio/test_gpio_controller.cpp
```diff
@@ -13,36 +13,43 @@
...
+    (override));
+  MOCK_METHOD(
+    void, ChangePinDirection, (const GPIOPin pin, const gpiod::line::direction direction),
+    (override));
+  MOCK_METHOD(bool, IsPinAvailable, (const GPIOPin pin), (const, override));
+  MOCK_METHOD(bool, IsPinActive, (const GPIOPin pin), (override));
+  MOCK_METHOD(bool, SetPinValue, (const GPIOPin pin, const bool value), (override));
 };
 
 class GPIOControllerWrapper : public panther_hardware_interfaces::GPIOControllerPTH12X
```
rafal-gorecki, line 45: We decide to not check GPIOControler for older version?

rafal-gorecki (commented):
Some questions

panther_utils/include/panther_utils/common_utilities.hpp
pawelirh: It's the case of floating point arithmetics. IMO your suggestion is good too, but I didn't see a need to change it compared to the original version.

panther_hardware_interfaces/test/panther_system/gpio/test_gpio_controller.cpp
pawelirh: Since this pin is not set during the initialization

panther_hardware_interfaces/test/panther_system/gpio/test_gpio_controller.cpp
pawelirh: We don't check here. When we operate on mocks we define their behavior after specific methods are called. So here our goal is to tell the mock to return `true` twice when the `SetPinValue` method is invoked with any pin and `true` as a second parameter. The `SetPinValue` is invoked during the initialization of GPIOController (we don't use a test fixture so it's needed to do it manually).

PS. We could use the `ON_CALL` as well but it's known that initialization demands the `SetPinValue` method to be called twice and verify the flow by chance.

panther_hardware_interfaces/test/panther_system/gpio/test_gpio_controller.cpp
pawelirh: I left it as it was (no unit tests) because of the following:
- we're approaching the end of new features for the older version
- the majority of gpio controller methods are not implemented for older version

panther_hardware_interfaces/test/panther_system/gpio/test_gpio_controller.cpp
rafal-gorecki: ok

rafal-gorecki (approved):

PR #430, opened 2024-10-29
Ros2 docking manager no lights

rafal-gorecki:
### Description

- Adds BT for managing the docking
- Add use_docking to simulation.launch.py
- Rename are_buttons_pressed -> check_joy_msg and extend functionalities
- Add right now unused check_bool_msg plugin (to use it we need to modify BT code to be able to set up custom QOS setting)

### Requirements

- [x] Documentation updated

### Tests 🧪

- [ ] Robot
- [ ] Container
- [x] Simulation

## Release Notes

- **New Features**
  - Introduced new launch arguments for docking management (`docking_bt_project_path`, `use_docking`).
  - Added a new `DockingManagerNode` to manage docking operations.
  - Implemented behavior trees for docking and undocking operations with joystick input handling.

- **Documentation**
  - Enhanced README and CONFIGURATION files for clearer setup instructions and added details on new features.

- **Bug Fixes**
  - Improved error handling and validation in various nodes to ensure more robust operation.

- **Tests**
  - Added unit tests for new `CheckBoolMsg` and `CheckJoyMsg` plugins to ensure functionality and reliability.


delihus:
@coderabbitai review this, please.

README.md
```diff
@@ -102,8 +102,10 @@ Launch arguments are largely common to both simulation and physical robot. Howev
...
 | 🤖🖥️  | `publish_robot_state`        | Whether to publish the default Panther robot description. <br/> ***bool:*** `True`                                                                                                                                                                                                                                                |
 | 🖥️   | `robots`                     | The list of the robots spawned in the simulation e.g. `robots:='robot1={x: 1.0, y: -2.0}; robot2={x: 1.0, y: -4.0}'` <br/> ***string:*** `''`                                                                                                                                                                                                                  |
+| 🤖🖥️  | `docking_bt_project_path` | Path to BehaviorTree project file, responsible for robot states management. <br/> ***string:*** [`DockingBT.btproj`](./panther_manager/behavior_trees/DockingBT.btproj)                                                                                                                                                                   |
```
delihus:
```suggestion
| 🤖🖥️  | `docking_bt_project_path` | Path to BehaviorTree project file, responsible for docking management. <br/> ***string:*** [`DockingBT.btproj`](./panther_manager/behavior_trees/DockingBT.btproj)                                                                                                                                                                   |
```

panther_docking/launch/docking.launch.py
```diff
@@ -79,28 +90,15 @@ def generate_launch_description():
                 "use_sim_time": use_sim,
             },
         ],
-        namespace=namespace,
-    )
-
-    station_launch = IncludeLaunchDescription(
```
delihus, line 85: Why is it removed?

panther_docking/launch/docking.launch.py
```diff
@@ -58,18 +66,21 @@ def generate_launch_description():
...
         executable="opennav_docking",
+        namespace=namespace,
         parameters=[
             namespaced_docking_server_config,
             {"use_sim_time": use_sim},
         ],
-        namespace=namespace,
+        remappings=[("~/transition_event", "~/_transition_event")],
         emulate_tty=True,
+        condition=IfCondition(use_docking),
```
delihus, line 76: This is docking launch so it is obvious that it should be run. 

We should move this condition where we include `docking.launch.py` file.

panther_gazebo/launch/simulate_robot.launch.py
```diff
@@ -185,6 +185,19 @@ def generate_launch_description():
...
+        PythonLaunchDescriptionSource(
+            PathJoinSubstitution(
+                [
+                    FindPackageShare("panther_docking"),
+                    "launch",
+                    "docking.launch.py",
+                ]
+            ),
+        ),
+        launch_arguments={"namespace": namespace, "use_sim": "True"}.items(),
```
delihus, line 198: Use docking condition should be used here
```suggestion
        condition=IfCondition(use_docking),
        launch_arguments={"namespace": namespace, "use_sim": "True"}.items(),
```

panther_manager/CMakeLists.txt
```diff
@@ -73,16 +80,19 @@ foreach(bt_plugin ${plugin_libs})
 endforeach()
 
 add_executable(
-  safety_manager_node src/safety_manager_node_main.cpp
-                      src/safety_manager_node.cpp src/behavior_tree_manager.cpp)
+  safety_manager_node
+  src/safety_manager_node_main.cpp src/safety_manager_node.cpp
+  src/behavior_tree_manager.cpp src/behavior_tree_manager.cpp)
```
delihus: Is it double?
```suggestion
src/behavior_tree_manager.cpp)
```

panther_manager/CMakeLists.txt
```diff
@@ -73,16 +80,19 @@ foreach(bt_plugin ${plugin_libs})
...
 ament_target_dependencies(
```
delihus: We can depend on all of the packages `PACKAGE_DEPENDENCIES`

panther_manager/CMakeLists.txt
```diff
@@ -91,16 +101,34 @@ add_executable(
...
```
delihus: Same here  PACKAGE_DEPENDENCIES

panther_manager/CMakeLists.txt
```diff
@@ -91,16 +101,34 @@ add_executable(
...
+ament_target_dependencies(
+  docking_manager_node
+  behaviortree_ros2
+  geometry_msgs
+  panther_msgs
+  panther_utils
+  rclcpp
+  sensor_msgs
+  std_msgs
+  tf2_geometry_msgs)
```
delihus:
```suggestion
ament_target_dependencies(
  docking_manager_node
 ${PACKAGE_DEPENDENCIES})
```

panther_manager/include/panther_manager/behavior_tree_utils.hpp
```diff
@@ -93,28 +96,76 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
 namespace BT
 {
 /**
- * @brief Converts a string to a vector of integers.
+ * @brief Converts a string to a vector of float.
```
delihus, line 99: Instead changing vector of int to vector of floats please add another template definition for `float` and leave `int`.

panther_manager/include/panther_manager/behavior_tree_utils.hpp
```diff
@@ -93,28 +96,76 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
...
+  for (const StringView & part : parts) {
+    output.push_back(convertFromString<float>(part));
   }
+  return output;
 }
+
+/**
+ * @brief Converts a string to a PoseStamped message.
+ *
+ * The string format should be "roll,pitch,yaw,x,y,z,frame_id" where:
```
delihus: In the definition of function a sting is divided with `;` not `,`. Please unify this.

panther_manager/include/panther_manager/behavior_tree_utils.hpp
```diff
@@ -93,28 +96,76 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
...
+ *  - roll, pitch, yaw: Euler angles in radians.
```
delihus: Are you sure that there are Euler angles? In the definition I see set `RPY`.

http://docs.ros.org/en/jade/api/tf2/html/classtf2_1_1Quaternion.html#a2ffba08e3fb4c10f9d5cbd6fc43399de

So the axes are X, Y, Z.

panther_manager/include/panther_manager/behavior_tree_utils.hpp
```diff
@@ -93,28 +96,76 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
...
+
+  if (parts.size() != 7) {
+    throw BT::RuntimeError(
+      "Invalid input for PoseStamped. Expected 7 values: x;y;z;roll;pitch;yaw;frame_id");
+  }
+
+  geometry_msgs::msg::PoseStamped pose_stamped;
+
+  try {
+    // Position (x, y, z)
```
delihus:
```suggestion
```

panther_manager/include/panther_manager/behavior_tree_utils.hpp
```diff
@@ -93,28 +96,76 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
...
+    pose_stamped.pose.position.x = convertFromString<double>(parts[0]);
+    pose_stamped.pose.position.y = convertFromString<double>(parts[1]);
+    pose_stamped.pose.position.z = convertFromString<double>(parts[2]);
+
+    // Orientation (R,P,Y -> Quaternion)
```
delihus:
```suggestion
```

panther_manager/include/panther_manager/behavior_tree_utils.hpp
```diff
@@ -93,28 +96,76 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
...
+    double roll = convertFromString<double>(parts[3]);
```
delihus: If the `convertFromString` accepts this you can change `double` to `const auto` or `const double`.

panther_manager/include/panther_manager/behavior_tree_utils.hpp
```diff
@@ -93,28 +96,76 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
...
+    double pitch = convertFromString<double>(parts[4]);
+    double yaw = convertFromString<double>(parts[5]);
+    tf2::Quaternion quaternion;
+    quaternion.setRPY(roll, pitch, yaw);
+    pose_stamped.pose.orientation = tf2::toMsg(quaternion);
+
+    // Frame ID and current time
```
delihus:
```suggestion
```

panther_manager/include/panther_manager/behavior_tree_utils.hpp
```diff
@@ -93,28 +96,76 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
...
+ *
+ * @throw BT::RuntimeError Throws if the input is invalid or cannot be parsed.
+ */
+template <>
+inline geometry_msgs::msg::PoseStamped convertFromString<geometry_msgs::msg::PoseStamped>(
+  StringView str)
+{
+  auto parts = splitString(str, ';');
+
+  if (parts.size() != 7) {
```
delihus: Let's change `7` to some `constexpr` value. Don't use magic numbers.

panther_manager/include/panther_manager/docking_manager_node.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <string>
+
+#include "behaviortree_cpp/bt_factory.h"
+#include "rclcpp/rclcpp.hpp"
+
+#include "std_msgs/msg/bool.hpp"
+
+#include "panther_utils/moving_average.hpp"
+
+#include <panther_manager/behavior_tree_manager.hpp>
```
delihus: According to coding guidelines it should be:
```suggestion
#include <behaviortree_cpp/bt_factory.h>
#include <rclcpp/rclcpp.hpp>

#include <std_msgs/msg/bool.hpp>

#include "panther_utils/moving_average.hpp"
#include "panther_manager/behavior_tree_manager.hpp"
```

panther_manager/include/panther_manager/plugins/action/shutdown_single_host_node.hpp
```diff
@@ -38,14 +38,17 @@ class ShutdownSingleHost : public ShutdownHosts
...
-      BT::InputPort<std::string>("ip", "ip of the host to shutdown"),
-      BT::InputPort<std::string>("username", "user to log into while executing shutdown command"),
-      BT::InputPort<unsigned>("port", "SSH communication port"),
-      BT::InputPort<std::string>("command", "command to execute on shutdown"),
-      BT::InputPort<float>("timeout", "time in seconds to wait for host to shutdown"),
+      BT::InputPort<std::string>("ip", "IP address of the host to shut down."),
+      BT::InputPort<std::string>(
+        "username", "Username to use for logging in and executing the shutdown command."),
+      BT::InputPort<unsigned>("port", "SSH port used for communication (default is usually 22)."),
+      BT::InputPort<std::string>("command", "Shutdown command to execute on the remote host."),
```
delihus:
```suggestion
      BT::InputPort<std::string>("command", "A command to execute on the remote host."),
```

It can be any command.

panther_manager/include/panther_manager/plugins/action/signal_shutdown_node.hpp
```diff
@@ -35,7 +35,7 @@ class SignalShutdown : public BT::SyncActionNode
   static BT::PortsList providedPorts()
   {
     return {
-      BT::InputPort<std::string>("reason", "", "reason to shutdown robot"),
+      BT::InputPort<std::string>("reason", "", "Reason to shutdown robot."),
```
delihus:
```suggestion
      BT::InputPort<std::string>("reason", "", "A reason to shut down a robot."),
```

panther_manager/include/panther_manager/plugins/action/undock_robot_node.hpp
```diff
@@ -49,16 +49,17 @@ class UndockRobot : public BT::RosActionNode<opennav_docking_msgs::action::Undoc
...
   {
-    return providedBasicPorts({
-      BT::InputPort<std::string>(
-        "dock_type", "The dock plugin type, if not previous instance used for docking"),
-      BT::InputPort<float>(
-        "max_undocking_time", 30.0, "Maximum time to get back to the staging pose"),
+    return providedBasicPorts(
+      {BT::InputPort<std::string>(
+         "dock_type",
+         "Specifies the dock plugin type to use for undocking. If empty, the previously used dock "
```
delihus: What does `the previously used` mean? Where can we define a previous dock type?

panther_manager/include/panther_manager/plugins/condition/check_bool_msg.hpp
```diff
@@ -22,34 +22,34 @@
...
-#include <sensor_msgs/msg/joy.hpp>
+#include <std_msgs/msg/bool.hpp>
 
 #include "panther_manager/behavior_tree_utils.hpp"
 
 namespace panther_manager
 {
 
-class AreButtonsPressed : public BT::RosTopicSubNode<sensor_msgs::msg::Joy>
+// FIXME: There is no possibility to set QoS profile. Add it in the future to subscribe e_stop.
```
delihus, line 32: @coderabbitai Could you create an issue with that?

panther_manager/include/panther_manager/plugins/condition/check_joy_msg.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+  BT::NodeStatus onTick(const JoyMsg::SharedPtr & last_msg);
+
+  static BT::PortsList providedPorts()
+  {
+    return providedBasicPorts(
+      {BT::InputPort<std::vector<float>>(
+         "axes", "",
+         "Specifies the expected state of the axes field. An empty string (\"\") means the value "
+         "is ignored."),
```
delihus:
```suggestion
         "Specifies the expected state of the axes field. An empty string (\"\") means values "
         "are ignored."),
```

panther_manager/include/panther_manager/plugins/condition/check_joy_msg.hpp
```diff
@@ -0,0 +1,68 @@
...
+       BT::InputPort<std::vector<int>>(
+         "buttons", "",
+         "Specifies the expected state of the buttons field. An empty string (\"\") means the "
+         "value is ignored."),
```
delihus:
```suggestion
         "Specifies the expected state of the buttons field. An empty string (\"\") means  "
         "values are ignored."),
```

panther_manager/include/panther_manager/plugins/decorator/tick_after_timeout_node.hpp
```diff
@@ -31,7 +31,8 @@ class TickAfterTimeout : public BT::DecoratorNode
 
   static BT::PortsList providedPorts()
   {
-    return {BT::InputPort<float>("timeout", "time in s to wait before ticking child again")};
+    return {BT::InputPort<float>(
+      "timeout", "Time in seconds to wait before ticking the child node again")};
```
delihus:
```suggestion
      "timeout", "Time in seconds to wait before ticking the child node again.")};
```

panther_manager/launch/manager.launch.py
```diff
@@ -32,6 +32,15 @@ def generate_launch_description():
     panther_version = EnvironmentVariable(name="PANTHER_ROBOT_VERSION", default_value="1.0")
     panther_manager_dir = FindPackageShare("panther_manager")
 
+    docking_bt_project_path = LaunchConfiguration("docking_bt_project_path")
+    declare_docking_bt_project_path_arg = DeclareLaunchArgument(
+        "docking_bt_project_path",
+        default_value=PathJoinSubstitution(
+            [panther_manager_dir, "behavior_trees", "DockingBT.btproj"]
+        ),
+        description="Path to BehaviorTree project file, responsible for robot states management.",
```
delihus:
```suggestion
        description="Path to BehaviorTree project file, responsible for docking management.",
```

panther_manager/src/plugins/condition/check_bool_msg.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+namespace panther_manager
+{
+
+BT::NodeStatus CheckBoolMsg::onTick(const BoolMsg::SharedPtr & last_msg)
+{
+  bool expected_data;
+  getInput<bool>("data", expected_data);
+
+  return (last_msg && expected_data == last_msg->data) ? BT::NodeStatus::SUCCESS
```
delihus: Maybe instead of the one big line if statement we can split that and add some logs?

panther_manager/src/plugins/condition/check_joy_msg.cpp
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  if (last_msg->axes.size() != expected_axes.size()) {
+    RCLCPP_WARN_STREAM(
+      this->logger(), GetLoggerPrefix(name())
+                        << "Joy message has " << last_msg->axes.size()
+                        << " axes, expected at least " << expected_axes.size());
+    return false;
+  }
+
```
delihus, line 45: Here should be checking a timeout.

panther_manager/src/plugins/condition/check_joy_msg.cpp
```diff
@@ -0,0 +1,81 @@
...
+  return std::equal(expected_axes.begin(), expected_axes.end(), last_msg->axes.begin());
```
delihus: I don't know if checking axes is useful but on the other hand it does not look good when we check floats with the equality.

Please use somewhere `std::numeric_limits<double>::epsilon() `

panther_manager/src/plugins/condition/check_joy_msg.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+bool CheckJoyMsg::checkButtons(const JoyMsg::SharedPtr & last_msg)
+{
+  std::vector<int> expected_buttons;
+  getInput<std::vector<int>>("buttons", expected_buttons);
```
delihus, line 54: How does it work where there is no `convertFromString<std::vector<int>>`?

panther_manager/src/plugins/condition/check_joy_msg.cpp
```diff
@@ -0,0 +1,81 @@
...
+  return std::equal(expected_buttons.begin(), expected_buttons.end(), last_msg->buttons.begin());
+}
+
+bool CheckJoyMsg::checkTimeout(const JoyMsg::SharedPtr & last_msg)
+{
+  double max_timeout;
+  getInput<double>("timeout", max_timeout);
+
+  return (max_timeout <= 0.0) ||
+         (rclcpp::Clock().now().seconds() < last_msg->header.stamp.sec + max_timeout);
```
delihus: The base class has `node_`. I think we should use this.
```suggestion
         (this->node_->now().seconds() < last_msg->header.stamp.sec + max_timeout);
```

panther_manager/test/plugins/condition/test_check_bool_msg.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+struct TestCase
+{
+  BT::NodeStatus result;
+  bt_ports input;
+  BoolMsg msg;
+};
+
+constexpr auto TOPIC = "bool";
+constexpr auto PLUGIN = "CheckBoolMsg";
```
delihus, line 42:
```suggestion
constexpr auto PLUGIN = "CheckBoolMsg";

```

panther_manager/test/plugins/condition/test_check_bool_msg.cpp
```diff
@@ -0,0 +1,107 @@
...
+{
+  BoolMsg msg;
+  msg.data = data;
+  return msg;
+}
+
+TEST_F(TestCheckBoolMsg, NoTopicSet)
+{
+  bt_ports input = {{"topic_name", ""}, {"data", "true"}};
+  ASSERT_THROW(CreateTree(PLUGIN, input), std::logic_error);
```
delihus: Why `logic_error`?

panther_manager/test/plugins/condition/test_check_joy_msg.cpp
```diff
@@ -0,0 +1,195 @@
...
+     {{"topic_name", TOPIC}, {"axes", "1"}, {"buttons", ""}, {"timeout", "1.0"}},
+     CreateMsg({1})},
+    {BT::NodeStatus::SUCCESS,
+     {{"topic_name", TOPIC}, {"axes", ""}, {"buttons", "1"}, {"timeout", "1.0"}},
+     CreateMsg({}, {1})},
+    {BT::NodeStatus::SUCCESS,
+     {{"topic_name", TOPIC}, {"axes", "0;0"}, {"buttons", "0;0"}, {"timeout", "1.0"}},
+     CreateMsg({0, 0}, {0, 0})},
+    {BT::NodeStatus::SUCCESS,
+     {{"topic_name", TOPIC}, {"axes", "34;0;1"}, {"buttons", "0;0;10;0"}, {"timeout", "1.0"}},
```
delihus: The `axes` should be check with float values from -1.0 to 1.0:
{"axes", "0.1;-0.2;1.0"}

panther_manager/src/docking_manager_node.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include <vector>
+
+#include "ament_index_cpp/get_package_share_directory.hpp"
+#include "behaviortree_ros2/ros_node_params.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+#include "panther_utils/moving_average.hpp"
+
+#include <panther_manager/behavior_tree_manager.hpp>
+#include <panther_manager/behavior_tree_utils.hpp>
```
delihus: According to code guidelines.
```suggestion
#include <ament_index_cpp/get_package_share_directory.hpp>
#include <behaviortree_ros2/ros_node_params.hpp>
#include <rclcpp/rclcpp.hpp>

#include <panther_manager/behavior_tree_manager.hpp>
#include <panther_manager/behavior_tree_utils.hpp>
#include <panther_utils/moving_average.hpp>
```

panther_manager/src/docking_manager_node.cpp
```diff
@@ -0,0 +1,125 @@
...
+{
+
+DockingManagerNode::DockingManagerNode(
+  const std::string & node_name, const rclcpp::NodeOptions & options)
+: Node(node_name, options)
+{
+  RCLCPP_INFO(this->get_logger(), "Constructing node.");
+
+  DeclareParameters();
+  const std::map<std::string, std::any> docking_initial_bb = {};
```
delihus: Why dose something empty is `const`? 

Can the constructor look like this?
```cpp
docking_tree_manager_ = std::make_unique<BehaviorTreeManager>(
    "Docking", {}, bt_server_port);
    ```

File: panther_manager/src/docking_manager_node.cpp [hunk 36]
```diff
@@ -0,0 +1,125 @@
...
+void DockingManagerNode::Initialize()
+{
+  RCLCPP_INFO(this->get_logger(), "Initializing.");
+
+  RegisterBehaviorTree();
+  docking_tree_manager_->Initialize(factory_);
+
+  using namespace std::placeholders;
+
+  const float timer_freq = this->get_parameter("timer_frequency").as_double();
```
delihus: We get double and then convert to float.
```suggestion
  const double timer_freq = this->get_parameter("timer_frequency").as_double();
```

or

```suggestion
  const auto timer_freq = this->get_parameter("timer_frequency").as_double();
```

panther_manager/src/docking_manager_node.cpp
```diff
@@ -0,0 +1,125 @@
...
+  const auto timer_period_ms =
+    std::chrono::milliseconds(static_cast<unsigned>(1.0f / timer_freq * 1000));
+
+  docking_tree_timer_ = this->create_wall_timer(
+    timer_period_ms, std::bind(&DockingManagerNode::TimerCB, this));
```
delihus: You can skip converting to ms.
```suggestion
  const auto timer_period = std::chrono::duration<double>(1.0 / timer_freq);

  docking_tree_timer_ = this->create_wall_timer(
    timer_period, std::bind(&DockingManagerNode::TimerCB, this));
```

panther_manager/src/docking_manager_node.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+  RCLCPP_INFO(this->get_logger(), "Initialized successfully.");
+}
+
+void DockingManagerNode::DeclareParameters()
+{
+  const auto panther_manager_pkg_path =
+    ament_index_cpp::get_package_share_directory("panther_manager");
+  const std::string default_bt_project_path = panther_manager_pkg_path +
```
delihus: It also can be `auto`

panther_manager/src/docking_manager_node.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+  const auto service_availability_timeout =
+    this->get_parameter("ros_communication_timeout.availability").as_double();
+  const auto service_response_timeout =
+    this->get_parameter("ros_communication_timeout.response").as_double();
+
+  BT::RosNodeParams params;
+  params.nh = this->shared_from_this();
+  params.wait_for_server_timeout =
+    std::chrono::milliseconds(static_cast<int>(service_availability_timeout * 1000));
```
delihus: Can we try like this?
```suggestion
  params.wait_for_server_timeout = std::chrono::duration<double>(service_availability_timeout);
```

panther_manager/src/docking_manager_node.cpp
```diff
@@ -0,0 +1,125 @@
...
+  params.server_timeout =
+    std::chrono::milliseconds(static_cast<int>(service_response_timeout * 1000));
```
delihus:
```suggestion
    std::chrono::duration<double>(service_response_timeout);
```

panther_manager/src/docking_manager_node.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+  behavior_tree_utils::RegisterBehaviorTree(
+    factory_, bt_project_path, plugin_libs, params, ros_plugin_libs);
+
+  RCLCPP_INFO(
+    this->get_logger(), "BehaviorTree registered from path '%s'", bt_project_path.c_str());
```
delihus:
```suggestion
  RCLCPP_INFO_STREAM(
    this->get_logger(), "BehaviorTree registered from path '" << bt_project_path << "'");
```

delihus (changes requested):
Thank you! 

Small things and a couple of questions.

panther_manager/src/plugins/condition/check_joy_msg.cpp
rafal-gorecki: There conversion is already implemanted by default in BT.
https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/include/behaviortree_cpp/basic_types.h#L180

panther_manager/src/plugins/condition/check_joy_msg.cpp
rafal-gorecki: Timeout is checking inside `onTick` function. This is only helper function, which check if Axes are the same as provided.

panther_manager/src/plugins/condition/check_joy_msg.cpp
rafal-gorecki: One sensible use is to check if, for example, the speed value is 0. In some cases this can be helpful. However, in most cases this field will most likely be empty.

panther_manager/test/plugins/condition/test_check_bool_msg.cpp
rafal-gorecki: This is what this function throw. I found and use `ASSERT_ANY_THROW` function.

panther_manager/src/plugins/condition/check_bool_msg.cpp
rafal-gorecki: I like it as it is. You could add bool is_data_equal = last_msg->data == expected_data, but this is a simple comparison and I need to check if last_msg is not None. I don't know if it makes sense. As for logs, the only sensible log-level is debug (we don't want to spam that the result of the check is incorrect, because maybe someone doesn't want to start docking now). At the same time, if necessary, the result can be easily viewed in Groot.

panther_manager/src/docking_manager_node.cpp
rafal-gorecki: Almost all of the suggestions for this file also applied to the light_manger and safety manager I based them on. They were fixed there as well.

panther_manager/include/panther_manager/plugins/action/undock_robot_node.hpp
rafal-gorecki: I meant that the action would guess if the robot was previously docked, but I could be wrong.

panther_manager/include/panther_manager/plugins/action/shutdown_single_host_node.hpp
rafal-gorecki: Yes, but if you don't type shutdown command, behaviour of this tree will not be reasonable.

panther_gazebo/launch/simulate_robot.launch.py
rafal-gorecki: This may seem strange, but we already have it done in other packages as well (on latest `ros2-lynx-devel`). Since the arguments are passed to the parent launch file, there is no need to specify this parameter in `bringup.launch.py` ​​and `simulation.launch.py`, which improves the readability of these two most important files. Also this way reduce redundancy of declarated argument.

We can discuss about this, but not in this task.

panther_docking/launch/docking.launch.py
rafal-gorecki: Answered in simulation_robot.launch.py

panther_manager/include/panther_manager/behavior_tree_utils.hpp
rafal-gorecki: It is not neccesary, because is already implemented in BT.CPP

panther_manager/include/panther_manager/behavior_tree_utils.hpp
rafal-gorecki: There seems to be anything wrong. roll, pitch, yaw are Euler angle. Please explain further.

panther_manager/src/docking_manager_node.cpp
rafal-gorecki: Expected type is `std::chrono::milliseconds`.

panther_manager/src/docking_manager_node.cpp
rafal-gorecki: No it return invalid_type.

panther_docking/launch/docking.launch.py
rafal-gorecki: This do not work with simulation. The reason seams to be `TagGenerator2` and I'm not sure how to solve it.

README.md
delihus: I believe it should not be here.

panther_manager/include/panther_manager/behavior_tree_utils.hpp
delihus: It depends on the implementation of a order of axis rotation.


The best description will be from `tf2::Quaternion::setRPY()` what you used bellow. To clarify the description please use exact axis names.

```suggestion
 *  - roll, pitch, yaw: Rotation around axes XYZ.
```


 All of the angle are called roll, pitch, roll in functions:
 - [setRPY](https://docs.ros.org/en/humble/p/tf2/generated/classtf2_1_1Quaternion.html#_CPPv4N3tf210Quaternion6setRPYERK9tf2ScalarRK9tf2ScalarRK9tf2Scalar)
 - [setEuler](https://docs.ros.org/en/humble/p/tf2/generated/classtf2_1_1Quaternion.html#_CPPv4N3tf210Quaternion8setEulerERK9tf2ScalarRK9tf2ScalarRK9tf2Scalar)

panther_manager/include/panther_manager/behavior_tree_utils.hpp
delihus: What about these comments?

panther_manager/include/panther_manager/behavior_tree_utils.hpp
delihus: What about these comments?

panther_manager/include/panther_manager/behavior_tree_utils.hpp
delihus: What about these comments?

panther_manager/include/panther_manager/behavior_tree_utils.hpp
```diff
@@ -93,28 +96,78 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
...
+ * @param str The string to convert.
+ * @return geometry_msgs::msg::PoseStamped The converted PoseStamped message.
+ *
+ * @throw BT::RuntimeError Throws if the input is invalid or cannot be parsed.
+ */
+template <>
+inline geometry_msgs::msg::PoseStamped convertFromString<geometry_msgs::msg::PoseStamped>(
+  StringView str)
+{
+  constexpr size_t expected_parts_size = 7;
```
delihus: I would add `std`
```suggestion
  constexpr std::size_t expected_parts_size = 7;
```

panther_manager/src/docking_manager_node.cpp
delihus: Ah. Okay so maybe just put constructor there but it looks not so good.
```cpp
docking_tree_manager_ = std::make_unique<BehaviorTreeManager>( "Docking",  std::map<std::string, std::any>({}), bt_server_port);
```

Okay so maybe just let's change the name of this variable to `empty_blackboard` or `empty_bb`.

panther_manager/src/docking_manager_node.cpp
delihus: okay...  Then we can cast it. 
```suggestion
 auto timeout_s = std::chrono::duration<double>(service_availability_timeout);
  params.wait_for_server_timeout = std::chrono::duration_cast<std::chrono::milliseconds>(timeout_s);
```

panther_manager/src/docking_manager_node.cpp
delihus: same as above. let's cast  the duration to milliseconds.

panther_manager/src/plugins/condition/check_joy_msg.cpp
delihus: right!

panther_manager/src/plugins/condition/check_joy_msg.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "panther_manager/plugins/condition/check_joy_msg.hpp"
+
+#include "panther_manager/behavior_tree_utils.hpp"
+
+namespace panther_manager
+{
+
+BT::NodeStatus CheckJoyMsg::onTick(const JoyMsg::SharedPtr & last_msg)
+{
+  return (last_msg && checkAxes(last_msg) && checkButtons(last_msg) && checkTimeout(last_msg))
```
delihus: I would change the order.
I would place the `checkTimeout` first because it looks like "static" comparison.   The `checkAxes` and the `checkButtons` are depend on dynamic vectors.This change can skip checking the vectors when the message reaches timeout.

panther_manager/src/plugins/condition/check_joy_msg.cpp
delihus: Wait... :D
there is also double
https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/include/behaviortree_cpp/basic_types.h#L183

Why do we implement it on ourselves?

panther_manager/src/plugins/condition/check_joy_msg.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+bool CheckJoyMsg::checkTimeout(const JoyMsg::SharedPtr & last_msg)
+{
+  double max_timeout;
+  getInput<double>("timeout", max_timeout);
+
+  return (max_timeout <= 0.0) || (this->node_.lock() && this->node_.lock()->now().seconds() <
+                                                          last_msg->header.stamp.sec + max_timeout);
+}
```
delihus: I see now you are not including nano seconds to checking a timeout.

Let's do it using `rclcpp:Time` and let's limit time and amount of locking a node.

```suggestion
bool CheckJoyMsg::checkTimeout(const JoyMsg::SharedPtr & last_msg)
{
  double max_timeout;
  getInput<double>("timeout", max_timeout);
  
    if(max_timeout <= 0.0){
      return false;
    }

    rclcpp::Time now;
    {
      auto node = node_.lock();
      if (!node){
        return false;
      }
      now = node->now();
    }
    

    const auto duration = now - last_msg->header.stamp;
    return duration < rclcpp::Duration::from_seconds(now);
}
```

panther_manager/test/plugins/condition/test_check_bool_msg.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+  BoolMsg msg;
+  msg.data = data;
+  return msg;
+}
+
+TEST_F(TestCheckBoolMsg, NoTopicSet)
+{
+  bt_ports input = {{"topic_name", ""}, {"data", "true"}};
+  ASSERT_ANY_THROW(CreateTree(PLUGIN, input));
```
delihus: I'm not so sure if  we should test this. This exception is on `behaviortree_ros2` level.

I opt to get out of this test. If you want to leave it then please don't use `ANY`. The previous version was okay with catching `std::logic_error`

panther_manager/test/plugins/condition/test_check_joy_msg.cpp
```diff
@@ -0,0 +1,196 @@
...
+void TestCheckJoyMsg::SetCurrentMsgTime(JoyMsg & msg)
+{
+  msg.header.stamp.sec = bt_node_->now().seconds();
+  msg.header.stamp.nanosec = bt_node_->now().nanoseconds();
+}
+
+TEST_F(TestCheckJoyMsg, NoTopicSet)
+{
+  bt_ports input = {{"topic_name", ""}, {"axes", "0;0"}, {"buttons", "0;0"}, {"timeout", "1.0"}};
+  ASSERT_ANY_THROW(CreateTree(PLUGIN, input));
```
delihus: Same  here. As above

delihus:
This test doesn't pass 
```bash
- panther_manager.TestConvertFromString InvalidVectorIntWithFloat
  <<< failure message
    /home/husarion/ros2_ws/src/panther_ros/panther_manager/test/test_behavior_tree_utils.cpp:144
    Expected: BT::convertFromString<std::vector<int>>(str) throws an exception of type BT::RuntimeError.
      Actual: it throws nothing.
  >>>
  ```

**Review by delihus — CHANGES_REQUESTED** [review]

File: panther_manager/include/panther_manager/behavior_tree_utils.hpp (hunk 11 above)
Comment (rafal-gorecki) [reply panther_manager/include/panther_manager/behavior_tree_utils.hpp]: ok

File: panther_manager/include/panther_manager/plugins/action/undock_robot_node.hpp (hunk 20 above)
Comment (delihus) [reply panther_manager/include/panther_manager/plugins/action/undock_robot_node.hpp]: up, you are right https://github.com/open-navigation/opennav_docking/blob/main/opennav_docking/src/docking_server.cpp#L573

File: panther_manager/src/plugins/condition/check_bool_msg.cpp (hunk 26 above)
Comment (delihus) [reply panther_manager/src/plugins/condition/check_bool_msg.cpp]: Okay

File: panther_manager/include/panther_manager/behavior_tree_utils.hpp [hunk 47]
```diff
@@ -93,28 +96,75 @@ inline std::string GetLoggerPrefix(const std::string & bt_node_name)
...
  * @param str The string to convert.
- * @return std::vector<int> The vector of integers.
+ * @return std::vector<float> The vector of float.
  *
- * @throw BT::RuntimeError Throws when there is no input or cannot parse int.
+ * @throw BT::RuntimeError Throws when there is no input or cannot parse float.
  */
 template <>
-inline std::vector<int> convertFromString<std::vector<int>>(StringView str)
+inline std::vector<float> convertFromString<std::vector<float>>(StringView str)
```
delihus, line 107: Actually do we event used that?

Moreover why have we implemented int before?
https://github.com/BehaviorTree/BehaviorTree.CPP/blob/4.5.x/src/basic_types.cpp#L177 

o.O

rafal-gorecki: Vector float is used in joy check, because axes is define as a float vector.

Probably someone do not check it is already implemented.

delihus (approved):

delihus:
Thank you!

panther_docking/launch/docking.launch.py
delihus: I move this to next version

PR #433, opened 2024-10-31
Ros2 add RViz launch

rafal-gorecki:
### Description

- Add rviz launch
- Hide transition_event topics in gazebo

### Requirements

- [x] Documentation updated

### Tests 🧪

- [ ] Robot
- [ ] Container
- [x] Simulation


panther_description/launch/rviz.launch.py
```diff
@@ -0,0 +1,90 @@
...
+        description="RViz configuration file.",
+    )
+
+    use_rviz = LaunchConfiguration("use_rviz")
+    declare_use_rviz_arg = DeclareLaunchArgument(
+        "use_rviz",
+        default_value="True",
+        description="Run RViz simultaneously.",
+        choices=["True", "true", "False", "false"],
+    )
```
KmakD: Shouldn't this be in `simulation.launch.py`? What's the point of argument that makes the same file do nothing, better make that decision on higher level

panther_description/rviz/husarion_ugv.rviz
KmakD, line 1: Should we enable tf, and odometry in default config?

KmakD (changes requested):

panther_description/launch/rviz.launch.py
rafal-gorecki: This may seem strange, but we already have it done in other packages as well (on latest ros2-lynx-devel). Since the arguments are passed to the parent launch file, there is no need to specify this parameter in bringup.launch.py ​​and simulation.launch.py, which improves the readability of these two most important files.

The launch files are not ideal, but imo this solution is cleaner and reduce redundancy of declarated argument.

panther_description/rviz/husarion_ugv.rviz
rafal-gorecki: At the moment we don't have much to show, so there's not much going on there. We can hide them, but it's worth having the visualizations already configured and just clicking the checkbox to reveal them.

panther_description/launch/rviz.launch.py
KmakD: That's not my point, this argument should be on simulation launch file level. There you decide if you should launch rviz or not, not inside rviz launching file. Why would you ever want to use rviz.launch file with argument that says "don't launch it"?
Picture that, If you go to simulation file you will see that rviz is called without any arguments so you assume that it is going to be launched. You don't have any indication that it should be otherwise. It's good to have arguments that are rviz specific "hidden" in simulation.launch, but this one refers to the way the simulation should be launched.

panther_description/rviz/husarion_ugv.rviz
KmakD: I didn't mean to remove them, but just have them disabled by default

rafal-gorecki: Okay, I'll make sure about that

panther_description/launch/rviz.launch.py
rafal-gorecki: Ok, I made sure about that

KmakD (approved):

PR #453, opened 2024-12-03
Ros2 docking lifecycle node

delihus:
### Description

`DockPosePublisher` listens to `/tf` so when there is no detection (most of the time, only when docking) and searches tf_buffer.

This PR changes `DockPosePublisher` to `LifecycleNode` and activates it when there is a docking attempt and deactivates when robot is docked or docking goes wrong.

### Requirements

- [x] Documentation updated

### Tests 🧪

- [ ] Robot
- [x] Container
- [x] Simulation

### DIFF
--- a/panther_docking/src/panther_charging_dock.cpp
+++ b/panther_docking/src/panther_charging_dock.cpp
@@ -67,17 +67,23 @@ void PantherChargingDock::activate()
     std::bind(&PantherChargingDock::setDockPose, this, std::placeholders::_1));
   staging_pose_pub_ = node->create_publisher<PoseStampedMsg>("docking/staging_pose", 1);
 
+  dock_pose_publisher_change_state_client_ =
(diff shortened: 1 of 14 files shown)


panther_docking/src/panther_charging_dock.cpp
```diff
@@ -211,11 +224,12 @@ bool PantherChargingDock::isCharging()
     }
 
     if (wibotic_info_->i_charger > kWiboticChargingCurrentThreshold) {
+      setDockPosePublisherState(lifecycle_msgs::msg::Transition::TRANSITION_DEACTIVATE);
       return true;
     }
   } catch (const opennav_docking_core::FailedToDetectDock & e) {
     RCLCPP_ERROR_STREAM(logger_, "An occurred error while checking if charging: " << e.what());
```
rafal-gorecki, line 216: This catch only seems to catch one throw. Besides, there are quite a lot of checks in this function. Maybe we could pack some of them into a separate helper function?

rafal-gorecki (approved):

PR #467, opened 2024-12-13
Add docker into repository

rafal-gorecki:
### Description

- add panther-docker functionalities into husarion_ugv repo
- clean up workflows

### Requirements

- [x] Documentation updated

### Tests 🧪

- [x] Robot (build in clean workspace test pass)
- [x] Container
- [x] Simulation


docker/demo/compose.simulation.yaml
```diff
@@ -19,4 +22,6 @@ services:
     volumes:
       - /tmp/.X11-unix:/tmp/.X11-unix:rw
     command: >
-      ros2 launch panther_gazebo simulation.launch.py robot_model:=panther namespace:=panther
+      ros2 launch panther_gazebo simulation.launch.py
+        namespace:=panther
+        robot_model:=panther
```
KmakD: Just a suggestion. Adding env `ROBOT_MODEL=${ROBOT_MODEL: -panther}`, so user don't have to modify compose to choose between robots

.github/workflows/build-docker.yaml
```diff
@@ -32,30 +49,24 @@ on:
           YYYYMMDD (eg. 20220124)
         type: string
         default: '20131206'
-      feature_branch_tag_suffix:
-        description: In case of "development" release from a feature branch specify the custom tag suffix
-          for the docker image
-        type: string
-        default: ''
-
```
KmakD, line 40: Why are we removing this?

.github/workflows/release-candidate.yaml
```diff
@@ -15,121 +15,90 @@ env:
...
-      - name: Trigger repository build workflow
-        uses: convictional/trigger-workflow-and-wait@v1.6.1
+      - name: Check docs build
+        uses: convictional/trigger-workflow-and-wait@v1.6.5
         with:
           owner: husarion
           repo: docs_new
-          github_token: ${{ secrets.GH_PAT }}
-          workflow_file_name: test-release.yml
+          github_token: ${{ secrets.RAFAL_ACCESS_TOKEN }}
```
KmakD, line 27: That private token looks weird. Can we change it to sth more general?

.github/workflows/release-candidate.yaml
```diff
@@ -15,121 +15,90 @@ env:
...
+    name: Unit tests
     runs-on: ubuntu-22.04
     steps:
-      - name: Trigger repository build workflow
-        run: echo "Unit tests for panther_ros are not fully implemented yet -> SKIPPING!"
+      - name: Run unit tests
+        run: echo "Unit tests are not fully implemented yet -> SKIPPING!"
 
-  create_release_candidate_branches:
-    name: Create release candidate branches
```
KmakD, line 41: release candidate branches should be created for both `panther_ros` and `panther-rpi-os-img` repos. Only, then the docker tag should be updated (on the new release branch).

.github/workflows/release-candidate.yaml
```diff
@@ -15,121 +15,90 @@ env:
...
-              "branch_name": "${{ env.RC_BRANCH_NAME }}",
-              "image_tag": "humble-${{ env.RC_BRANCH_NAME }}"
-            }
+          message: Update tags in compose files
+          author_name: action-bot
+          author_email: action-bot@action-bot.com
+          new_branch: ${{ env.RC_BRANCH_NAME }}
 
-  build_and_push_docker_images:
-    name: Build panther docker images
```
KmakD, line 79: You removed this step that is required

.github/workflows/release-candidate.yaml
```diff
@@ -15,121 +15,90 @@ env:
...
+      - name: Run unit tests
+        run: echo "Unit tests are not fully implemented yet -> SKIPPING!"
 
-  create_release_candidate_branches:
-    name: Create release candidate branches
+  docker:
+    name: Docker
     runs-on: ubuntu-22.04
-    needs:
-      - unit_test_panther_ros
```
KmakD, line 44: You removed this everywhere. It tells in what order all of this should be executed and that this order matters

.github/workflows/release-candidate.yaml
```diff
@@ -15,121 +15,90 @@ env:
...
             {
               "dev_image": "true",
-              "panther_codebase_version": "${{ env.RC_BRANCH_NAME }}",
+              "husarion_ugv_version": "${{ env.RC_BRANCH_NAME }}",
               "image_tag": "${{ github.event.inputs.version }}"
             }
 
-  build_and_publish_rpi_flash_os_image:
-    name: Build panther flash OS image
-    if: ${{ fromJSON(github.event.inputs.automatic_mode) == true }}
```
KmakD, line 122: Why are we removing this condition?

.github/workflows/release-project.yaml
```diff
@@ -31,64 +31,29 @@ env:
...
-            }
+          release_candidate: ${{ env.RC_BRANCH_NAME }}
+          version: ${{ github.event.inputs.version }}
+          release_name: ${{ github.event.inputs.release_name }}
+          automatic_mode: ${{ github.event.inputs.automatic_mode }}
+          prerelease: ${{ github.event.inputs.prerelease }}
 
-      - name: Rebuild panther docker images with new version
-        uses: convictional/trigger-workflow-and-wait@v1.6.1
+      - name: Build development Docker
```
KmakD:
```suggestion
      - name: Rebuild panther docker images with new version
```
Old name was more descriptive. This is required to run after repository is released and version in package.xml was updated. This is not development image, it is release image. We pass build_type: development so the image tag does not have `-stable` at the end

docker/Dockerfile.hardware
```diff
@@ -2,21 +2,20 @@ ARG ROS_DISTRO=humble
...
 
-ARG BRANCH_NAME=ros2-devel
 ARG BUILD_TEST=OFF
 
 ENV HUSARION_ROS_BUILD_TYPE=hardware
 
 STOPSIGNAL SIGINT
 
 WORKDIR /ros2_ws
+COPY .. /ros2_ws/src/husarion_ugv
```
KmakD:
```suggestion
COPY . src/husarion_ugv
```
I believe this will be run from panther_ros directory so context will be set there. Also no point using full path when we specify WORKDIR

docker/Dockerfile.hardware
```diff
@@ -28,17 +27,17 @@ RUN apt-get update  && \
...
     apt-get remove -y \
         ros-dev-tools && \
     apt-get autoremove -y && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-COPY update_config_directory.sh /usr/local/sbin/update_config_directory
-COPY setup_environment.sh /usr/local/sbin/setup_environment
+COPY utils/update_config_directory.sh /usr/local/sbin/update_config_directory
+COPY utils/setup_environment.sh /usr/local/sbin/setup_environment
```
KmakD:
```suggestion
COPY docker/utils/update_config_directory.sh /usr/local/sbin/update_config_directory
COPY docker/utils/setup_environment.sh /usr/local/sbin/setup_environment
```
Talking about right context

docker/Dockerfile.simulation
```diff
@@ -10,13 +10,13 @@ ENV HUSARION_ROS_BUILD_TYPE=simulation
 STOPSIGNAL SIGINT
 
 WORKDIR /ros2_ws
+COPY .. /ros2_ws/src/husarion_ugv
```
KmakD:
```suggestion
COPY . src/husarion_ugv
```
Same as above

KmakD (changes requested):
Can you please add instructions to README, how to build Docker images locally?

.github/workflows/build-docker.yaml
rafal-gorecki: So as not to complicate and not to mix things up. The action works by default in such a way that if the build is not from the main branch, then the `{distro}-{branch_name}` image is created.

I see no reason to leave the argument, related to another argument, which we also remove `panther_codebase_version`.

If this is not clear I can give a broader example.

.github/workflows/release-candidate.yaml
rafal-gorecki: I need to read more about that. Earlier `GH_PAT` was Pawel Irzyk tolken and for today this token hes expired, so I exchanged it for my own.

.github/workflows/release-candidate.yaml
rafal-gorecki: This step is right now just executing `./.github/workflows/build-docker.yaml`

.github/workflows/release-candidate.yaml
rafal-gorecki: Because there is not such input as `automatic_mode` ;)

docker/Dockerfile.hardware
rafal-gorecki: Thanks for catching that

docker/Dockerfile.simulation
rafal-gorecki: Thanks for catching that!

docker/demo/compose.simulation.yaml
rafal-gorecki: Personally, I stopped being a fan of this solution when I had more than 5 exports to autonomy. Then we add a way of configuring it, which should be listed. As far as I'm concerned, we have a table with arguments for this, so that everyone can look at it and add their own settings if necessary.
I added what you asked for, because in this case it is written in the readme that it is based on env, so it makes sense

rafal-gorecki, line 27:
```suggestion
        namespace:=${ROBOT_NAMESPACE:-panther}
        robot_model:=${ROBOT_MODEL:-panther}
```

docker/demo/compose.rviz.yaml
```diff
@@ -17,12 +17,12 @@ x-gpu-config:
...
+    container_name: rviz
     <<:
       - *common-config
       - *cpu-config
     volumes:
       - /tmp/.X11-unix:/tmp/.X11-unix:rw
-      - ./config/panther.rviz:/root/.rviz2/default.rviz
-    stop_signal: SIGINT
+    command: >
+      ros2 launch husarion_ugv_description rviz.launch.py
```
rafal-gorecki, line 28:
```suggestion
      ros2 launch husarion_ugv_description rviz.launch.py
          namespace:=${ROBOT_NAMESPACE:-panther}
```

.github/workflows/release-project.yaml
rafal-gorecki: Steps are executed one by one, there is no need to create sepret jobs with `need` field for that.

.github/workflows/release-candidate.yaml
rafal-gorecki: Steps are executed one by one. There is no need of creating seprate jobs for that.

docker/demo/compose.simulation.yaml
```diff
@@ -3,12 +3,17 @@ x-common-config: &common-config
   ipc: host
   environment:
     - RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
+    - ROBOT_MODEL=${ROBOT_MODEL:-panther}
```
KmakD:
```suggestion
    - ROBOT_MODEL=${ROBOT_MODEL_NAME:-panther}
```
according to launch file

KmakD (changes requested):

.github/workflows/release-candidate.yaml
KmakD: Ideally we have some "organization" token here

KmakD (approved):
github tokens are one last thing to change. 
I'm giving approve not to block you when I'm away for Christmas.

PR #491, opened 2025-02-05
Add hw and os checks

KmakD:
### Description

Added following checks in launch files:
- Check for incorrect hardware configuration
- Check for incorrect OS version
- Prevent starting driver if hardware configuration is incorrect

image

### Requirements

- [x] Documentation updated

### Tests 🧪

- [ ] Robot
- [ ] Container
- [ ] Simulation

## Summary by CodeRabbit

- **New Features**
  - Enhanced the startup process with robust hardware and operating system compatibility checks.
  - Standardized error and warning displays now alert users when system requirements are not met.
  - New launch configuration options allow users to customize the system’s behavior in response to incompatibility issues.

### DIFF
--- a/husarion_ugv_bringup/launch/bringup.launch.py
+++ b/husarion_ugv_bringup/launch/bringup.launch.py
@@ -15,20 +15,67 @@
 # limitations under the License.
 
 
-from husarion_ugv_utils.messages import welcome_msg
+import os
+import re
+
+from husarion_ugv_utils.messages import (
+    ErrorMessages,
+    error_msg,
+    warning_msg,
+    welcome_msg,
+)
 from launch import LaunchDescription
-from launch.actions import DeclareLaunchArgument, IncludeLaunchDescription, TimerAction
-from launch.conditions import UnlessCondition
+from launch.actions import (
+    DeclareLaunchArgument,
+    ExecuteProcess,
+    GroupAction,
+    IncludeLaunchDescription,
+    TimerAction,
+)
+from launch.conditions import IfCondition, UnlessCondition
 from launch.launch_description_sources import PythonLaunchDescriptionSource
 from launch.substitutions import (
     EnvironmentVariable,
     LaunchConfiguration,
     PathJoinSubstitution,
+    PythonExpression,
 )
 from launch_ros.substitutions import FindPackageShare
 
+MIN_REQUIRED_OS_VERSION = [2, 2, 0]
+
+
+def check_os_version_compatibility(version_string: str, min_required_version: list[int]) -> bool:
+    match = re.search(r"v(\d+\.\d+\.\d+)", version_string)
+
+    if not match:
+        return False
+
+    version_str = match.group(1)
+    version = version_str.split(".")
+
+    if int(version[0]) > min_required_version[0]:
+        return True
+
+    if int(version[0]) == min_required_version[0]:
+        if int(version[1]) > min_required_version[1]:
+            return True
+
+        if int(version[1]) == min_required_version[1]:
+            return int(version[2]) >= min_required_version[2]
+
+    return False
+
 
 def generate_launch_description():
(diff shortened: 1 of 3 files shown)


husarion_ugv_bringup/launch/bringup.launch.py
```diff
@@ -139,6 +186,37 @@ def generate_launch_description():
...
+            error_msg(ErrorMessages.INCORRECT_HW_CONFIG),
+            prevent_exit_action,
+        ],
+        condition=UnlessCondition(hw_config_correct),
+    )
+
+    os_version = os.environ.get("SYSTEM_BUILD_VERSION", "v0.0.0")
+    os_version_correct = PythonExpression(
+        f"{check_os_version_compatibility(os_version, MIN_REQUIRED_OS_VERSION)}"
+    )
```
rafal-gorecki, line 207: PythonExpression probably can be omitted, because we based on `os.environ`

rafal-gorecki: PythonExpression probably can be omitted, because `os.environ` is just a string
```suggestion
    os_version_correct = check_os_version_compatibility(os_version, MIN_REQUIRED_OS_VERSION)
```

husarion_ugv_bringup/launch/bringup.launch.py
```diff
@@ -15,20 +15,67 @@
...
+            return int(version[2]) >= min_required_version[2]
+
+    return False
+
 
 def generate_launch_description():
+    exit_on_wrong_hw = LaunchConfiguration("exit_on_wrong_hw")
+    declare_exit_on_wrong_hw_arg = DeclareLaunchArgument(
+        "exit_on_wrong_hw",
+        default_value="false",
```
rafal-gorecki, line 52: Why not default true?

husarion_ugv_bringup/launch/bringup.launch.py
```diff
@@ -15,20 +15,67 @@
...
+        return True
+
+    if int(version[0]) == min_required_version[0]:
+        if int(version[1]) > min_required_version[1]:
+            return True
+
+        if int(version[1]) == min_required_version[1]:
+            return int(version[2]) >= min_required_version[2]
+
+    return False
```
rafal-gorecki: Propose to move it to `*_utils` and simplify it. Like in following example
```python
import re

MIN_REQUIRED_OS_VERSION = "v2.2.2"


def extract_version_tuple(version_string: str) -> tuple[int, int, int]:
    match = re.search(r"v(\d+)\.(\d+)\.(\d+)", version_string)
    return tuple(map(int, match.groups())) if match else (0, 0, 0)

def check_os_version_compatibility(version1: str, min_required_version: str) -> bool:
    return extract_version_tuple(version1) >= extract_version_tuple(min_required_version)

# Example usage
print(check_os_version_compatibility("v1.9.5", MIN_REQUIRED_OS_VERSION))  # False
print(check_os_version_compatibility("v2.2.2", MIN_REQUIRED_OS_VERSION))  # True
print(check_os_version_compatibility("v2.3.0", MIN_REQUIRED_OS_VERSION))  # True
```

rafal-gorecki (changes requested):

husarion_ugv_bringup/launch/bringup.launch.py
KmakD: UnlessCondition doesn't accept bool, that's why this is wrapped with PythonExpression

husarion_ugv_bringup/launch/bringup.launch.py
KmakD: It's the debate what is our default state. I assumed using docker and avoiding unnecessary restarts of it

husarion_ugv_bringup/launch/bringup.launch.py
KmakD: That is smart

rafal-gorecki (approved):

PR #493, opened 2025-02-20
Ros2 shutdown request

KmakD:
MERGE AFTER RELEASE

### Description

- Add execute command bt node
- use execute command to shutdown localhost
- remove shutdown single host
- use HTTP  call to shutdown remote hosts
- Update tests

### Requirements

- [ ] Documentation updated

### Tests 🧪

- [ ] Robot
- [ ] Container
- [ ] Simulation

## Summary by CodeRabbit

- **New Features**
  - Introduced an HTTP-based shutdown mechanism with updated configuration options (default port 3003 and a security token) for multi-host management.
  - Added a flexible command execution capability to enhance behavior management and system control.
  - Enhanced container configuration to support system power-off via D-Bus.

- **Refactor**
  - Updated the internal security framework by replacing legacy SSH dependencies with a more robust solution.
  - Streamlined shutdown logic to improve reliability and maintainability.

- **Documentation**
  - Revised user guides and configuration instructions to reflect the new shutdown and command execution options.

- **Tests**
  - Expanded and reorganized tests to validate the new shutdown and command execution functionalities.


husarion_ugv_manager/include/husarion_ugv_manager/plugins/shutdown_host.hpp
```diff
@@ -120,152 +142,89 @@ class ShutdownHost
...
-
-    ssh_channel_send_eof(channel_);
-    ssh_channel_close(channel_);
-    ssh_channel_free(channel_);
-    ssh_disconnect(session_);
-    ssh_free(session_);
+    return std::chrono::duration_cast<std::chrono::seconds>(
+             std::chrono::system_clock::now().time_since_epoch())
+      .count();
   }
```
rafal-gorecki, line 168: The function looking like this suggests that it is important that we are running time_since_epoch, but I don't quite understand why it would matter, we can't just use
```suggestion
  std::int64_t GetTimeSinceEpoch()
  {
    return std::chrono::duration_cast<std::chrono::seconds>(
             std::chrono::system_clock::now()).count();
  }
```
And if we can, I suggest adding this function to *_utils.

rafal-gorecki, line 168: The function looking like this suggests that it is important that we are running time_since_epoch, but I don't quite understand why it would matter, we can't just use
```suggestion
  std::int64_t GetTimeSinceEpoch()
  {
    return std::chrono::duration_cast<std::chrono::seconds>(
             std::chrono::system_clock::now()).count();
  }
```
And if we can, I suggest adding this function to *_utils.

rafal-gorecki (commented):
Works well, no major errors

rafal-gorecki (changes requested):
Works well, no major errors

husarion_ugv_manager/include/husarion_ugv_manager/plugins/shutdown_host.hpp
KmakD: system_clock::now() returns a `time_point`, and we need `duration` to use count. Also, shutdown http server expects to pass timestamp of the request in format of "time since epoch in seconds". In that context this method is quite specific for calling http server. Where in utils would you suggest to place this?

husarion_ugv_manager/src/plugins/action/shutdown_single_host_node.cpp
```diff
@@ -1,80 +0,0 @@
...
-    RCLCPP_ERROR_STREAM(*this->logger_, GetLoggerPrefix(name()) << "Failed to get input [timeout]");
-    return false;
-  }
-
-  bool ping_for_success;
-  if (!getInput<bool>("ping_for_success", ping_for_success)) {
-    RCLCPP_ERROR_STREAM(
-      *this->logger_, GetLoggerPrefix(name()) << "Failed to get input [ping_for_success]");
-    return false;
-  }
```
rafal-gorecki, line 67: Some of these parameters are visible in documentation

rafal-gorecki (approved):

PR #494, opened 2025-02-25
ROS 2 lights new review

KmakD:
### Description

- Address my own review comments xD
- Add what seems to be final color correction for led driver
- Update animations so they looks nice with new color correction
- Update tests
- Some other minor changes

### Requirements

- [ ] Documentation updated

### Tests 🧪

- [ ] Robot
- [ ] Container
- [ ] Simulation


husarion_ugv_lights/CONFIGURATION.md
```diff
@@ -2,19 +2,26 @@
...
+|   6   | CRITICAL_BATTERY  |    2     | ![CRITICAL_BATTERY](.docs/CRITICAL_BATTERY.webp)   |
+|   7   | CHARGING_BATTERY  |    2     | ![CHARGING_BATTERY](.docs/ERROR.webp)              |
+|   8   | BATTERY_CHARGED   |    2     | ![CHARGING_BATTERY](.docs/CHARGING_BATTERY.webp)   |
+|   9   | CHARGER_INSERTED  |    1     | ![CHARGER_INSERTED](.docs/ERROR.webp)              |
+|   10  | BATTERY_NOMINAL   |    2     | ![BATTERY_NOMINAL](.docs/ERROR.webp)               |
+|   11  | AUTONOMOUS_READY  |    3     | ![AUTONOMOUS_READY](.docs/ERROR.webp)              |
+|   12  | AUTONOMOUS_ACTION |    3     | ![AUTONOMOUS_ACTION](.docs/AUTONOMOUS_ACTION.webp) |
+|   13  | GOAL_ACHIEVED     |    1     | ![GOAL_ACHIEVED](.docs/GOAL_ACHIEVED.webp)         |
+|   14  | BLINKER_LEFT      |    1     | ![BLINKER_LEFT](.docs/ERROR.webp)                  |
+|   15  | BLINKER_RIGHT     |    1     | ![BLINKER_RIGHT](.docs/ERROR.webp)                 |
```
rafal-gorecki: Most ot this data is right now out of date. Also the link to `ERROR.webp` are incorrect. Suggest comment it out and add comment **comming soon**.

husarion_ugv_lights/include/husarion_ugv_lights/apa102.hpp
```diff
@@ -155,9 +155,10 @@ class APA102 : public APA102Interface
   static constexpr std::uint8_t kBits = 8;
 
   // Color correction constants
-  static constexpr std::uint16_t kCorrRed = 255;
-  static constexpr std::uint16_t kCorrGreen = 200;
-  static constexpr std::uint16_t kCorrBlue = 62;
+  static constexpr std::uint16_t kCorrRed = 245;
+  static constexpr std::uint16_t kCorrGreen = 255;
+  static constexpr std::uint16_t kCorrBlue = 240;
+  static constexpr float kCorrectionGamma = 2.2;
```
rafal-gorecki, line 161: IMO its worth to mention how/why this values are set like that.

husarion_ugv_lights/test/unit/led_components/test_led_segment.cpp
```diff
@@ -218,31 +219,34 @@ TEST_F(TestLEDSegment, UpdateAnimation)
...
 
-int main(int argc, char ** argv)
+TEST_F(TestLEDSegment, MergeFrames)
 {
-  testing::InitGoogleTest(&argc, argv);
-  return RUN_ALL_TESTS();
-}
+  const std::uint8_t c_1 = 50;
+  const std::uint8_t alpha_1 = 100;
+  const std::uint8_t c_2 = 100;
```
rafal-gorecki: What c_1, c_2 represent? color? If yes maybe we can use `pixel` or `rgba` var name instead (or based on frame)?
std::vector<std::uint8_t> pixel1 = {50, 50, 50, 100};

husarion_ugv_lights/test/unit/led_components/test_led_segment.cpp
```diff
@@ -32,7 +32,11 @@ class LEDSegmentWrapper : public husarion_ugv_lights::LEDSegment
   {
   }
 
-  // std::shared_ptr<husarion_ugv_lights::Animation> GetAnimation() const { return animation_; }
+  void MergeFrames(
+    std::vector<std::uint8_t> & frame, const std::vector<std::uint8_t> & input_frame) const
+  {
```
rafal-gorecki: A bit enigmatic names. IMO tells from which frame alpha will be taken.
`(std::vector<std::uint8_t> & base_frame, const std::vector<std::uint8_t> & overlay_frame) const)`

husarion_ugv_lights/src/led_components/led_segment.cpp
```diff
@@ -106,28 +105,36 @@ bool LEDSegment::IsAnimationFinished(AnimationPriority layer) const
...
 
+void LEDSegment::MergeFrames(
+  std::vector<std::uint8_t> & frame, const std::vector<std::uint8_t> & input_frame) const
+{
+  for (std::size_t i = 0; i < num_led_; i++) {
+    for (std::size_t j = 0; j < 3; j++) {
+      frame[i * 4 + j] = (input_frame[i * 4 + j] * input_frame[i * 4 + 3] +
+                          frame[i * 4 + j] * (255 - input_frame[i * 4 + 3])) /
+                         255;  // value * alpha + background_value * (1 - alpha)
+    }
```
rafal-gorecki: Variable `alpha` should be created before second loop.

husarion_ugv_lights/src/led_components/led_segment.cpp
```diff
@@ -106,28 +105,36 @@ bool LEDSegment::IsAnimationFinished(AnimationPriority layer) const
...
+    frame[i * 4 + 3] = input_frame[i * 4 + 3] + frame[i * 4 + 3] * (255 - input_frame[i * 4 + 3]) /
+                                                  255;  // alpha + (1 - alpha) * background_alpha
```
rafal-gorecki: Why is like that?
I wonder about case alpha=0.5 background_alpha=0.5 so the result will be 0.75. Is the result correct?

Also how is display lights if alpha != 1?

rafal-gorecki (changes requested):
I almost check everything. The two new tests I will check later.

husarion_ugv_lights/test/unit/animation/test_moving_image_animation.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+TEST_F(TestMovingImageAnimation, Initialize)
+{
+  YAML::Node animation_description = YAML::Load("{duration: 2.0}");
+
+  // missing image in description
+  EXPECT_THROW(animation_->Initialize(animation_description, 10, 10.0), std::runtime_error);
+
+  animation_description["image"] = this->test_image_path;
+  EXPECT_NO_THROW(animation_->Initialize(animation_description, 10, 10.0));
```
rafal-gorecki, line 68: What if path does not exist?

husarion_ugv_lights/test/unit/animation/test_moving_image_animation.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+namespace gil = boost::gil;
+
+class MovingImageAnimationWrapper : public husarion_ugv_lights::MovingImageAnimation
+{
+public:
+  MovingImageAnimationWrapper() {}
+
+  std::vector<uint8_t> UpdateFrame() { return MovingImageAnimation::UpdateFrame(); }
+  void SetParam(const std::string & param) { MovingImageAnimation::SetParam(param); }
```
rafal-gorecki, line 37: If I understand it correctly SetParam(param) can only change image_position. If yes, this function have to generic name.

husarion_ugv_lights/CONFIGURATION.md
KmakD: I'm working on it

husarion_ugv_lights/src/led_components/led_segment.cpp
KmakD: It doesn't make difference if it is created before or after the loop.

husarion_ugv_lights/src/led_components/led_segment.cpp
KmakD: Yes, it is correct. Imagine putting dimmed glass on top of another one, the more you put the less transparent it becomes. Similar adding 0.5 image on top of another one will decrease transparency of base image by half.

Alpha is used only for merging images it's value does not affect display

husarion_ugv_lights/test/unit/animation/test_moving_image_animation.cpp
KmakD: Since this class inherits from ImageAnimation this is already covered in `test_image_animation.cpp`

husarion_ugv_lights/test/unit/animation/test_moving_image_animation.cpp
KmakD: `SetParam` is a generic function of class `Animation`. IMO it would look weird if we change name of method for wrapper class as it should only wrap the base class and expose protected methods for test purposes.

husarion_ugv_lights/src/led_components/led_segment.cpp
rafal-gorecki: IMO extract alpha variable have 2 advantages.
1. Cleaner: because `alpha` is easier to read than `input_frame[i * 4 + 3]`.
2. Faster: the value is readed opce from vector instead of 3 times for each rgb color.

husarion_ugv_lights/test/unit/animation/test_moving_image_animation.cpp
rafal-gorecki: Nice

husarion_ugv_lights/test/unit/animation/test_moving_image_animation.cpp
rafal-gorecki: Okey I didint notice it based on virtual class

rafal-gorecki (approved):

PR #497, opened 2025-03-04
release fixes

KmakD:
### Description

- hot-fixes for release

### Requirements

- [ ] Documentation updated

### Tests 🧪

- [ ] Robot
- [ ] Container
- [ ] Simulation

### DIFF
--- a/.github/workflows/release-project.yaml
+++ b/.github/workflows/release-project.yaml
@@ -16,10 +16,6 @@ on:
         description: Name of the release to be created. Version in the first place is recommended (e.g.
           `2.0.1-alpha`).
         required: true
-      automatic_mode:
-        type: boolean
-        default: false
-        description: Automatically merge PRs, create releases and build images.
       prerelease:
         type: boolean
         default: false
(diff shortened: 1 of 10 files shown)


.github/workflows/release-project.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
     needs: release_repository
     uses: ./.github/workflows/build-docker.yaml
     with:
+      branch_name: ros2
```
rafal-gorecki: IMO this does do any thing. Accortding to [action](https://github.com/husarion-ci/ros-docker-img-action/blob/main/action.yaml) this parameter only sets up only BRANCH_NAME arg in DOckerfile. Right nowe we do not have such arguments.
1. The simplest solution will be add this branch instead of copying all repository.
2. We can also delete branch at the and of the release process. It should also work correctly.

Also we should build docker with existing in compose tag 2.2.0-20240227 or override it again in release cation.

rafal-gorecki (approved):

rafal-gorecki (changes requested):
Release action may not pass

.github/workflows/release-project.yaml
KmakD: Well yes, but the build-docker action has branch_name as a required parameter: https://github.com/husarion/husarion_ugv_ros/blob/3ebd780330bbfeb327106d06f3363f1dbb370568/.github/workflows/build-docker.yaml#L9

rafal-gorecki (approved):

PR #595, opened 2025-07-11
more worlds in gazebo

husarafal:
### Description

I have added a new `scene` launch argument in `husarion_ugv_gazebo`, in `simulation.launch.py`, which can be set to predefined strings like `outdoor`, `ioffice`, etc. This allows to easily switch between worlds.

Specifying the argument effectively updates the launch arguments passed to descendant launch files, `husarion_gz_worlds/gz_sim.launch.py`, and `husarion_ugv_gazebo/simulate_robot.launch.py`. This is done in order to load a different world and to set a more suitable initial position of the robot.

Try it out like this:

```bash
ros2 launch husarion_ugv_gazebo simulation.launch.py scene:=outdoor
```

Scenes to review:
- [ARG NOT SPECIFIED] - default Husarion world from `husarion_gz_worlds`
- `office` - Husarion office from `husarion_gz_worlds`
- `raceway` - modified [Sonoma Raceway](https://app.gazebosim.org/OpenRobotics/fuel/models/Sonoma%20Raceway) from `husarion_gz_worlds`
- `outdoor` - [Rubicon World from Gazebo Fuel @ Cole](https://app.gazebosim.org/Cole/fuel/worlds/Rubicon%20World)
- `cave` - [Cave World from Gazebo Fuel @ OpenRobotics](https://app.gazebosim.org/OpenRobotics/fuel/worlds/Cave%20World)

### Requirements

- [ ] Backport
- [x] Documentation updated

### Tests 🧪

- [ ] Container
- [x] Simulation

## Summary by CodeRabbit

* **Documentation**
  * Added a "Worlds" section to the README explaining how to run simulations with different world scenes and listing available built-in worlds.

* **Chores**
  * Updated the version of the simulation worlds dependency.

* **Refactor**
  * Changed the default initial robot position along the 'y' axis in the simulation launch settings from -2.0 to 0.0.

### DIFF
--- a/husarion_ugv/simulation_deps.repos
+++ b/husarion_ugv/simulation_deps.repos
@@ -10,7 +10,7 @@ repositories:
   husarion_gz_worlds:
     type: git
     url: https://github.com/husarion/husarion_gz_worlds.git
-    version: 97275568c34e6fe191ca34fa1c1c2b897faf9610
+    version: 3cc8018a49de082f5e4d3fa7afa4e30dda9307c6
   husarion_components_description:
     type: git
     url: https://github.com/husarion/husarion_components_description.git
--- a/husarion_ugv/simulation_deps.repos
+++ b/husarion_ugv/simulation_deps.repos
@@ -10,7 +10,7 @@ repositories:
   husarion_gz_worlds:
     type: git
     url: https://github.com/husarion/husarion_gz_worlds.git
-    version: 3cc8018a49de082f5e4d3fa7afa4e30dda9307c6
+    version: 56b5788a9584d3bc31ba85d36582ba965752c27e
   husarion_components_description:
     type: git
     url: https://github.com/husarion/husarion_components_description.git
--- a/husarion_ugv_gazebo/README.md
+++ b/husarion_ugv_gazebo/README.md
@@ -9,6 +9,24 @@ The package contains a launch file and source files used to run the robot simula
 - `simulate_multiple_robots.launch.py`: Similar to the above with logic allowing you to quickly add a swarm of robots.
 - **`simulation.launch.py`**: A target file that runs the gazebo simulator that adds and simulates the robot's behavior in accordance with the given arguments.
 
+### Worlds
+
+To run the simulation with a different world, use:
+
+```bash
+ros2 launch husarion_ugv_gazebo simulation.launch.py gz_world:=<world>
+```
+
+Where `<world>` is either a path to an SDF file or the name of a built-in world in `husarion_gz_worlds`. The default world is `husarion_world`.
+
+Available worlds:
+- `cave`
+- `empty_with_plugins`
+- `husarion_office`
+- `husarion_world`
+- `rubicon`
+- `sonoma_raceway`
+
 ## Configuration Files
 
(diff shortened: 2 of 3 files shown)


rafal-gorecki (changes requested):
Sorry, I forget about this review (ping me in the future ;) ). 

Looking on this PR IMO, we duplicate the logic of `gz_world` arg. What about adding additional world like this: https://github.com/husarion/husarion_gz_worlds/blob/main/worlds/sonoma_raceway.sdf
where we can specify x, y, z offset.

At best, could we simplify the execution of this argument so that we don't have to enter the entire path, but instead have aliases to the ones we already have?
So both
`gz_world:=/path/to/racaway.sdf`
and
`gz_world:=raceway`
could work.

husarafal:
Hi @rafal-gorecki,

The reason why it is a new argument in `husarion_gv_gazebo` is because it's not only exchanging the SDF, but also setting up the initial position of the robot.

I attempted moving the entire world in reverse by creating a new SDF in the `husarion_gz_worlds` that referenced the remote world from fuel, but this didn't work well for Rubicon World. There, the heightmap terrain didn't follow the  transformation and became misaligned with the mesh entities.

So that leads me to a conclusion that we shouldn't be trying to modify the worlds themselves, but only move the robot within them.

If we really do not wish to add new arguments, I could modify `husarion_gz_worlds`' `gz_world` arg to accept single word filenames as you suggested, and then in `husarion_ugv_gazebo` I could tap into this argument and override the initial position of the robot while passing the arg down to `husarion_gz_worlds` as usual.

husarafal:
Okay, I have refactored everything, implementing new worlds as .sdf files in `husarion_gz_worlds`. Also modified the `gz_world` parameter to accept single-word identifiers as alternative to full paths to .sdf files.

Most changes were introduced in `husarion_gz_worlds`: https://github.com/husarion/husarion_gz_worlds/pull/11
Those in **this PR right here** is just docs. And I have changed the initial Y position in `spawn_robot.launch.py` to be at zero.
Initial Z position was left at 0.2 so that the footprint is at world origin instead of base_link.

Test each new world using:

```bash
ros2 launch husarion_ugv_gazebo simulation.launch.py gz_world:=rubicon
```

Worlds to test:

- `cave`
- `empty_with_plugins`
- `husarion_office`
- `husarion_world`
- `rubicon`
- `sonoma_raceway`

husarion_ugv/simulation_deps.repos
```diff
@@ -10,7 +10,7 @@ repositories:
   husarion_gz_worlds:
     type: git
     url: https://github.com/husarion/husarion_gz_worlds.git
-    version: 97275568c34e6fe191ca34fa1c1c2b897faf9610
+    version: 3cc8018a49de082f5e4d3fa7afa4e30dda9307c6
```
rafal-gorecki: After merge, commit has changed
```suggestion
    version: 56b5788a9584d3bc31ba85d36582ba965752c27e
```

rafal-gorecki (approved):
Please update hash in vcs after merge.

rayferric:
Shouldn't have merged this one then, right? Instead we could've updated the hash after merging `husarion_gz_worlds`, but before merging here.

PR #640, opened 2026-03-02
Fix duplicate `robot_state_publisher` nodes during sim

teemingc:
### Description

This PR ensures only one `robot_state_publisher` node runs when the simulation is started.

The controller isn't supposed to start its own state publisher by passing the following argument https://github.com/husarion/husarion_ugv_ros/blob/a260518d91a5393825dc38aa53e3c379023eefd6/husarion_ugv_gazebo/launch/simulate_robot.launch.py#L167 but that argument is never recognised. Our PR fixes that. This ensures the state publisher started by the simulation launch is the only one https://github.com/husarion/husarion_ugv_ros/blob/a260518d91a5393825dc38aa53e3c379023eefd6/husarion_ugv_gazebo/launch/spawn_robot.launch.py#L92

This helps avoid the warning shown when running `ros2 node list` and other related issues
```sh
❯ ros2 node list
WARNING: Be aware that there are nodes in the graph that share an exact name, which can have unintended side effects.
```

### Requirements

- [ ] (Dependency PR)
- [ ] Backport
- [ ] Documentation updated
- [ ] Other repositories updated `.repos`

### Tests 🧪

- [ ] Robot
- [ ] Container
- [ ] Simulation

## Summary by CodeRabbit

* **New Features**
  * Added a launch-time option to enable or disable robot state publishing (enabled by default).
  * URDF/model resources and related startup steps are now loaded only when robot state publishing is enabled, reducing unnecessary initialization when disabled.
  * The new option is exposed alongside existing launch settings (namespace, log level) for consistent runtime configuration.

### DIFF
--- a/husarion_ugv_controller/launch/controller.launch.py
+++ b/husarion_ugv_controller/launch/controller.launch.py
@@ -230,6 +230,7 @@ def generate_launch_description():
         declare_wheel_type_arg,  # wheel_type is used by controller_config_path
         declare_controller_config_path_arg,
         declare_namespace_arg,
+        declare_publish_robot_state_arg,
         declare_use_sim_arg,
         declare_log_level_arg,
         SetParameter(name="use_sim_time", value=use_sim),
--- a/husarion_ugv_controller/launch/controller.launch.py
+++ b/husarion_ugv_controller/launch/controller.launch.py
@@ -76,15 +76,18 @@ def generate_launch_description():
         ),
     )
 
-    actions = [
-        declare_common_dir_path_arg,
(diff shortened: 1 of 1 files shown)


delihus:
Thank you, @teemingc.
It does solve that problem, but I wonder if this parameter is even necessary. I did PR with removed parameter and the second loading of `load_urdf` https://github.com/husarion/husarion_ugv_ros/pull/642

@rafal-gorecki Could you look at both changes?

rafal-gorecki:
Solved here: #642. Closing

PR #651, opened 2026-06-12
Rework simulation e-stop and LED strip visualization

### LINKED ISSUE #623: gz_ros2_control/GazeboSimSystem fails to load after husarion_ugv_gazebo/EStopSystem
**Problem description:**

For our use-case, we would like to spawn multiple robots in Gazebo.
All the robots are controlled using ROS2 control.
Therefore, these robots have the `gz_ros2_control/GazeboSimSystem` plugin defined in their URDF.

We also use the Panther as one of the robots.
The Panther uses the custom made `husarion_ugv_gazebo/EStopSystem` plugin.
As far as we know, this plugin extends the E-Stop system to the original ros2_control plugin.

We obtain the following error when we first spawn the Panther (with the `husarion_ugv_gazebo/EStopSystem` plugin) and next another robot (with the  `gz_ros2_control/GazeboSimSystem` plugin):

`[ros_control.GZResourceManager]: The plugin failed to load for some reason. Error: MultiLibraryClassLoader: Could not create class of type gz_ros2_control::GazeboSimSystem`.

Note that there is no problem when we first spawn the other robot followed by the Panther.
Note that there is also no problem when spawning multiple Panthers or multiple other robots.

**A minimal example:**

This examples requires *husarion_ugv_ros* and the *gz_ros2_control_demos* package.

simulation.launch.py:
```
import os

from ament_index_python.packages import get_package_share_directory
from launch import LaunchDescription
from launch.actions import ExecuteProcess, IncludeLaunchDescription
from launch_ros.actions import Node


def generate_launch_description() -> LaunchDescription:
    gazebo = ExecuteProcess(cmd=["gz", "sim", "empty.sdf"])

    gazebo_package = get_package_share_directory("ros_gz_sim")
    gazebo_launch = os.path.join(gazebo_package, "launch", "gz_sim.launch.py")
    gazebo = IncludeLaunchDescription(
        gazebo_launch, launch_arguments=[("gz_args", ["empty.sdf"])]
    )

    bridge = Node(
        package="ros_gz_bridge",
        executable="parameter_bridge",
        arguments=["/clock@rosgraph_msgs/msg/Clock[gz.msgs.Clock"],
        output="screen",
    )

    return LaunchDescription(
        [
            gazebo,
            bridge,
        ]
    )
```

panther.launch.py:
```
import os

import xacro
from ament_index_python.packages import get_package_share_directory
from launch import LaunchDescription
from launch_ros.actions import Node


def generate_launch_description() -> LaunchDescription:
    namespace = "panther"
    panther_package = get_package_share_directory("husarion_ugv_description")
    panther_xacro_path = os.path.join(panther_package, "urdf", "panther.urdf.xacro")
    panther_robot_description = xacro.process_file(
        panther_xacro_path, mappings={"use_sim": "true", "namespace": namespace}
    ).toxml()

    panther_state_publisher = Node(
        package="robot_state_publisher",
        executable="robot_state_publisher",
        parameters=[{"robot_description": panther_robot_description}],
        namespace=namespace,
    )

    panther_spawner = Node(
        package="ros_gz_sim",
        executable="create",
        arguments=["-topic", f"/{namespace}/robot_description", "-name", namespace],
        namespace=namespace,
    )

    return LaunchDescription(
        [
            panther_state_publisher,
            panther_spawner,
        ]
    )
```

other_robot.launch.py:
```
import os

import xacro
from ament_index_python.packages import get_package_share_directory
from launch import LaunchDescription
from launch_ros.actions import Node


def generate_launch_description() -> LaunchDescription:
    namespace = "diff_drive_2"
    diff_drive_package = get_package_share_directory("gz_ros2_control_demos")
    diff_drive_xacro_path = os.path.join(
        diff_drive_package, "urdf", "test_diff_drive.xacro.urdf"
    )
    diff_drive_description = xacro.process_file(
        diff_drive_xacro_path, mappings={"namespace": namespace}
    ).toxml()

    diff_drive_state_publisher = Node(
        package="robot_state_publisher",
        executable="robot_state_publisher",
        parameters=[{"robot_description": diff_drive_description}],
        namespace=namespace,
    )

    diff_drive_spawner = Node(
        package="ros_gz_sim",
        executable="create",
        arguments=["-topic", f"/{namespace}/robot_description", "-name", namespace],
        namespace=namespace,
    )

    return LaunchDescription(
        [
            diff_drive_state_publisher,
            diff_drive_spawner,
        ]
    )
```

First launch *simulation.launch.py*.

- The error does not occur when next launching *panther.launch.py* followed by *other_robot.launch.py*.
- The problem occurs when next launching *other_robot.launch.py* followed by  *panther.launch.py*.

We think loading of the `husarion_ugv_gazebo/EStopSystem` plugin somehow causes the `gz_ros2_control/GazeboSimSystem` plugin to fail when loaded afterwards.

rafal-gorecki:
## Description

Fixes #623.

- **E-stop**: `EStopSystem` rewritten from a `gz_ros2_control` hardware plugin subclass into a standalone Gazebo System plugin. The robot now uses the stock `gz_ros2_control/GazeboSimSystem`, which removes the copied `GazeboSimSystemPrivate` internals (the plugin-loader symbol clash behind #623). While the E-stop is active the plugin zeroes `JointVelocityCmd`, so the robot actually stops instead of keeping the last commanded velocity. Same ROS API (`hardware/e_stop*`).
- **LEDStrip**: markers are re-sent only for LEDs whose color changed (previously ~920 `/marker` requests/s re-sent every frame). New `led_range` and `points` parameters allow an instance to display a subrange of a channel frame along an arbitrary polyline; polyline LEDs are drawn as continuous triangle ribbons (no gaps at corners). Marker parent uses the model name instead of a hardcoded `panther`.
- **Lynx**: 4 corner LED segments (front angled strips, rear U-shaped strips) with wide downward spots approximating a strip glowing radially; lights and manager enabled in simulation.
- **Launch**: gz bridge `model_name` falls back to `robot_model` instead of hardcoded `panther` (battery state was not bridged for Lynx, leaving `lights_manager` waiting forever).

## Requirements

- [x] Documentation updated (`husarion_ugv_gazebo/README.md`)

## Tests

- [ ] Robot
- [ ] Container
- [x] Simulation

Tested in Gazebo for both Panther and Lynx: full e-stop cycle (reset → drive → trigger → stop → reset → drive), LED visualization verified visually on GPU. Not tested on hardware or in a container.

## Release Notes

* **New Features**
  * E-stop is now a standalone Gazebo system plugin that publishes E-stop state and exposes reset/trigger ROS services, with configurable namespace and initial state.
  * Enhanced LED strip visualization using polyline geometry, including `led_range` and optional `points`, plus spot/light rendering options and corner LED anchors on Lynx.
  * Added `lights` launch control and updated Gazebo URDF xacro macros for E-stop and lights.

* **Bug Fixes**
  * Improved LED marker rendering performance and robustness, including safer parameter/image handling and more efficient per-frame updates.

* **Documentation**
  * Updated E-stop and LED strip configuration docs to match the new options.


husarion_ugv_description/urdf/lynx/lynx_macro.urdf.xacro
```diff
@@ -187,10 +186,18 @@
...
-      <!-- TODO Add lights visualization -->
-      <!-- <xacro:gazebo.light reference_frame="lights_channel_1_link" name="lights_channel_1" topic="lights/channel_1_frame" namespace="${namespace}" /> -->
-      <!-- <xacro:gazebo.light reference_frame="lights_channel_2_link" name="lights_channel_2" topic="lights/channel_2_frame" namespace="${namespace}" /> -->
+      <!-- Corner LED strips. Front: a straight 5 cm strip angled 60 deg to the front fascia,
+           rising from 0.15 m (inner end) to 0.18 m (outer end). Rear: a U-shaped strip (inner
+           edge, back, outer edge). Channel layout from lynx_animations.yaml: channel 1 fl=23-12
+           rl=0-11, channel 2 fr=23-12 rr=0-11. -->
+      <!-- Wide downward-facing spots approximate a strip glowing radially in the horizontal
+           plane (no light cast into the sky). -->
+      <xacro:gazebo.light reference_frame="lights_fl_link" name="lights_fl" topic="lights/channel_1_frame" namespace="${namespace}" spot_outer_angle="1.5" led_range="23-12" points="-0.01 -0.065 0.0; 0.0 -0.063 0.0; 0.005 0.056 0.0" />
```
KmakD: Image

I think we can better approximate actual led.

husarion_ugv_description/urdf/lynx/lynx_macro.urdf.xacro
```diff
@@ -187,10 +186,18 @@
...
+      <xacro:gazebo.light reference_frame="lights_fr_link" name="lights_fr" topic="lights/channel_2_frame" namespace="${namespace}" spot_outer_angle="1.5" led_range="23-12" points="-0.01 0.065 0.0; 0.0 0.063 0.0; 0.005 -0.056 0.0" />
+      <xacro:gazebo.light reference_frame="lights_rl_link" name="lights_rl" topic="lights/channel_1_frame" namespace="${namespace}" spot_outer_angle="3.0" led_range="0-11" points="-0.029 0.04 0.01; 0.0 0.023 0.006; 0.006 -0.03 -0.005; -0.067 -0.035 -0.005" />
```
KmakD: Image

There are some overlaps with model leds and the fender

husarion_ugv_gazebo/launch/simulate_robot.launch.py
KmakD, line 1: It would be nice if there was a possibility to disable lights in simulation

husarion_ugv_gazebo/src/led_strip.cpp
```diff
@@ -153,6 +195,12 @@ void LEDStrip::ImageCallback(const gz::msgs::Image & msg)
     return;
   }
 
+  const size_t step = msg.pixel_format_type() == gz::msgs::PixelFormatType::RGBA_INT8 ? 4 : 3;
+  if (msg.data().size() < msg.width() * msg.height() * step) {
+    ignerr << "Error: Image data smaller than the declared dimensions." << std::endl;
```
KmakD: `ignerr` is deprecated. New gazebo uses `gzerr`. I guess we can change this in whole file, also other logging macros - ignlog igndbg etc.

KmakD (changes requested):
I have still experienced the simulation slowdown after some time, for lynx at least. 

As I pointed in one of the comments it would be nice if it was possible to disable lights simulation using launch arg. If that is out of scope of this PR then it's ok.

No more thing I thought about for future. It would be nice If we could add some Gaussian blur for the sim leds. Real robot has a plastic on leds that gives this effect. It would be nice if we could produce the same effect in simulation. This might be a bit more complicated tho

KmakD (commented):
I see one problem with current blur implementation. The resolution for the blur depends on leds resolution. This makes some of animations appear different. For example MANUAL_ACTION (id: 4), instead of having two distinct strips, blur makes it looks like one big.
Alternative approach I think would give better result would be to have multiple markers per led (could be adjusted with a param like `render_markers_per_led`). Don't know how this would affect the simulation performance with increasing markers.
Let me know what you think, we don't have to change this right now, maybe just decrease sigma_blur to make MANUAL_ACTION visible.

One other thing I thought about, was to make the "background color" of the strip white (or rather light grey). Animations with transparency are now blending into black, which makes them darker compared to real LEDs. This of course can be outside of scope of this PR, if you rather want to create task or issue to track this, then I'm ok. Current implementation is good enough.

rafal-gorecki:
1. **Blur / MANUAL_ACTION merge:** dropped the Gaussian (it blurred the per-LED array, so width scaled with LED count and merged the bars) and went with your `render_markers_per_led` idea — local interpolation between LED centers, so the two bars stay distinct. As you suspected, `>1` multiplies markers and stutters the GUI, so the default is `1` (sharp), which already fixes MANUAL_ACTION. A truly smooth blur isn't cheaply doable in gz8 (markers don't render per-vertex colors on triangles; dynamic textures are blocked by `SetTexture`).

2. **White background:** done — new `strip_base_color` param (default white). Transparent animations now composite over it instead of blending into black.

Also batched markers into one /marker_array per frame + run-length merge to kill the full-strip-fade stutter.

husarion_ugv_gazebo/include/husarion_ugv_gazebo/led_strip.hpp
```diff
@@ -183,20 +196,38 @@ class LEDStrip : public gz::sim::System,
...
+  struct ColorRun
+  {
+    size_t start;
+    size_t end;
+    gz::math::Color color;
+    bool operator==(const ColorRun & other) const
+    {
+      return start == other.start && end == other.end && color == other.color;
+    }
+  };
```
KmakD: This should be outside of LEDStrip class definition

husarion_ugv_gazebo/src/led_strip.cpp
```diff
@@ -303,71 +319,116 @@ std::vector<gz::math::Color> LEDStrip::ExtractLedColors(const gz::msgs::Image &
...
-  std::vector<double> kernel(radius + 1);
-  double norm = 0.0;
-  for (int k = 0; k <= radius; ++k) {
-    kernel[k] = std::exp(-(k * k) / (2.0 * blur_sigma_ * blur_sigma_));
-    norm += (k == 0 ? kernel[k] : 2.0 * kernel[k]);
-  }
+// Non-blocking /marker_array response handler: the result is irrelevant (markers are dropped when
+// no MarkerManager is running, e.g. headless), but a callback keeps the request asynchronous.
+void OnMarkerArrayResponse(const gz::msgs::Boolean & /*reply*/, const bool /*result*/) {}
+}  // namespace
```
KmakD: Why those are in namespace and not defined as LEDStrip class methods?

husarion_ugv_gazebo/src/led_strip.cpp
```diff
@@ -303,71 +319,116 @@ std::vector<gz::math::Color> LEDStrip::ExtractLedColors(const gz::msgs::Image &
...
+  // Remove markers left over from a frame that had more runs than this one.
+  for (size_t r = runs.size(); r < last_runs_.size(); ++r) {
+    auto * marker = marker_array.add_marker();
+    marker->set_action(gz::msgs::Marker::DELETE_MARKER);
+    marker->set_ns(ns_ + light_name_);
+    marker->set_id(r + 1);
+  }
+
+  if (marker_array.marker_size() > 0) {
+    node_.Request("/marker_array", marker_array, &OnMarkerArrayResponse);
```
KmakD: Is this OnMarkerArrayResponse required? It's an empty method and previously when calling `node_.Request("/marker", marker_msg);` there were no response callback used

KmakD (changes requested):
I've changed `render_markers_per_led` to `2` and the simulation was working without lags for both Panther and Lynx, and the visuals were way better. Tested locally not in a docker. Tell me what you think about increasing this.

I don't see the difference before and after `strip_base_color`, it looks the same:

Image

Image

husarion_ugv_gazebo/include/husarion_ugv_gazebo/led_strip.hpp
rafal-gorecki: I move it outside.

husarion_ugv_gazebo/src/led_strip.cpp
rafal-gorecki: Leave only static `OnMarkerArrayResponse` in empty namespace prevent name conflict.
Alternative:
```cpp
class MyClass
{
public:
  static void OnMarkerArrayResponse(
      const gz::msgs::Boolean& reply,
      bool result)
  {
  }
};
```

and

```cpp
node_.Request(
    "/marker_array",
    marker_array,
    &MyClass::OnMarkerArrayResponse);
```

husarion_ugv_gazebo/src/led_strip.cpp
rafal-gorecki: The callback is required.

rafal-gorecki:
I set `render_markers_per_led` to 1 because I noticed that GUI lag was visible starting from 3. It's hard to predict how it will behave on computers with different hardware. If 2 markers per led look fine for you, we can leave it

The slowdown occurs regardless of RTF. RTF is unaffected because the plugin no longer touches the server layer, only the GUI.

husarion_ugv_description/urdf/lynx/lynx_macro.urdf.xacro
```diff
@@ -187,10 +186,18 @@
...
       <xacro:gazebo.imu reference_frame="imu_link" namespace="${namespace}" />
-      <!-- TODO Add lights visualization -->
-      <!-- <xacro:gazebo.light reference_frame="lights_channel_1_link" name="lights_channel_1" topic="lights/channel_1_frame" namespace="${namespace}" /> -->
-      <!-- <xacro:gazebo.light reference_frame="lights_channel_2_link" name="lights_channel_2" topic="lights/channel_2_frame" namespace="${namespace}" /> -->
+      <!-- Corner LED strips. Front: a straight 5 cm strip angled 60 deg to the front fascia,
+           rising from 0.15 m (inner end) to 0.18 m (outer end). Rear: a U-shaped strip (inner
+           edge, back, outer edge). Channel layout from lynx_animations.yaml: channel 1 fl=23-12
+           rl=0-11, channel 2 fr=23-12 rr=0-11. -->
+      <!-- Wide downward-facing spots approximate a strip glowing radially in the horizontal
+           plane (no light cast into the sky). -->
```
rafal-gorecki:
```suggestion
```

husarion_ugv_gazebo/src/led_strip.cpp
KmakD: I still don't get this. I think it would work if you define it as class method and  reference it in Request call like this:
```
node_.Request("/marker_array", marker_array, &LEDStrip::OnMarkerArrayResponse, this);
```

husarion_ugv_gazebo/include/husarion_ugv_gazebo/led_strip.hpp
```diff
@@ -116,26 +132,109 @@ class LEDStrip : public gz::sim::System,
...
+   * color is blended over strip_base_color_ by its brightness so dim LEDs reveal the diffuser
+   * instead of going black. Returns led_count * render_markers_per_led_ opaque colors.
+   */
+  std::vector<gz::math::Color> ResampleLedColors(
+    const std::vector<gz::math::Color> & led_colors) const;
+
+  /**
+   * @brief Linearly interpolate between two colors, per RGBA channel.
+   */
+  static gz::math::Color Lerp(const gz::math::Color & a, const gz::math::Color & b, double t);
```
KmakD, line 156:
```suggestion
  gz::math::Color Lerp(const gz::math::Color & a, const gz::math::Color & b, double t) const;
```
Why define it as static and not simply const?

KmakD (changes requested):
Something about blending into white is not working. Maybe we should drop this?
Image

husarion_ugv_gazebo/src/led_strip.cpp
rafal-gorecki: Added

husarion_ugv_gazebo/include/husarion_ugv_gazebo/led_strip.hpp
rafal-gorecki: Lerp is pure mathematics on its arguments (a, b, t). It doesn't read any field of the `LEDStrip` object—if you removed all the class fields, Lerp would still work identically.

- `static` is recommended to say: "this function doesn't depend on the state of the instance." There's no hidden this; it can be called without an object. This is true of Lerp.
- `const` is recommended to say: "this is an instance method that doesn't modify the object, but can read its fields." It suggests that the function conceptually belongs to the object and uses it—which is misleading, because Lerp doesn't take anything from the object.

husarion_ugv_gazebo/src/led_strip.cpp
```diff
@@ -215,65 +290,240 @@ void LEDStrip::VisualizeLights(gz::sim::EntityComponentManager & ecm, const gz::
...
+  return colors;
+}
 
-  double step_width = marker_width_ / image.width();
-  double step_height = marker_height_ / image.height();
+namespace
+{
+// Thickness (X) of the flat legacy-strip marker box; the strip is rendered nearly 2D.
+constexpr double kLegacyStripThickness = 0.001;
+}  // namespace
```
KmakD: I think we can finally remove that namespace, and the `kLegacyStripThickness` can be moved to header file

KmakD (changes requested):
What about this one? https://github.com/husarion/husarion_ugv_ros/pull/651#pullrequestreview-4572901938

rafal-gorecki:
> Something about blending into white is not working. Maybe we should drop this? Image

What do you mean? For some reason, it isn't snow-white (maybe it's the lighting), but the functionality works.
```xml
strip_base_color:='1 0 1'
```
image

KmakD:
> What do you mean? For some reason, it isn't snow-white (maybe it's the lighting), but the functionality works.
> 
The animation should be light red stripes on darker red background. Here the background is white instead

3024x4032

rafal-gorecki:
Thanks @KmakD  for the real photo – I’d forgotten what it looks like in reality. I fixed it by adding an alpha channel to `led_strip_color`, and now the LED strip looks better. 
image

KmakD (approved):

PR #659, opened 2026-07-14
Rt mutex fix

DominikN:
### Description

- The motor e-stop was latching on its own under load on the RPi4. The control loop reads the motor state timestamps under a std::mutex it shares with the CANopen thread, and since std::mutex has no priority inheritance the loop could end up waiting on that lock long enough to hit the PDO timeout and latch the e-stop.
- Made those timestamps atomic nanoseconds so the loop reads them without locking.
- Also fixed the CANopen thread priority. ConfigureRT was called in Initialize(), which only sets the calling thread, so the CANopen loop thread was left to inherit it. Now it's set on the thread itself and bumped to 70 so reading CAN frames isn't starved by the control loop when the CPU is busy.
- Tested on a Lynx with a 10h drive soak - 0 e-stop latches and 0 overruns (was ~4 latches per 10h before).

### Requirements

- [ ] (Dependency PR)
- [ ] Backport
- [ ] Documentation updated 
- [ ] Other repositories updated `.repos`

### Tests 🧪

- [x] Robot 
- [ ] Container 
- [ ] Simulation 

## Summary by CodeRabbit

* **Performance**
  * Improved CANopen real-time scheduling behavior.
  * Reduced synchronization overhead for Roboteq state and timestamp processing.
* **Reliability**
  * Added CANopen PDO timeout watchdog that can trigger a controlled restart after sustained resync timeout.
  * Added Roboteq channel validation during timestamp updates.
* **New Features**
  * Added `launch_gamepad` option to control teleop gamepad startup.
  * Added `BRINGUP_EXTRA_ARGS` support for demo launch customization.
* **Tests**
  * Extended unit test coverage for the CANopen resynchronization watchdog.
* **Chores**
  * Updated the hardware Docker image to a safer multi-stage build with a dedicated entrypoint.


husarion_ugv_hardware_interfaces/include/husarion_ugv_hardware_interfaces/robot_system/robot_driver/canopen_manager.hpp
```diff
@@ -112,8 +112,12 @@ class CANopenManager
...
   void NotifyCANCommunicationStarted(const bool result);
 
-  // Priority set to be higher than the priority of the main ros2 control node (50)
-  static constexpr unsigned kCANopenThreadSchedPriority = 60;
+  // Priority of the CANopen communication thread. It has to be higher than the control
+  // loop so that reading CAN frames (and stamping the timestamps the loop reads) is not
+  // starved when the CPU is busy. If the CAN thread is starved the timestamps go stale and
+  // the driver reports a PDO timeout, which latches the e-stop. The control loop runs at
+  // FIFO 60 here, so this is set above it.
+  static constexpr unsigned kCANopenThreadSchedPriority = 70;
```
KmakD, line 120: Why are we increasing this to 70? The control loop priority is 50, it looks like increasing this further won't make any difference

husarion_ugv_hardware_interfaces/src/robot_system/robot_driver/canopen_manager.cpp
```diff
@@ -69,6 +60,19 @@ void CANopenManager::Activate()
   }
 
   canopen_communication_thread_ = std::thread([this]() {
+    // Set the RT priority here, on the thread that actually runs the CANopen loop. It used
+    // to be done in Initialize(), which only changed the calling thread's priority and left
+    // this one to inherit it, which is easy to get wrong.
```
KmakD, line 65:
```suggestion
    // Set the RT priority here, on the thread that actually runs the CANopen loop.
```
Simple and concise comments makes code easier to read. There usually is no need to explain how something used to be done in past in the code. PR description is enough for this purpose. Applies to all comments.

husarion_ugv_hardware_interfaces/include/husarion_ugv_hardware_interfaces/robot_system/robot_driver/roboteq_driver.hpp
```diff
@@ -109,17 +129,17 @@ class RoboteqDriver : public DriverInterface, public lely::canopen::LoopDriver
    */
   timespec GetPositionTimestamp(const std::uint8_t channel)
   {
-    std::lock_guard<std::mutex> lck(position_timestamp_mtx_);
-    return last_position_timestamps_.at(channel);
+    return NanosecondsToTimespec(
+      last_position_timestamps_ns_.at(channel - kChannel1).load(std::memory_order_acquire));
```
KmakD, line 133: Can we add helper method to extract channel index? Something like:
```
std::size_t ChannelIndex(const std::uint8_t channel) const;
```

KmakD (changes requested):

delihus:
Changing to draft. More features will be added or splited from this PR.

